PR #4, opened 2024-08-15
feat: add native azure openai configuration

wey-gu:
chore: also fixed markdown lint errors

Sorry, I just noticed #2 already supports Azure after I started the change, it's quite a chore in such cases...

Thus this change mainly:

- introduce a separate config interface
- apply auth headers based on Azure or openai, instead of both


wey-gu:
ps: I tested Azure with this new config flavor, but didn't regression verify the openai case as I don't have a handy API as of now in hand but I stared at the change again to ensure it's not impacting vanilla openai.

yetone:
LGTM!

PR #5, opened 2024-08-15
fix: escape original_content

yuchanns:
`str:gsub` cannot correctly handle `original_content` if it contains certain Lua special symbols.

For example:
```go
package main

import "strings"

func main() {
	strings.Replace("https://login.xxx.com/query?redirect=https%3A%2F%2Fxx.com", "%2F", "/", -1)
}
```
The problematic symbols are `%2` and `%3`.

Let's fix this.


yetone:
‌‌LGTM! There are some code conflicts that need to be resolved.

PR #13, opened 2024-08-15
refactor(ui): bounding popover

aarnphm:
Bounds the buffer to a separate windows

add initial OOP for view and renderer sidebar

add support for toggle and focus correctly to navigate


lua/avante/view.lua
```diff
@@ -0,0 +1,67 @@
...
+---setup view buffer
+---@param split_command string A split command to position the side bar to
+---@param size integer a given % to resize the chat window
+---@return avante.View
+function View:setup(split_command, size)
+  -- create a scratch unlisted buffer
+  self.buf = vim.api.nvim_create_buf(false, true)
+
+  -- set filetype
+  vim.api.nvim_set_option_value("filetype", "Avante", { buf = self.buf })
```
yetone, line 24: ‌Why isn't the filetype here markdown?

aarnphm: Ah I make it a separate filetype so that it is easier for us to control, instead of coerce the buffer to markdown.

but I might revert it

PR #18, opened 2024-08-16
feat: configurable history directory

### LINKED ISSUE #17: AzureAPI should not force to setup OPENAI_API_KEY
https://github.com/yetone/avante.nvim/blob/26748729455cfc95704c204ec0d6cb548d99f493/lua/avante/ai_bot.lua#L156

yuchanns:
Users should be able to set up a custom history directory.

Also fix #17

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -7,6 +7,7 @@ local config = require("avante.config")
 local ai_bot = require("avante.ai_bot")
 local api = vim.api
 local fn = vim.fn
+local utils = require("avante.utils")
 
 local RESULT_BUF_NAME = "AVANTE_RESULT"
 local CONFLICT_BUF_NAME = "AVANTE_CONFLICT"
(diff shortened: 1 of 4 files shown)


lua/avante/sidebar.lua
```diff
@@ -227,9 +228,16 @@ end
...
 -- Function to get the chat history file path
 local function get_chat_history_file()
+  local history_dir = config.get().history_dir
   local project_root = get_project_root()
+  if history_dir == nil then
+    history_dir = Path:new(project_root, ".avante_chat_history")
+  elseif type(history_dir) == "function" then
+    history_dir = history_dir(project_root)
+  else
+    history_dir = Path:new(history_dir, project_root:match("([^/]+)$"), ".avante_chat_history")
```
yetone, line 238: ‌‌‌‌If placed outside the project_root directory, wouldn't it be impossible to use relative_path?

https://github.com/yetone/avante.nvim/pull/18/files#diff-ddcb4139fbc53f068a08051ae32665c76507f66698fe427fcfdd384a36dce2fcR222

yuchanns: SGTM

yetone:
‌‌‌‌Rather than making it user-configurable, why don't we implement a more disruptive update (since we're still in early versions) by moving the history directory to the $HOME directory and recording the absolute path? What do you think?

lua/avante/sidebar.lua
yuchanns: Is this what you want? I'm sorry, but I don't quite understand your meaning.

yuchanns:
Says that we have a project under $HOME/coding/proj1, the history will be saved to `$HOME/.avante_history/$HOME/coding/proj1`? Or `$HOME/.avante_history/coding/proj1`? Or `$HOME/.avante_history/proj1`?

yetone:
For example, my chat history in `/Users/yetone/workspace/projects/test/foo/bar.go` will be saved in `$HOME/.avante_history/chat/Users__yetone__workspace__projects__test__foo__bar.go.json`

yetone:
However, according to @aarnphm's suggestion, we should use [shada](https://neovim.io/doc/user/starting.html#shada-file) to store the history. I'm concerned that future refactoring might disrupt your current efforts. Are you interested in doing this refactoring?

yuchanns:
Nevermind. I'll close this PR. Since using shade is better IMO.

PR #39, opened 2024-08-17
feat: ask selected code block

yetone:


lua/avante/config.lua
```diff
@@ -53,6 +54,10 @@ M.options = {}
 ---@param opts? avante.Config
 function M.setup(opts)
   M.options = vim.tbl_deep_extend("force", M.defaults, opts or {})
+  if M.options.mappings.show_sidebar ~= nil then
```
aarnphm, line 57: I think it is ok to make breaking change, since it is still alpha, no need to add more debt here

PR #54, opened 2024-08-17
feat(ui): add envvar popup with warnings

aarnphm:
- **feat: add support for input popup of envvar**
- **chore: update README and ignore certain filetypes for popup**


yetone:
great!

PR #55, opened 2024-08-17
feat(experimental): add section for behaviour

aarnphm:

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -649,16 +649,15 @@ function Sidebar:render()
           end
         )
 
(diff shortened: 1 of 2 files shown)


lua/avante/sidebar.lua
```diff
@@ -649,16 +649,15 @@ function Sidebar:render()
           end
         )
 
-        api.nvim_set_current_win(self.winid.result)
```
yetone, line 652: ‌‌‌Is the cursor still remaining in the input after completion?

aarnphm: yes

PR #60, opened 2024-08-18
fix: azure api_key fallback

yuchanns:
af1f51455eab6ced269ec8152a0a1f1ab4ed3e3c has broken azure api_key fallback

### DIFF
--- a/lua/avante/ai_bot.lua
+++ b/lua/avante/ai_bot.lua
@@ -12,17 +12,13 @@ local Tiktoken = require("avante.tiktoken")
 ---@class avante.AiBot
 local M = {}
 
----@class Environment: table<[string], any>
----@field [string] string the environment variable name
----@field fallback? string Optional fallback API key environment variable name
-
 ---@class EnvironmentHandler: table<[Provider], string>
 local E = {
-  ---@type table<Provider, Environment | string>
+  ---@type table<Provider, string>
   env = {
     openai = "OPENAI_API_KEY",
     claude = "ANTHROPIC_API_KEY",
-    azure = { "AZURE_OPENAI_API_KEY", fallback = "OPENAI_API_KEY" },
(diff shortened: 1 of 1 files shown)


lua/avante/ai_bot.lua
```diff
@@ -353,7 +353,7 @@ local function call_openai_api_stream(question, code_lang, code_content, selecte
 
   local url, headers, body
   if Config.provider == "azure" then
-    api_key = os.getenv(E.key("azure"))
```
aarnphm, line 356: my reasoning behind this is that for azure, they should configure the azure env instead of the openai env.

It should already check whether azure openai env is set on UI popup.

Maybe we should add the check in setup instead of this fallback? I would rather setting api_key explicitly instead of type coercion here.

yuchanns: SGTM

yuchanns: How about removing the fallback in the `E.env`? So we don't need to care about fallback anymore

aarnphm: that works as well.

aarnphm:
can you also update the setmetatable logic, since now the type should be string only.

lua/avante/ai_bot.lua
```diff
@@ -31,19 +27,8 @@ E = setmetatable(E, {
...
-      local value = os.getenv(main_key)
-      if value then
-        return true
-      elseif envvar.fallback then
-        return os.getenv(envvar.fallback) and true or false
-      end
-    end
-    return false
+    local value = os.getenv(envvar)
+    return value and true or false
```
aarnphm, line 31:
```suggestion
    return os.getenv(envvar) and true or false
```

aarnphm (approved):

PR #61, opened 2024-08-18
fix(ui): invalid buf id on hide the sidebar during generation

yuchanns:
When the AI bot is generating answers and users happen to close the sidebar, it occurs:

```

Error executing vim.schedule lua callback: ...local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:231: Invalid 'buffer': Expected Lua number
stack traceback:
	[C]: in function 'nvim_buf_set_lines'
	...local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:231: in function ''
	vim/_editor.lua: in function ''
	vim/_editor.lua: in function <vim/_editor.lua:0>

```

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -224,6 +224,9 @@ end
 function Sidebar:update_content(content, focus, callback)
   focus = focus or false
   vim.defer_fn(function()
+    if not self.view.buf then
+      return
+    end
     api.nvim_set_option_value("modifiable", true, { buf = self.view.buf })
     api.nvim_buf_set_lines(self.view.buf, 0, -1, false, vim.split(content, "\n"))
     api.nvim_set_option_value("modifiable", false, { buf = self.view.buf })


lua/avante/sidebar.lua
```diff
@@ -224,6 +224,9 @@ end
 function Sidebar:update_content(content, focus, callback)
   focus = focus or false
   vim.defer_fn(function()
+    if not self.view.buf then
```
aarnphm, line 227: This doesn't actually terminate the buffer, and probably a quick hack.

I think a correct solution for this is to terminate the stream from plenary.

We should still keep this, but add the plenary check.

yuchanns: I don't think so. Users may not intend to end the plenary stream when they close the sidebar. This may be an accidental action or just want to close it and check the answer later, as it's saved in the history.

aarnphm: if the buffer won't get updated regardless, since https://github.com/yetone/avante.nvim/blob/c19ea9a48a9eb7c2cf17c507d3e59a68214130f3/lua/avante/view.lua#L55

yuchanns: Well, that's interesting, cause I can get an updated buffer by reopening it no matter how many times as long as it is streaming.

And if the stream is ended, history is loaded when open.

aarnphm: I think this has to do with the code being messy hehe. I was thinking to refactor the stream to a Plenary.Job to handle termination correctly.

This is probably a bug

yuchanns: Great to hear that! I would close this PR and please fix it in your refactoring :)

PR #73, opened 2024-08-18
refactor(ai): streaming chunks to avoid excessive redraw.

aarnphm:
It should also cancel the job should the users quit the chat windows


aarnphm:
Tested with openai, claude I got rate limited :)

https://github.com/user-attachments/assets/c1d93713-21bb-4ed7-8589-793d990af446

PR #75, opened 2024-08-19
feat(ui): add config to turn off hints popup

ahonn:
The default hints popup is displayed every time, which is quite annoying. 
Add a new config 'hints.enabled' to control its display.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,10 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
     ---@type string | fun(): any
     list_opener = "copen",
   },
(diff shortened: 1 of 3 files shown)


atlas-comstock:
I suggested that the hint is off automatically, just written it in the readme.

ahonn:
> I suggested that the hint is off automatically, just written it in the readme.

In my opinion, hints should be enabled by default. If not, their function may be lost. 
Simply add this config for users like me who are aware that it can be selected and used with `<Leader> a a`.

yetone:
LGTM!

PR #77, opened 2024-08-19
fix: missing event_state in handler options of LLM stream function

yetone:


lua/avante/llm.lua
```diff
@@ -481,15 +481,14 @@ local active_job = nil
...
-  local event_state = nil
 
   local code_opts = {
     question = question,
     code_lang = code_lang,
     code_content = code_content,
     selected_code_content = selected_content_content,
   }
-  local handler_opts = vim.deepcopy({ on_chunk = on_chunk, on_complete = on_complete, event_state = event_state }, true)
+  local handler_opts = { on_chunk = on_chunk, on_complete = on_complete, event_state = nil }
```
yetone, line 491: @aarnphm I don't understand why deepcopy is needed here.

aarnphm: oh I don't think it need to be deep copy here

PR #145, opened 2024-08-21
added a check for git root query

MijikHna:
`git -C ...` command produces error on my machine if directory is not a git repository and `.avante_chat_history/` will be created in `fatal: not a git repository` folder.

PS: In my opinion this git command call isn't necessary as the `local current_dir = fn.fnamemodify(current_file, ":h")
` already returns the correct working  directory.


bassamsdata:
I've seen this while updating `avante.nvim`, so I wanted to mention that there's actually a more performant way to get the Git root of the current file in Neovim v0.10. Since this plugin already requires v0.10, you can use:

```lua
vim.fs.root(current_file, { ".git" })
```

PR #195, opened 2024-08-24
feat: add support of light theme

tang-hi:
Users can set a light theme by adjusting the config. This allows Avante to better match their setup if they're using a light theme, like I do. :)
Below is a preview from my own nvim
20240825_00h28m16s_grim

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -119,6 +119,10 @@ M.defaults = {
   hints = {
     enabled = true,
   },
+  --- @class AvanteThemeConfig
+  --- @field theme "light" | "dark"
+  --- @type string
+  theme = "dark",
 }
 
 ---@type avante.Config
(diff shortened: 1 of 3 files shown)


lua/avante/config.lua
```diff
@@ -119,6 +119,10 @@ M.defaults = {
   hints = {
     enabled = true,
   },
+  --- @class AvanteThemeConfig
```
aarnphm, line 122: I don't think we should do this. Users should change the target color highlight group if they are using a light theme.

I'm fine with updates on the documentation of the highlight group for people to use. Other than that, this would not be maintainable.

yetone: ‌‌‌‌Makes sense, I'll revert it first. cc @tang-hi

tang-hi: #201 looks better. I'm glad to see such a quick response! thanks

PR #199, opened 2024-08-24
feat: update max_tokens for anthropic api (3.5 sonnet)

theesfeld:
updating max_tokens for anthropic from 4096 to 8192 per the august 19th update from anthropic raising the max_tokens cap for general availability: https://docs.anthropic.com/en/release-notes/api#august-19th-2024

```
August 19th, 2024

We’ve moved 8,192 token outputs from beta to general availability for Claude 3.5 Sonnet.
```


aarnphm:
you can override it in the config settings, otherwise anytime a model changes we will have to update it.

PR #219, opened 2024-08-25
Include ARCH in asset name

rkk1995:
Currently, the makefile tries to cURL an asset that is no longer part of the latest releases.
- For example, previously the release had the file `tiktoken_core-linux-luajit.so`. 
- Now, the CPU arch is part of the release asset. See https://github.com/gptlang/lua-tiktoken/releases/tag/v0.2.2

This change adds `ARCH` so that we fetch an existing asset. 

I tested this by telling `lazy` to use my own branch and the file downloaded is correct

```
➜  build git:(84fa8df) ls -lh
Permissions Size User Date Modified Name
.rw-r--r--  3.1M rkk  25 Aug 15:50  tiktoken_core.so
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ define build_from_source
 endef
 
 define download_release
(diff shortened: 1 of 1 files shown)


yetone:
LGTM

Issue #117, opened 2024-08-20
Chat never responds


yuvals1:
![Screenshot 2024-08-20 at 17 29 39](https://github.com/user-attachments/assets/9fca5b5c-9f8a-43a5-b250-95a4e4cf6e20)
any reason why the chat won't respond? 
I exported Claude api key and switched the provider in the config to 'claude'
im using kickstart.nvim with this config for avante:

```lua
return {
  'yetone/avante.nvim',
  event = 'VeryLazy',
  build = 'make',
  opts = {
    -- add any opts here
  },
  dependencies = {
    'nvim-tree/nvim-web-devicons',
    'stevearc/dressing.nvim',
    'nvim-lua/plenary.nvim',
    {
      'grapp-dev/nui-components.nvim',
      dependencies = {
        'MunifTanjim/nui.nvim',
      },
    },
    --- The below is optional, make sure to setup it properly if you have lazy=true
    {
      'MeanderingProgrammer/render-markdown.nvim',
      opts = {
        file_types = { 'markdown', 'Avante' },
      },
      ft = { 'markdown', 'Avante' },
    },
  },
}

**Comment by aarnphm** [comment]
what platform are you on?

**Comment by yuvals1** [comment]
> what platform are you on?

macos, version 14.4, m3 pro

**Comment by yetone** [comment]
> > what platform are you on?
> 
> macos, version 14.4, m3 pro

When you just start chatting, Can you list your curl command (make sure to remove the API token)?

```bash
ps aux | grep curl
```

yuvals1:
it seems like it's not calling the api at all. no curl command is created. 
also it says that "tiktoken core in not found" although i definitley have tiktoken in the python that the neovim uses.
vim.g.python3_host_prog = vim.fn.expand '~/.virtualenvs/neovim311/bin/python3'

aarnphm:
do you have cargo setup on mac?

can you show `ls -Rl $HOME/.local/share/nvim/lazy/avante.nvim`?

yuvals1:
i do have cargo:
```bash
❯ cargo --version
cargo 1.80.1 (376290515 2024-07-16)
```

```bash
❯ ls -Rl $HOME/.local/share/nvim/lazy/avante.nvim
drwxr-xr-x@    - yuvals1 20 Aug 21:56  build
.rw-r--r--@  11k yuvals1 20 Aug 21:55  LICENSE
drwxr-xr-x@    - yuvals1 20 Aug 21:55  lua
[...]

PR #221, opened 2024-08-26
feat(provider): initial error handling on claude

Munsterberg:
Wasn't the only one running into this issue and it's more prominent due to how these companies are handling their API usage now. Can see an example in #117. Will likely have to improve this at some point and support more cases, but these seem to be the two most common problems that people run into with no feedback

### DIFF
--- a/lua/avante/llm.lua
+++ b/lua/avante/llm.lua
@@ -150,22 +150,10 @@ M.stream = function(question, code_lang, code_content, selected_content_content,
     end,
     callback = function(result)
       if result.status >= 400 then
-        local error_msg
-        if result.body then
-          local body = vim.json.decode(result.body)
-          if body and body.error then
-            if body.error.type == "insufficient_quota" then
-              error_msg =
-                "You don't have any credits or have exceeded your quota. Please check your plan and billing details."
-            elseif body.error.code == "invalid_request_error" and body.error.param == "temperature" then
-              error_msg = "Invalid temperature value. Please ensure it's between 0 and 1."
-            else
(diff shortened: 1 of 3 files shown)


lua/avante/llm.lua
```diff
@@ -148,7 +148,26 @@ M.stream = function(question, code_lang, code_content, selected_content_content,
...
+            if body.error.type == "insufficient_quota" then
+              error_msg =
+                "You don't have any credits or have exceeded your quota. Please check your plan and billing details."
+            elseif body.error.code == "invalid_request_error" and body.error.param == "temperature" then
+              error_msg = "Invalid temperature value. Please ensure it's between 0 and 1."
+            else
+              error_msg = body.error.message or "Unknown API error"
+            end
+          end
+        end
```
aarnphm, line 166: This seems too specific to anthropic to be here. Better to just handle by error code. Or to add a interface for each provider to have its own error handling.

aarnphm (changes requested):
I think we should have a `providers.on_error` function instead of this.

lua/avante/providers/init.lua
```diff
@@ -72,6 +72,7 @@ local Dressing = require("avante.ui.dressing")
 ---@field model? string
 ---@field parse_api_key fun(): string | nil
 ---@field parse_stream_data? AvanteStreamParser
+---@field on_error fun(result: table): nil
```
aarnphm, line 75:
```suggestion
---@field on_error? fun(result: table): nil
```

aarnphm (approved):
let's do it iteratively. I think this is a good start

PR #223, opened 2024-08-26
docs: add reasonable mappings

LintaoAmons:


aarnphm (approved):
For future reference we shouldn't just dump the whole config to readme, rather keeping fields that are necessary here.

But submit and toggle seems to be reasonable to set in README. :)

PR #224, opened 2024-08-26
fix: refresh colorscheme after apply diff

PJalv:
Added line to reapply colorscheme, diff highlights were not showing after applying.
This is the case for me since I have dynamic colorscheme based on filetype.
Before:
image
After:
image


lua/avante/sidebar.lua
```diff
@@ -291,6 +291,7 @@ function Sidebar:apply()
     Diff.add_visited_buffer(self.code.bufnr)
     Diff.process(self.code.bufnr)
     api.nvim_win_set_cursor(self.code.winid, { 1, 0 })
+    vim.cmd.colorscheme(vim.g.colors_name)
```
aarnphm, line 294: This seems like a hack, we should investigate further.

afaik doing this in a plugin is often frowned upon.

aarnphm (changes requested):
check comment.

aarnphm:
I will include this in my refactor of diff highlights and will keep this in mind.

aarnphm:
Can you provide a small reproducible so that I can test with this?

PJalv:
Sure, here is the autocmd I have been using. 
In hindsight my suggestion seems like a hacky way to fix. Will look further.
```lua
vim.api.nvim_create_autocmd({ "BufEnter", "BufRead", "BufWinEnter", "InsertLeave" }, {
  callback = function()
    if vim.bo.filetype == "zig" or vim.bo.filetype == "c" or vim.bo.filetype == "cpp" then
      vim.cmd.colorscheme("gruvbox")
      require("lualine").setup({ options = { theme = "gruvbox" } })
    else
      vim.cmd.colorscheme("tokyonight")
      require("lualine").setup({ options = { theme = "tokyonight" } })
    end
  end,
})
```

aarnphm:
closed due to #247

PR #242, opened 2024-08-26
fix: do not need compatible with old nvim

yetone:


lua/avante/floating_window.lua
```diff
@@ -240,15 +238,22 @@ end
 
 ---@param mode string|string[] check `:h :map-modes`
 ---@param lhs string|string[]
```
aarnphm, line 240:
```suggestion
---@param lhs string
```

aarnphm (approved):

PR #252, opened 2024-08-27
fix: remove temporary build directory before cloning in Makefile

### LINKED ISSUE #251: Failed to update
While updating I enountered this error, I suppose it errored in the middle of cloning on the last update. The folder exists and was not removed.

image

cyyyu:
fix #251


aarnphm (approved):
weird since we already remove it in L28, but sure

Issue #273, opened 2024-08-27
Cmd + K equivalent of Cursor


luisiacc:
They have this very handy functionality with Cmd + K where they select a block of text and prompt it to do an action in that block of text, and it replaces the entire text selected with the AI response.

This can be done with `AvanteAsk` of course, but this is way handier for quick changes.

I think it'd look good using something like `vim.ui.input` on the selected block and then replacing the text with the ai response, like:

![CleanShot 2024-08-27 at 02 19 20@2x](https://github.com/user-attachments/assets/cc47e437-37f2-4d15-800b-bcb0736969d4)

It looks like this on cursor:

https://github.com/user-attachments/assets/190f1c29-52bb-4f59-b176-581e4b54b20f

I don't think we need the diff for this case, it'll just be in the way.

aarnphm:
I think this makes sense, will put up a PR

luisiacc:
that was fast

yetone:
The first version of the editing mode has been implemented, and a better editing experience will be implemented later.

PR #281, opened 2024-08-27
feat: editing mode

yetone:
https://github.com/yetone/avante.nvim/issues/273

### DIFF
--- a/lua/avante/selection.lua
+++ b/lua/avante/selection.lua
@@ -1,5 +1,7 @@
 local Utils = require("avante.utils")
 local Config = require("avante.config")
+local Llm = require("avante.llm")
+local Highlights = require("avante.highlights")
 
 local api = vim.api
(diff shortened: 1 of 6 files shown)


lua/avante/selection.lua
```diff
@@ -40,35 +55,242 @@ function Selection:get_virt_text_line()
...
+          { once = true, title = "Avante" }
+        )
+        return
+      end
+      vim.defer_fn(function()
+        self:close_editing_input()
+      end, 0)
+    end
+
+    Llm.stream(input, filetype, code_content, self.selection.content, "editing", on_chunk, on_complete)
```
aarnphm, line 246: oh we can actually stream the generation here the hl the diff later?

yetone: It's not yet possible to achieve such an advanced level.

PR #284, opened 2024-08-27
fix(openai): enable image support on OpenAI platform only (fixes #282)

### LINKED ISSUE #282: Copilot integration broken
hello,

i'm seeing an error with copilot integration after recent changes. when i try to use avante for anything i see the following
```
Received error from Copilot API: Bad Request
```

when i checkout commit hash 07af0a91f29584cdabf440dd3d4a0ed677ecf63e then the error goes away and functionality is restored
```
# this fixes it
cd ~/.local/share/nvim/lazy/avante.nvim
git checkout 07af0a91f29584cdabf440dd3d4a0ed677ecf63e
<restart nvim>
```

aarnphm:

### DIFF
--- a/lua/avante/providers/azure.lua
+++ b/lua/avante/providers/azure.lua
@@ -13,7 +13,14 @@ local M = {}
 
 M.api_key_name = "AZURE_OPENAI_API_KEY"
 
-M.parse_message = O.parse_message
+M.parse_message = function(opts)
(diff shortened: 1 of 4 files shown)


oldgalileo:
Didn't immediately get how to actually use this. Would it be possible to get a follow-up doc PR?

PR #287, opened 2024-08-27
fix(copilot): make sure to set integration id

aarnphm:


lua/avante/providers/copilot.lua
```diff
@@ -137,6 +137,7 @@ M.parse_curl_args = function(provider, code_opts)
     headers = {
       ["Authorization"] = "Bearer " .. M.copilot.token.token,
       ["Content-Type"] = "application/json",
+      ["Copilot-Integration-Id"] = "vscode-chat",
```
Zamua, line 140: does this also need to be added to the headers in refresh_token?

Issue #296, opened 2024-08-27
Edit code not working properly with copilot


chgara:
When trying to edit a piece of code with the edit functionality using copilot as provider it does not edit the code properly it gives the full file prefixed by the type of it. If seen the demos of the feature working so I must asume that copilot needs other type of prompt or something like that.

Also the UI of it its kinda bugged when doing the response, but it can be that I use dressing.nvim for changing the default input.

I attach a video of it.

Thks.

https://github.com/user-attachments/assets/51a15bf0-0168-4187-b2c9-7bc73f666d98

pidgeon777:
> When trying to edit a piece of code with the edit functionality using copilot as provider it does not edit the code properly it gives the full file prefixed by the type of it. If seen the demos of the feature working so I must asume that copilot needs other type of prompt or something like that.

I confirm this issue using copilot.

yetone:
Support for the Copilot provider has been removed

PR #301, opened 2024-08-28
refactor(llm): update edit mode prompt to avoid unnecessary code block markers

corenel:
Similar to the issue noted at https://github.com/yetone/avante.nvim/issues/296, the response from OpenAI models (e.g. `gpt-4o`) isn’t producing the expected outcome.

For instance, it generates:

<pre><code>
```lua
  {{ modified code }}
```
</pre></code>

instead of providing the direct code, such as:

```markdown
{{ modified code }}
```

The revised prompt generally addresses the issue, though it still requires further polishing.

Moreover, I'm not sure whether a separate post-processing function is a better solution to this particular case.

### DIFF
--- a/lua/avante/llm.lua
+++ b/lua/avante/llm.lua
@@ -77,9 +77,10 @@ Your task is to modify the provided code according to the user's request. Follow
 7. Do not omit any parts of the code, even if they are unchanged.
 8. Maintain the SAME indentation in the returned code as in the source code
 9. When suggesting modifications:
-   a. Please keep your suggested code changes minimal, and do not include irrelevant lines in the code snippet.
-   b. Ensure that the modified code has the same indentation as the original code.
-   c. Provide the exact code snippet to be replaced using this format:
+   a. Please keep your suggested code changes minimal, and only append the diff between your suggestion versus what is PROVIDED.
+   b. Please make sure NOT TO generate CODE BLOCK when suggesting diff.
+   c. Ensure that the modified code has the SAME indentation as the original.
+   d. Provide the exact code snippet to be replaced using this format:
 
 {{modified_code}}
 
--- a/lua/avante/llm.lua
+++ b/lua/avante/llm.lua
@@ -76,8 +76,24 @@ Your task is to modify the provided code according to the user's request. Follow
 6. Preserve the original structure, indentation, and formatting of the code as much as possible.
 7. Do not omit any parts of the code, even if they are unchanged.
 8. Maintain the SAME indentation in the returned code as in the source code
+9. When suggesting modifications:
+   a. Please keep your suggested code changes minimal, and do not include irrelevant lines in the code snippet.
+   b. Ensure that the modified code has the same indentation as the original code.
+   c. Provide the exact code snippet to be replaced using this format:
 
-Remember: Your response should contain nothing but the modified code, ready to be used as a direct replacement for the original file.
+{{modified_code}}
+
+   instead of
+
+```{{language}}
+{{modified_code}}
+```
+
+10. Final check:
+   - Do not wrap the modified code in any markdown code block markers or language identifiers.
+   - Do not show the full content after these modifications.
+
+Remember: Your response should contain nothing but ONLY the modified code, ready to be used as a direct replacement for the original file.
 ]]
 
 local group = api.nvim_create_augroup("avante_llm", { clear = true })


lua/avante/llm.lua
```diff
@@ -76,8 +76,24 @@ Your task is to modify the provided code according to the user's request. Follow
 6. Preserve the original structure, indentation, and formatting of the code as much as possible.
 7. Do not omit any parts of the code, even if they are unchanged.
 8. Maintain the SAME indentation in the returned code as in the source code
+9. When suggesting modifications:
+   a. Please keep your suggested code changes minimal, and do not include irrelevant lines in the code snippet.
+   b. Ensure that the modified code has the same indentation as the original code.
+   c. Provide the exact code snippet to be replaced using this format:
```
aarnphm, line 82: I was thinking along the line of

```
9. When suggesting modification, please KEEP your suggestion to a minimal, and only append the diff between your suggestion versus what is PROVIDED.
10. Please make sure NOT TO generate CODE BLOCK when suggesting diff
```

corenel: Thanks for your review!

These lines you've provided are more precise, but it appears that a specific output format is still necessary for some corner cases.

For example, in the scenario where the user requests `use higher log level` for the given code:

```cpp
RosRaftHarnessFixture() {
  // 0. Prepare
  // 0.1. Init ROS2 node
  auto nh = yari::ros::util::RosNodeHandler::Instance();
  if (!nh->Ok()) {
    nh->InitNode(0, nullptr);
  }
  // Set up the logging
  loguru::g_stderr_verbosity = loguru::Verbosity_2;
}
```

When only the following lines are used:

```plaintext
9. When suggesting modifications:
   a. Please keep your suggested code changes minimal, and only append the diff between your suggestion versus what is PROVIDED.
   b. Please make sure NOT TO generate CODE BLOCK when suggesting diff.
```

The `gpt-4o` model still outputs the modified code within the code block.

However, when we add these extra lines, the code block disappears:

<pre><code>
   c. Ensure that the modified code has the SAME indentation as the original.
   d. Provide the exact code snippet to be replaced using this format:

{{modified_code}}

   instead of

```{{language}}
{{modified_code}}
```
</code></pre>

Prompt engineering is quite challenging, and I am curious if there are any references or examples available for this editing mode feature.

aarnphm: Hmm I wonder why gpt-4 does this. Claude doesn't seem to run into this problem at all.

yetone: This is very much as expected, gpt-4 is often disobedient compared to claude

lua/avante/llm.lua
```diff
@@ -76,8 +76,25 @@ Your task is to modify the provided code according to the user's request. Follow
 6. Preserve the original structure, indentation, and formatting of the code as much as possible.
 7. Do not omit any parts of the code, even if they are unchanged.
 8. Maintain the SAME indentation in the returned code as in the source code
+9. When suggesting modifications:
+   a. Please keep your suggested code changes minimal, and only append the diff between your suggestion versus what is PROVIDED.
```
yetone, line 80: Now in edit mode `only append the diff` doesn't work, you need to return the full code snippet to be modified.

corenel: Thanks for your review!

Sorry for that I don't quite understand your comments. This line stems from the previous conversation and simply instructs the model to apply the diff to the provided code itself before output.

Based on my test cases in some C++ & Python code, the model actually returns the complete modified code instead of just the diff part.

---

FWIW, I've conducted some tests to utilize the `openai/gpt-4o-2024-08-06` and `anthropic/claude-3.5-sonnet` models from the OpenRouter API, as well as the `gpt-4o` model from OpenAI. Unfortunately, since I haven't subscribed to Anthropic's services, I am unable to test using their official API.

Currently, the `gpt-4o` and `claude-3.5-sonnet` models almost always follows the updated instructions, but smaller models (like `gpt-4o-mini`) still tend to output code blocks.

Perhaps we should permit responses to include some text and code blocks, and then parse them in a post-processing step before integrating them into the code space? Just similar to the planning mode but in an automatic manner.

yetone: What I mean is, if the code block you have selected in edit mode is:

```python
def fib(n):
    if n < 1:
        return n
    else:
        return fib(n-1) + fib(n-2)
```

I expect the LLM to return:

```python
def fib(n):
    if n <= 1:
        return n
    else:
        return fib(n-1) + fib(n-2)
```

Rather than:

```python
    if n <= 1:
```

corenel: As mentioned earlier, the LLM, when given this prompt, still returns the complete modified code instead of just the diff in my test cases.

Does this prompt result in the LLM outputting only the diff in your tests?

yetone: Because I have run out of OpenAI quota, I am currently unable to test.

yetone:
I tested this PR with Claude and found no issues except for one indentation problem. I've merged it first, and I'll quickly fix the indentation issue next. Thanks for your contribution!

PR #306, opened 2024-08-28
fix(highlights): give user config higher priority

jkunlin:


lua/avante/highlights.lua
```diff
@@ -67,11 +67,11 @@ M.conflict_highlights = function(opts)
   local get_default_colors = function(key, hl)
     --- We will first check for the hl name. If plugins set this then we won't use the default opts[var_accesor]
```
yetone, line 68: This comment should also be changed cc @aarnphm

aarnphm: The reason I do this is that for colorscheme the can set the specific highlight group, for example the following is being used.

Then we fall back to user highlight group.

aarnphm (changes requested):
refer to comment

lua/avante/highlights.lua
aarnphm, line 68:
```suggestion
  local get_default_colors = function(key, hl)
    --- We will first check for user custom highlight. Then fallback to default name highlight.
```

aarnphm (approved):

PR #318, opened 2024-08-28
fix: remove copilot provider

yetone:


martinbjeldbak:
Thanks for this pull request! Any chance we can update the wiki to reflect this as well? https://github.com/yetone/avante.nvim/wiki#want-to-use-copilot Was just about to set it up when I found this PR 😄

PR #326, opened 2024-08-28
fix: AvanteEdit highlight for visual line mode

brentyi:
To reproduce: `<Shift-V><Leader>ae`.

Before:

https://github.com/user-attachments/assets/0076f193-5b60-4c23-ab1a-96184d830c06

After:

https://github.com/user-attachments/assets/c237dc4c-d1ef-4820-a9a0-e4ab8e0a4c69

Notes:
- The visual block mode behavior is still wrong, but that seems lower priority.
- It seems like Avante (intentionally?) only supports editing whole lines. So it could also make sense to always use the visual line behavior as an indicator for what's actually being edited.

### DIFF
--- a/lua/avante/selection.lua
+++ b/lua/avante/selection.lua
@@ -283,22 +283,29 @@ function Selection:create_editing_input()
 
   self.selection = Utils.get_visual_selection_and_range()
 
-  local end_row = self.selection.range.finish.line - 1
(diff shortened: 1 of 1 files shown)


aarnphm (approved):
LGTM. Let's iterate to fix visual block next.

PR #328, opened 2024-08-28
make `nvim-web-devicons` an optional dependency

datsfilipe:
I've noticed the plugin worked normally without it, except for the error that is thrown when it tries to call this dep and it's not installed.

I personally don't like icons in my config and don't like to have unused dependencies, so I think it's a fair PR. This plugin is amazing, btw.


aarnphm (approved):
I'm ok with this change. cc @yetone as he will have the final say in this

PR #341, opened 2024-08-28
Allow disabling automatic keymaps

will-lynas:
This adds the ability to set certain keymaps to `false` in the config (instead of a string) to disable their creation, and allows users to create their own keymaps.

This is useful in case they want to set their own `desc` or other opts.

It also exposes the functions to toggle hints and debug so they can be used in keymaps.

I'm not sure where to best place to document this would be. Feedback is appreciated!


will-lynas:
Stylua seems to be failure from an unrelated commit. See #342

aarnphm (commented):
Hmm let me propose something.

aarnphm:
See #346

PR #342, opened 2024-08-28
fix: formatting causing stylua failure

will-lynas:
Seems to have been introduced in bb5411cd9a88f6da99f877229ea1365ee95388ba


aarnphm:
oh sorry I didn't see this, but this should be fixed on main

Issue #218, opened 2024-08-25
Feature idea: Using pictures in chat input


andnig:
One of the better-working features of cursor is to simply add a screenshot to the chat input and "ask" Claude to create a component similar to the one in the image. Or change the current component according to the image or picture.

Would this be useful or possible for avante?

Great extension by the way!

aarnphm:
supporting image in a terminal is complicated, but I'm not sure if this is even in the scope of this plugin.

theesfeld:
anthropic will accept a base64 encoded image or file
https://docs.anthropic.com/en/docs/build-with-claude/vision

aarnphm:
well ye, we also support openai don't we?

wey-gu:
I am working on a local PR on this during my trip today, will debug and push later(plane about to take off now, wip)

aarnphm:
https://github.com/yetone/avante.nvim/pull/279 provides initial support for clipboard pasting.

Right now if you copy the image and save to clipboard, for claude at least we will add it to the base message.

This feature is current opt-in, so if you want to test out add the following to behaviour:

```lua
behaviour = {
  support_paste_from_clipboard = true,
}
```

aarnphm:
We will still need to support drag and drop.

wey-gu:
oh, i see your #279 already implement it for clipboard 👍🏻

aarnphm:
I will have a follow up coming up soon with support for drag and drop (currently it is kinda magical)

aarnphm:
For everyone that is listening to this thread since https://github.com/yetone/avante.nvim/commit/c635f73748926f214cbc897a0dd2da7418aa4e86 if you wish to have picture support then we recommend u to have `img-clip.nvim` setup. See readme for more information.

Also if anyone has a windows box I will need some help with powershell implementation here (currently windows is disabled, since I don't have a windows machine to test it out) cc @yuchanns

aarnphm:
I will close this as it is implemented.

PR #345, opened 2024-08-29
feat(clipboard): support windows

yuchanns:
refer to https://github.com/yetone/avante.nvim/issues/218#issuecomment-2316030910

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ Install `avante.nvim` using [lazy.nvim](https://github.com/folke/lazy.nvim):
           drag_and_drop = {
             insert_mode = true,
           },
+          -- required for Windows users
+          use_absolute_path = true,
         },
       },
     },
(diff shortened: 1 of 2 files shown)


yuchanns:
There is a problem with the relative paths of images not working in Windows. Will need some time to find out solution.

aarnphm:
is there a relative path? I thought the path would be absolute in this case no?

yuchanns:
> is there a relative path? I thought the path would be absolute in this case no?

I paste the image into the chat box and get a relative path. Is this an unexpected behavior?

yuchanns:
The img-clip.nvim supports using the absolute path https://github.com/HakonHarnes/img-clip.nvim/issues/80#issuecomment-2309812863. Windows is working with `use_absolute_path=true`.

aarnphm (approved):

PR #355, opened 2024-08-29
feat: pick conflict under cursor

b0o:
Added `cc` which picks the conflict under the cursor, if there is one.


aarnphm (approved):
when this is ready to merge lmk but it looks good

b0o:
should be good to go if it looks good to you :+1:

PR #363, opened 2024-08-29
config.lua: improve check for `img-clip`

jmoudrik:
If img clip is installed not by lazy, it is checked agains `package.loaded` in `Util.has`. But the correct key to check there is `img-clip`, not `img-clip.nvim`.

Tested.

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -187,7 +187,7 @@ function M.setup(opts)
 end
 
 M.support_paste_image = function()
-  local supported = Utils.has("img-clip.nvim")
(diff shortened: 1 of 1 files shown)


lua/avante/config.lua
```diff
@@ -187,7 +187,7 @@ function M.setup(opts)
 end
 
 M.support_paste_image = function()
-  local supported = Utils.has("img-clip.nvim")
+  local supported = Utils.has("img-clip.nvim") or Utils.has("img-clip")
```
aarnphm, line 190: This is a hack, probably should fix through Utils.has

Issue #373, opened 2024-08-29
Error when adding code patch from side bar


1wayposting:
Whenever I attempt to apply a code patch from the sidebar I'm getting this error:

Error detected while processing User Autocommands for "AvanteConflictDetected":             
Error executing lua callback: /usr/share/nvim/runtime/lua/vim/diagnostic.lua:1868: bufnr: expected nu
mber, got boolean                                                                                    
stack traceback:                                                                                     
        [C]: in function 'error'                                                                     
        vim/shared.lua: in function 'validate'                                                       
        /usr/share/nvim/runtime/lua/vim/diagnostic.lua:1868: in function 'enable'                    
        ...b/.local/share/nvim/lazy/avante.nvim/lua/avante/diff.lua:520: in function <...b/.local/sha
re/nvim/lazy/avante.nvim/lua/avante/diff.lua:518>                                                    
        [C]: in function 'nvim_exec_autocmds'                                                        
        ...b/.local/share/nvim/lazy/avante.nvim/lua/avante/diff.lua:389: in function 'parse_buffer'  
        ...b/.local/share/nvim/lazy/avante.nvim/lua/avante/diff.lua:400: in function 'process'       
        ...local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:333: in function ''              
        vim/_editor.lua: in function ''                                                              
        vim/_editor.lua: in function <vim/_editor.lua:0>  
        
The diff's show up, but for every single change in the file i make after that point I get the error again. Also any of the bindings for accepting the code won't work. 

![image](https://github.com/user-attachments/assets/62af6749-8bf2-4e07-a5ac-7df319f671a1)

PR #374, opened 2024-08-29
Fix for: Error when adding code patch from side bar #373

1wayposting:
This fixes this error:

https://github.com/yetone/avante.nvim/issues/373

I don't know if it causes other issues, but I haven't come across any while using this fix.

### DIFF
--- a/lua/avante/diff.lua
+++ b/lua/avante/diff.lua
@@ -516,7 +516,7 @@ function M.setup()
     group = augroup,
     pattern = "AvanteConflictDetected",
     callback = function(ev)
-      vim.diagnostic.enable(false, { bufnr = ev.buf })
+      vim.diagnostic.enable(namespace_id, bufnr, opts)
       if is_inlay_enable then
         previous_inlay = vim.lsp.inlay_hint.is_enabled({ bufnr = ev.buf })
         vim.lsp.inlay_hint.enable(false, { bufnr = ev.buf })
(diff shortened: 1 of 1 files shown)


lua/avante/diff.lua
```diff
@@ -516,7 +516,7 @@ function M.setup()
     group = augroup,
     pattern = "AvanteConflictDetected",
     callback = function(ev)
-      vim.diagnostic.enable(false, { bufnr = ev.buf })
```
aarnphm, line 519: what neovim version are you on?

This is the current API for 0.10 and 0.11 afaik

aarnphm (changes requested):
see comments

1wayposting:
image

aarnphm:
Screenshot 2024-08-29 at 14 11 34

I'm running 0.10.1 and it seems to work?

lua/avante/diff.lua
aarnphm: https://github.com/neovim/neovim/blob/6ed530baa5050c4b63acd9d62c0884dd599c06fe/runtime/lua/vim/diagnostic.lua#L2010

I don't really know how to reproduce this error

aarnphm: Do you know what hashh you are building against?

aarnphm: from my end https://github.com/NixOS/nixpkgs/blob/05d8b74b04f1b16898c4b81da8eef65ac432e4e3/pkgs/by-name/ne/neovim-unwrapped/package.nix#L69

I suggest trying again with 0.10.1

1wayposting:
Ok yeah installed 0.10.1 and it works. Sorry about this, i saw this:

'avante.nvim is currently only compatible with Neovim 0.10.0 or later. Please ensure that your Neovim version meets these requirements before proceeding.'

and assumed i'd be alright

Thank you.

PR #403, opened 2024-08-30
fix(copilot): credentials

### LINKED ISSUE #390: bug: You must setup copilot with either copilot.lua or copilot.vim
Hi, I am getting this error when adding the new copilot provider:

```
Failed to run `config` for avante.nvim
...e/nvim/lazy/avante.nvim/lua/avante/providers/copilot.lua:150: You must setup copilot with either copilot.lua or copilot.vim
```

Copilot is already active and working, though.
Do I need to do something special for configuring it?

Thank you in advance!

yuchanns:
copilot broke on my side. I use copilot.vim.

Maybe also fix #390

### DIFF
--- a/lua/avante/providers/copilot.lua
+++ b/lua/avante/providers/copilot.lua
@@ -54,7 +54,10 @@ H.get_oauth_token = function()
 
   local yason = Path:new(config_dir):joinpath("github-copilot", "hosts.json")
   if not yason:exists() then
-    error("You must setup copilot with either copilot.lua or copilot.vim", 2)
+    yason = Path:new(config_dir):joinpath("github-copilot", "apps.json")
+    if not yason:exists() then
+      error("You must setup copilot with either copilot.lua or copilot.vim", 2)
+    end
   end
   return vim
     .iter(


lua/avante/providers/copilot.lua
```diff
@@ -54,7 +54,10 @@ H.get_oauth_token = function()
 
   local yason = Path:new(config_dir):joinpath("github-copilot", "hosts.json")
   if not yason:exists() then
-    error("You must setup copilot with either copilot.lua or copilot.vim", 2)
+    yason = Path:new(config_dir):joinpath("github-copilot", "apps.json")
```
aarnphm, line 57: ah i see

aarnphm: can you update with this snippet

```lua
  --- hosts.json (copilot.lua), apps.json (copilot.vim)
  ---@type Path[]
  local paths = vim.iter({ "hosts.json", "apps.json" }):fold({}, function(acc, path)
    local yason = Path:new(config_dir):joinpath("github-copilot", path)
    if yason:exists() then
      table.insert(acc, yason)
    end
    return acc
  end)
  if #paths == 0 then
    error("You must setup copilot with either copilot.lua or copilot.vim", 2)
  end

  local yason = paths[1]
```

PR #407, opened 2024-08-30
feat: tokenizers

aarnphm:
rust go vroom vroom, we can encode prompt and stuff now very nice very demure


aarnphm:
@yuchanns can you test with windows to see if this works lol

yuchanns:
Yes. The build script itself on Windows works, while the build option in the configuration does not work.

According to the [docs](https://lazy.folke.io/developers#building) of `lazy.nvim`, the build option should be a string.

crates/lua-tokenizers/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "lua-tokenizers"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+mlua = { version = "0.9.9", features = ["luajit", "async", "lua54"] }
```
yuchanns, line 7: Can we use `0.10.0-beta.1` as it provides better async support by removing lifetime marks? You don't have to fight against the disgusting `'lua` identifiers that are spreading everywhere.

lua/avante/init.lua
```diff
@@ -221,6 +253,48 @@ setmetatable(M.toggle, {
...
+    cmd = { "sh", "-c", ("make -C %s"):format(build_directory) }
+  elseif os_name == "windows" then
+    cmd = {
+      "powershell",
+      "-ExecutionPolicy",
+      "Bypass",
+      "-File",
+      "Build.ps1",
+      "-WorkingDirectory",
+      build_directory:gsub("/", "\\"),
```
yuchanns, line 279:
```
build_directory:gsub("/", "\\")[0], -- the row result is in the format of {"path", pos}
```

crates/lua-tokenizers/Cargo.toml
aarnphm: https://lazy.folke.io/spec#spec-setup it seems like you can set build to boolean?

yuchanns: Nope. I got error: `attempt to index local build (a boolean value)`

yuchanns:
Well, it still fails on the build. Need more investigation. If I manually dispatch `AvanteBuild` command, it success. Fails if delegated to the build option.

README.md
```diff
@@ -27,14 +27,10 @@ https://github.com/user-attachments/assets/86140bfd-08b4-483d-a887-1b701d9e37dd
   "yetone/avante.nvim",
   event = "VeryLazy",
   lazy = false,
+  build = true, -- optional, this requires cargo.
```
yuchanns, line 30: This works if I set it to `build = "build.lua"`, but neither `build = true` nor `build = "AvanteBuild"` does. Regarding the latter, I suspect it is because the custom commands have not yet been loaded into Lua.

yuchanns: I retract my earlier statement. `build = "build.lua"` is also ineffective; it simply prints the command without producing an error.

yuchanns: I managed to find out the error information:

 > -File 形式参数的实际参数"Build.ps1"不存在。请提供现有".ps1"文件的路径，作为 -File 形式参数的一个实际参数。

Translation: 

> The actual parameter for the -File form parameter "Build.ps1" does not exist. Please provide the path to an existing ".ps1" file as an actual parameter for the -File form parameter.

So the problem is the path.

yuchanns: I try to print the build_directory and get nil value.

yuchanns: Now it works with manually dispatch AvanteBuild, but still buggy with build option

aarnphm: it seems like for lazyvim to stream the output from build.lua

crates/lua-tokenizers/Cargo.toml
aarnphm: TIL

README.md
aarnphm: thanks for the feedback I will updated accordingly.

aarnphm:
Does AvanteBuild work for you? I think we do need this for vim-plug and adjacent.

README.md
aarnphm: Actually I think I know why it doesn't stream the log. Might have to do with `vim.system` being async.

aarnphm: https://github.com/folke/lazy.nvim/blob/077102c5bfc578693f12377846d427f49bc50076/lua/lazy/manage/task/plugin.lua#L72

crates/lua-tokenizers/Cargo.toml
aarnphm: but shouldn't we wait til 0.10.0 release?

afaik it doesn't affect us anything, since the crate for now will just be a simple binding between hf/tokenizers and tiktoken_core

yuchanns:
AvanteBuild works after startup. But not with build option.

crates/lua-tokenizers/Cargo.toml
yuchanns: That is ok if you didnt run into lifetime issues.

aarnphm:
well I guess we have to make build sync

aarnphm:
I have now confirmed that `:AvanteBuild` should work for build.

yuchanns:
It still does not work with `build = "AvanteBuild"`.

aarnphm:
no it has to be `build=":AvanteBuild"` (`build="AvanteBuild"` is incorrect syntax for lazy.nvim)

yuchanns:
Confirmed.

aarnphm:
@yuchanns will need your help testing on windows here. Everything is ready to go.

README.md
```diff
@@ -90,42 +85,47 @@ Plug 'HakonHarnes/img-clip.nvim'
 Plug 'zbirenbaum/copilot.lua'
 
 " Yay
-Plug 'yetone/avante.nvim'
+Plug 'yetone/avante.nvim', { 'branch': 'main', 'do': ':AvanteBuild', 'on': 'AvanteAsk' }
 ```
 
+> [!important]
+>
+> For `tiktoken` to work, make sure to call `require('avante_lib').load()` somewhere when entering the editor.
```
aarnphm, line 93:
```suggestion
> For `avante.tokenizers` to work, make sure to call `require('avante_lib').load()` somewhere when entering the editor.
```

crates/avante-tokenizers/README.md
```diff
@@ -0,0 +1 @@
+A simple unified crates to unify hf/tokenizers and tiktoken_core
```
aarnphm, line 1:
```suggestion
A simple unified crates to unify hf/tokenizers and tiktoken-rs
```

lua/avante/providers/claude.lua
```diff
@@ -28,8 +29,10 @@ M.parse_message = function(prompt_opts)
     local user_prompt_obj = {
       type = "text",
       text = user_prompt,
-      cache_control = { type = "ephemeral" },
     }
+    if Utils.tokens.calculate_tokens(user_prompt_obj.text) > 1024 then
+      user_prompt_obj.cache_control = { type = "ephemeral" }
```
yetone, line 34: Why do we have to put this back? @aarnphm

PR #455, opened 2024-09-02
feat: automatic suggestion (smart tab)

yetone:


lua/avante/suggestion.lua
```diff
@@ -0,0 +1,362 @@
...
+function Suggestion:destroy()
+  self:stop_timer()
+  self:reset()
+  self:delete_autocmds()
+  api.nvim_del_namespace(SUGGESTION_NS)
+end
+
+function Suggestion:setup_mappings()
+  local mappings = {
+    [{ "<Tab>" }] = "select",
```
aarnphm, line 52: hmm we shouldn't do this I think, we should have integration with cmp.mappings as we might have custom mappings of this.

yetone: I don't think we need to forcefully bind with cmp at this stage. I have already made the shortcut keys configurable.

lua/avante/api.lua
```diff
@@ -46,7 +46,7 @@ M.edit = function(question)
 end
 
 M.refresh = function()
```
aarnphm, line 48: lets also add a `M.suggestion` API which we expose the instance here for integration with mappings?

```lua

M.suggestion = function(motion)
...
end
```

lua/avante/suggestion.lua
```diff
@@ -0,0 +1,401 @@
...
+
+function Suggestion:destroy()
+  self:stop_timer()
+  self:reset()
+  self:delete_autocmds()
+  api.nvim_del_namespace(SUGGESTION_NS)
+end
+
+function Suggestion:setup_mappings()
+  if Config.mappings.suggestion and Config.mappings.suggestion.accept then
```
aarnphm, line 53: hmm should we also check behaviour `auto_set_keymap` here as well?

lua/avante/llm.lua
```diff
@@ -88,8 +88,36 @@ Your task is to modify the provided code according to the user's request. Follow
...
+2. You must follow this json format when suggesting modifications:
+
+[
+  {
+    "row": ${row},
+    "col": ${column},
+    "content": "Your suggested code here"
+  }
+]
+
```
aarnphm, line 105: note to @aarnphm: do structured outputs since claude, cohere, and openai all support this.

lua/avante/utils/init.lua
```diff
@@ -338,13 +340,15 @@ function M.debug(msg, opts)
...
   if type(msg) == "string" then
-    M.notify(msg, opts)
+    print(msg)
+    -- M.notify(msg, opts)
   else
-    opts.lang = "lua"
-    M.notify(vim.inspect(msg), opts)
+    print(vim.inspect(msg))
+    -- opts.lang = "lua"
+    -- M.notify(vim.inspect(msg), opts)
```
aarnphm, line 351: revert this back?

yetone: notify is too annoying, can it be changed to print so it won't interrupt operations?

lua/avante/llm.lua
yetone: Other scenarios may not be suitable for structured output because, in other scenarios, to optimize user experience, we provide real-time streaming feedback to the user interface.

aarnphm: you are right, since tool call and structured outputs are actually a few roundtrip

lua/avante/utils/init.lua
aarnphm: This is debug right? users don't really use this except us?

aarnphm: I guess noice.nvim doesn't actually block the buffer, but normal vim.notify do block.

lua/avante/api.lua
```diff
@@ -45,8 +45,22 @@ M.edit = function(question)
...
+    return
+  end
+  suggestion:suggest()
+end
+
+---@return avante.Suggestion | nil
+M.get_suggestion = function()
+  local _, _, suggestion = require("avante").get()
+  return suggestion
+end
```
aarnphm, line 60: Hmm this seems redundant.

We will probably only need get_suggestion, as M.suggest seems like a sugar of `M.get_suggestion():suggest()`

lua/avante/init.lua
```diff
@@ -185,7 +188,7 @@ H.autocmds = function()
   api.nvim_create_autocmd("VimResized", {
     group = H.augroup,
     callback = function()
-      local sidebar, _ = M.get()
+      local sidebar = M.get()
```
aarnphm, line 191:
```suggestion
      local sidebar, _, _ = M.get()
```

lua/avante/init.lua
```diff
@@ -288,7 +298,7 @@ M.toggle.hint = H.api(Utils.toggle_wrap({
 setmetatable(M.toggle, {
   __index = M.toggle,
   __call = function()
-    local sidebar, _ = M.get()
+    local sidebar = M.get()
```
aarnphm, line 301:
```suggestion
    local sidebar, _, _ = M.get()
```

lua/avante/init.lua
yetone: It's really uncomfortable to have to add a `, _` every time an additional return value is added.

lua/avante/init.lua
yetone: Because of the characteristics of Lua, not adding `_, _` will yield the same result. This way, it will reduce the hassle of having to add empty return values each time in the future.

lua/avante/init.lua
aarnphm: did lua unpack value automatically?

aarnphm: but then this will always return a tuple of three items right?

aarnphm: oh wow it will unpack based on the local variable set (TIL)

```lua
local sidebar = M.get()
-- returns:
-- {
--   code = {
--     bufnr = 0,
--     winid = 0
--   },
--   id = 1,
--   registered_cmp = true,
--   winids = {
--     input = 0,
--     result = 0,
--     selected_code = 0
--   },
--   <metatable> = {
--     __index = {...}
--   }
-- }
```

lua/avante/api.lua
```diff
@@ -35,7 +35,7 @@ end
 
 ---@param question? string
 M.edit = function(question)
-  local _, selection = require("avante").get()
+  local _, selection, _ = require("avante").get()
```
aarnphm, line 38:
```suggestion
  local _, selection = require("avante").get()
```

aarnphm: this works btw I just tested it

georgeguimaraes:
Hi, do you plan on having avante as a nvim-cmp source, like https://github.com/zbirenbaum/copilot-cmp does?

aarnphm:
not right now per say, as there are still some rough edges.

The API are there, so you can already incorporate it with Tab mapping, wrt to cmp source I'm not sure if the roundtrip is fast enough to be supplied as cmp source yet.

Issue #464, opened 2024-09-02
Allow configuring breakindent


will-lynas:
<img width="452" alt="image" src="https://github.com/user-attachments/assets/0d8086d0-d6eb-42bf-9506-deb08a0c66c3">

[render-markdown.nvim](https://github.com/MeanderingProgrammer/render-markdown.nvim) doesn't seem to understand the `breakindent` option when highlighting code blocks. This is presumably an upstream issue, but it would be good to be able to configure the `breakindent` option anyway. And the other options in `base_win_options` too: https://github.com/yetone/avante.nvim/blob/8a683b053bdb5cd871d9ba6699311aae04fb616e/lua/avante/sidebar.lua#L348-L366

aarnphm:
it is not a render-markdown.nvim bug persay,

Would be nice if we can use `guides.markers.veritical` in conjunction with showbreak, but there is no way afaik to style the color without accidentally breaking users colorscheme.

We can set showbreak to empty string

mcchrish:
Can this be done just by `ftplugin` or FileType autocommand on `Avante` filetype?

will-lynas:
Sounds sensible

MeanderingProgrammer:
An alternative to the `ftplugin` approach is to set it in the `render-markdown` plugin configuration:

```lua
require('render-markdown').setup({
    file_types = { 'markdown', 'Avante' },
    win_options = {
        breakindent = { default = true, rendered = false },
    },
})
```

Manipulating window options when rendering is already something my plugin needs to do to set the conceal behavior consistently hence the `win_options` field, but its left completely generic and can be used to manipulate all option values.

will-lynas:
That's a nice solution but I think a renderer-agnostic method would be better.

MeanderingProgrammer:
That's fair, depends on your preferred behavior.

For instance if you remove this particular renderer do you want `breakindent` to be re-enabled automatically or do you want that to be independent and update `ftplugin/Avante` separately.

aarnphm:
If you don't want to see this right now, you can try setting wrap=false in configuration (see https://github.com/yetone/avante.nvim/blob/f7c3003f4cff3620665f31f2732972ef14cdf9e3/lua/avante/sidebar.lua#L659)

But I agree with meandering here, it shouldn't rely on the renderer plugin.

will-lynas:
Closing this since the autocmd solution works well for me. Thanks @mcchrish !

PR #490, opened 2024-09-03
Expose window options

will-lynas:
See #464


will-lynas:
Closing as not needed

Issue #122, opened 2024-08-21
milestone... making llm aware of entire code base


theesfeld:
Perhaps by using the git repository information, it would be great for the llm to have awareness or knowledge of the entire code base, to know about custom imports, functions in other files, utility code files, etc. 

I know this is far fetched, so not really an issue but more of a dream :)

yetone:
‌We can refer to the implementation plan of Aider's repo map.

pidgeon777:
It would be a big feature, for sure!

theesfeld:
it would be game changing. this is the feature no one has but everyone wants

aarnphm:
my guess is that this would only works with claude or models that support 1M+ tokens

theesfeld:
just wanted to throw this out there
https://github.com/Doriandarko/claude-engineer

mrloop:
Started work on [repo-map.nvim](https://github.com/mrloop/repo-map.nvim)

aarnphm:
from #456 too track open buffer as an option.

SynapticSage:
bump! also very interested in this feature.

I do love the idea of defaulting to open buffers. though would be nice to have `/drop` and `/add` file features like aider.

pidgeon777:
Any news?

yetone:
Tend to use the Agent approach to automatically obtain context.

PR #496, opened 2024-09-03
feat: repo map

yetone:
https://github.com/yetone/avante.nvim/issues/122


lua/avante/utils/init.lua
```diff
@@ -2,6 +2,8 @@ local api = vim.api
 local fn = vim.fn
 local lsp = vim.lsp
 
+local RepoMap = require("avante.utils.repo_map")
```
aarnphm, line 5: oh you can just use `M.repo_map.<func>` since L12 to L21

aarnphm: also you can add the hints to it as well 😄

lua/avante/utils/init.lua
```diff
@@ -514,4 +516,32 @@ function M.debounce(func, delay)
   end
 end
 
+function M.get_project_root()
+  local Root = require("avante.utils.root")
+  return Root.git() or Root.cwd()
```
aarnphm, line 521:
```suggestion
	return M.root.get()
```

This should concat from lsp -> git -> cwd

lua/avante/api.lua
```diff
@@ -115,6 +118,11 @@ M.refresh = function()
   sidebar:render()
 end
 
+M.build_project_context = function()
+  local repo_map = RepoMap.repo_map(Utils.get_project_root())
+  Path.repo_map.save(vim.api.nvim_get_current_buf(), repo_map)
```
aarnphm, line 123:
```suggestion
  local repo_map = Utils.repo_map.repo_map(Utils.root.get())
  Path.repo_map.save(vim.api.nvim_get_current_buf(), repo_map)
```

lua/avante/init.lua
```diff
@@ -43,6 +43,11 @@ H.commands = function()
     function(opts) require("avante.api").edit(vim.trim(opts.args)) end,
     { desc = "avante: edit selected block", nargs = "*" }
   )
+  cmd(
+    "BuildProjectContext",
+    function() require("avante.api").build_project_context() end,
+    { desc = "avante: build project context", nargs = "*" }
+  )
```
aarnphm, line 50: I think we can build the project context asynchronously?

lua/avante/utils/repo_map.lua
```diff
@@ -0,0 +1,526 @@
+local parsers = require("nvim-treesitter.parsers")
+
+---@class avante.utils.repo_map
+local RepoMap = {}
+
+local dependencies_queries = {
+  lua = [[
```
aarnphm, line 7: hmm, are we down to maintain query for all supported language?

Then I suggest we should use query file instead?

yetone:
This PR is still in rapid iteration and is not yet ready for review.

lua/avante/utils/repo_map.lua
yetone: @aarnphm My brain is limited, I need to implement quickly, and think about project structure issues after implementation is complete.

jmmarotta:
You could probably use the [aider queries](https://github.com/paul-gauthier/aider/tree/main/aider/queries) for treesitter.

aarnphm:
we are iterating a lot atm, and will consider refactor the query later.

yetone:
> You could probably use the [aider queries](https://github.com/paul-gauthier/aider/tree/main/aider/queries?rgh-link-date=2024-09-05T19%3A57%3A44Z) for treesitter.您可能可以使用树木管理员的[辅助查询](https://github.com/paul-gauthier/aider/tree/main/aider/queries)。

aider's queries return all class/function definitions in the file, but I only need the externally exposed classes/functions (similar to a C header file), so their code doesn't suit my requirements.

yetone:
Here's an update for everyone: Actually, this PR was completed a long time ago. The functionality tests are all normal and the results are very good, but it consumes a large number of tokens (to test this PR, my Anthropic quota has already been exhausted 😢). We are currently considering how to optimize this aspect.

SchnozzleCat:
> Here's an update for everyone: Actually, this PR was completed a long time ago. The functionality tests are all normal and the results are very good, but it consumes a large number of tokens (to test this PR, my Anthropic quota has already been exhausted 😢). We are currently considering how to optimize this aspect.

Is the consumed token amount similar to what is consumed when using something like Aider (and adding all files in a project)?

jmmarotta:
> > Here's an update for everyone: Actually, this PR was completed a long time ago. The functionality tests are all normal and the results are very good, but it consumes a large number of tokens (to test this PR, my Anthropic quota has already been exhausted 😢). We are currently considering how to optimize this aspect.
> 
> Is the consumed token amount similar to what is consumed when using something like Aider (and adding all files in a project)?

The core issue seems to be that the current implementation is using the entire repomap instead of focusing on the most relevant tags. A potential solution would involve finding a way to rank the tags produced by tree-sitter.

In Python ecosystems, libraries like NetworkX (used in aider) can represent these tags as nodes in a graph and then apply algorithms like PageRank to rank them effectively.

aarnphm:
afaik nvim has binding to python, but it is kinda slow.

Maybe we can do reranking through colbert from rust/c++ or sth.

jmmarotta:
Pagerank might work better than colbert as it's well-suited for treesitter's tag structure, but using rust/c++ is a great idea.

[petgraph](https://github.com/petgraph/petgraph) might work well

aarnphm:
> [petgraph](https://github.com/petgraph/petgraph) might work well

Thanks for the knowledge sharing, will take a look into this

yetone:
> > Is the consumed token amount similar to what is consumed when using something like Aider (and adding all files in a project)?
> 
> 
> 
> The core issue seems to be that the current implementation is using the entire repomap instead of focusing on the most relevant tags. A potential solution would involve finding a way to rank the tags produced by tree-sitter.
> 
> 
> 
> In Python ecosystems, libraries like NetworkX (used in aider) can represent these tags as nodes in a graph and then apply algorithms like PageRank to rank them effectively.

In my local tests, aider's repo map tends to be inaccurate, bringing in unrelated files while ignoring related ones.

jmmarotta:
> In my local tests, aider's repo map tends to be inaccurate, bringing in unrelated files while ignoring related ones.

PageRank likely overweights common utilities. For a more localized context BFS might be the best.

PR #483, opened 2024-09-03
chore(build): support download Windows artifacts

yuchanns:
follow https://github.com/yetone/avante.nvim/issues/418

cc @aarnphm

PR #502, opened 2024-09-04
ci(release): archive Windows artifacts with zip

yuchanns:
Prerequisites for #483.

Also please release a new version after merging. @yetone

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -61,39 +61,54 @@ jobs:
       - name: Build all crates
         run: cargo build --release --features ${{ matrix.feature }}
       - name: Handle binaries
+        if: ${{ matrix.os != 'windows-latest' }}
         shell: bash
         run: |
           mkdir -p results
           if [ "${{ matrix.os }}" == "ubuntu-latest" ]; then
             OS="linux"
             EXT="so"
-          elif [ "${{ matrix.os }}" == "macos-latest" ]; then
+          else
             OS="macOS"
             EXT="dylib"
-          else
-            OS="windows"
-            EXT="dll"
           fi
-
-          if [ "${{ matrix.os }}" == "windows-latest" ]; then
-            cp target/release/avante_templates.$EXT results/avante_templates.$EXT
-            cp target/release/avante_tokenizers.$EXT results/avante_tokenizers.$EXT
-          else
-            cp target/release/libavante_templates.$EXT results/avante_templates.$EXT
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -61,39 +61,54 @@ jobs:
...
+        run: |
+          New-Item -ItemType Directory -Force -Path results
+
+          Copy-Item -Path "target\release\avante_templates.dll" -Destination "results\avante_templates.dll"
+          Copy-Item -Path "target\release\avante_tokenizers.dll" -Destination "results\avante_tokenizers.dll"
+
+          Set-Location -Path results
+
+          $dllFiles = Get-ChildItem -Filter "*.dll" | Select-Object -ExpandProperty Name
+          Compress-Archive -Path $dllFiles -DestinationPath "avante_lib-${{ matrix.os }}-${{ matrix.feature }}.zip"
```
aarnphm, line 92: why do need a separate branch for windows? afaik the runner also accept bash right?

aarnphm (commented):
qq

.github/workflows/release.yaml
yuchanns: FYI: https://github.com/yetone/avante.nvim/pull/483#issuecomment-2326707354

aarnphm: https://github.com/yetone/avante.nvim/actions/runs/10684672838/job/29615675352

yuchanns: Nope. External tools will be required to unarchive `tar.gz` in Windows. So we need to use zip and bash on Windows doesn't support zip in GHA.

yuchanns: Powershell doesn't have builtin command to unarchive `tar.gz`.

aarnphm: haizzz windows 😮‍💨

aarnphm: then it seems to me for L79 we should just zip it with tar right?

yuchanns: Neither `tar` nor `tar.gz` is supported. Zip the raw artifacts only.

aarnphm: i mean 

```yaml
if [ "${{ matrix.os }}" == "windows-latest" ]; then
  zip ... .zip
else
  tar
```

yuchanns: zip is not supported when `shell: bash`

aarnphm (approved):
LGTM, as I don't have access to Windows, but it seems correct to me 😄

PR #518, opened 2024-09-04
feat: Free storage for not existing files

GNITOAHC:
Add support to clean unused storage for not existing files.
```lua
cmd("Clean", function()
  local storage_path = require("avante.path").get_storage_path()
  for _, file in ipairs(vim.fn.readdir(storage_path)) do
    if file:match("%.json$") then
      local sourcefile = require("avante.path").get_path_from_filename(file)
      if not vim.loop.fs_stat(sourcefile) then
        vim.notify("Deleting " .. storage_path .. "/" .. file)
        vim.fn.delete(storage_path .. "/" .. file)
      end
    end
  end
end, { desc = "avante: clean up" })
```

### DIFF
--- a/lua/avante/init.lua
+++ b/lua/avante/init.lua
@@ -67,6 +67,18 @@ H.commands = function()
       return vim.tbl_filter(function(key) return key:find(prefix, 1, true) == 1 end, Config.providers)
     end,
   })
+  cmd("Clean", function()
+    local storage_path = require("avante.path").get_storage_path()
+    for _, file in ipairs(vim.fn.readdir(storage_path)) do
+      if file:match("%.json$") then
+        local sourcefile = require("avante.path").get_path_from_filename(file)
+        if not vim.loop.fs_stat(sourcefile) then
+          vim.notify("Deleting " .. storage_path .. "/" .. file)
+          vim.fn.delete(storage_path .. "/" .. file)
+        end
+      end
+    end
+  end, { desc = "avante: clean up" })
 end
 
 H.keymaps = function()
--- a/lua/avante/init.lua
+++ b/lua/avante/init.lua
@@ -67,7 +67,7 @@ H.commands = function()
       return vim.tbl_filter(function(key) return key:find(prefix, 1, true) == 1 end, Config.providers)
     end,
   })
-  cmd("Clean", function() require("avante.path").clean() end, { desc = "avante: clean all chat history" })
+  cmd("Clear", function() require("avante.path").clear() end, { desc = "avante: clear all chat history" })
 end
 
 H.keymaps = function()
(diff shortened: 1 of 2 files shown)


lua/avante/init.lua
```diff
@@ -67,6 +67,18 @@ H.commands = function()
       return vim.tbl_filter(function(key) return key:find(prefix, 1, true) == 1 end, Config.providers)
     end,
   })
+  cmd("Clean", function()
```
aarnphm, line 70: I think clean should just remove everything

lua/avante/path.lua
```diff
@@ -140,4 +148,8 @@ P.available = function() return templates ~= nil end
 
 P.clear = function() P.cache_path:rm({ recursive = true }) end
 
+P.get_path_from_filename = H.path_from_filename
```
aarnphm, line 151: The how purpose of H class (helper class) is to not expose it outside.

lua/avante/init.lua
aarnphm: also I think this should just call `require("avante.path").clear()` and the logic should reside there

GNITOAHC: > I think clean should just remove everything

How about a separate command that cleans none existing files?

aarnphm: wdym by none existing files? all files under storage path are json file for chat history.
So it is either just remove all of them or not remove it?

If you want to just clear the chat then use `/clear` in the input buffer.

GNITOAHC: If someone chat with a file `tmp.lua` then remove the file, the chat history will still be stored on the machine. So `Clean` command's purpose is to remove those chat history.

aarnphm: does it matter though? it is going to be like 2kb of json?

If you want to delete the file one can surely just cd to the directory and clear it.

I don't see the point of having this fine grained control

GNITOAHC: Yeah, you're right. I think just remove everything is fine.

lua/avante/path.lua
```diff
@@ -140,4 +140,12 @@ P.available = function() return templates ~= nil end
 
 P.clear = function() P.cache_path:rm({ recursive = true }) end
 
+P.clean = function()
```
aarnphm, line 143: just use clear

GNITOAHC: Sorry for misunderstanding.

lua/avante/path.lua
```diff
@@ -138,6 +138,13 @@ end
...
 P.available = function() return templates ~= nil end
 
-P.clear = function() P.cache_path:rm({ recursive = true }) end
+P.clear = function()
+  P.cache_path:rm({ recursive = true })
+  local storage_path = P.history_path:absolute()
+  for _, file in ipairs(vim.fn.readdir(storage_path)) do
+    if file:match("%.json$") then vim.fn.delete(storage_path .. "/" .. file) end
+  end
+  vim.notify("Clear successful!")
```
aarnphm, line 147:
```suggestion
	P.history_path:rm({recursive = true})
```

aarnphm (approved):
LGTM thanks for going through all the trouble.

PR #539, opened 2024-09-05
remove prefix msg when build because stdout/stderr can be table object

ryul99:
In my case, stderr can be a table value when the build failed. I remove the prefix string because the log level is already set in https://github.com/yetone/avante.nvim/blob/main/lua/avante/utils/init.lua#L301-L315

img

### DIFF
--- a/lua/avante/api.lua
+++ b/lua/avante/api.lua
@@ -73,9 +73,9 @@ M.build = function(opts)
       local output = stdout
       if #output == 0 then
         table.insert(output, "")
-        Utils.info("outputs: " .. output)
+        Utils.info(output)
       else
-        Utils.error("error: " .. stderr)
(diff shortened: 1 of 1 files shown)


lua/avante/api.lua
```diff
@@ -73,9 +73,9 @@ M.build = function(opts)
       local output = stdout
       if #output == 0 then
         table.insert(output, "")
-        Utils.info("outputs: " .. output)
+        Utils.info(output)
       else
-        Utils.error("error: " .. stderr)
```
aarnphm, line 78: oh this is a bug on my end, it should be `vim.inspect` instead. Can you update to that?

aarnphm (changes requested):
one comment.

Issue #453, opened 2024-09-02
Install Failed via vim-plug


Kaoruha:
I installed the lib via vim-plug, but got the following error log.

1   Updated. Elapsed time: 2.031265 sec.                                                                                                                                                                                                                                   
  1 [====================x]
  2                               
  3 - Finishing ... Done!
  4 x Post-update hook for avante.nvim ... Vim:E492: Not an editor command: AvanteBuild
how to fix~

snowphone:
Does `Plug 'yetone/avante.nvim', { 'branch': 'main', 'do': { -> avante#build() }, 'on': 'AvanteAsk' }` work on the latest vim-plug & neovim 0.10.1?
When I try it, it raises an error like this: 

```
x Post-update hook for avante.nvim ... Vim(source):E5113: Error while calling lua chunk: /home/sixtyfive/.vim/plugged/avante.nvim/plugin/avante.lua:13: module 'avante.clipboard' not found:
	no field package.preload['avante.clipboard']
	no file './avante/clipboard.lua'
	no file '/home/runner/work/neovim/neovim/.deps/usr/share/luajit-2.1/avante/clipboard.lua'
	no file '/usr/local/share/lua/5.1/avante/clipboard.lua'
	no file '/usr/local/share/lua/5.1/avante/clipboard/init.lua'
	no file '/home/runner/work/neovim/neovim/.deps/usr/share/lua/5.1/avante/clipboard.lua'
	no file '/home/runner/work/neovim/neovim/.deps/usr/share/lua/5.1/avante/clipboard/init.lua'
	no file './avante/clipboard.so'
	no file '/usr/local/lib/lua/5.1/avante/clipboard.so'
	no file '/home/runner/work/neovim/neovim/.deps/usr/lib/lua/5.1/avante/clipboard.so'
	no file '/usr/local/lib/lua/5.1/loadall.so'
	no file './avante.so'
	no file '/usr/local/lib/lua/5.1/avante.so'
	no file '/home/runner/work/neovim/neovim/.deps/usr/lib/lua/5.1/avante.so'
	no file '/usr/local/lib/lua/5.1/loadall.so'
stack traceback:
	[C]: in function 'require'
	/home/sixtyfive/.vim/plugged/avante.nvim/plugin/avante.lua:13: in main chunk
```

aarnphm:
did you call `require("avante").setup()`? please read the readme

Kaoruha:
Still not work
When PlugInstall, this occured

[...]

PR #548, opened 2024-09-06
Fix list join argument in avante#build function

snowphone:
Changes:
- Correct the `join` function arguments to be a list by adding square brackets around them
- This ensures proper concatenation of strings from `luaeval("require('avante_lib').load()")` and `luaeval("require('avante.api').build(_A)", l:source)`

Note that even after this commit is merged, still #453 error exists.


aarnphm:
#550 should fix this

Issue #445, opened 2024-09-01
Add way to apply changes from AvanteAsk without diff


luisiacc:
I think i'd be useful the ability to apply (<A>) the changes on the AvanteAsk window without having to go to the diff and select it, just apply it directly.

jikkuatwork:
Yes, I agree! We can simply undo if its wrong.

aarnphm:
There is already option for this. You can set `auto_apply_diff_after_generation = true` in config then cA to choose all theirs.

If you just want to get the change from diff then you should use suggest instead, it is basically the same.

luisiacc:
> There is already option for this. You can set `auto_apply_diff_after_generation = true` in config then cA to choose all theirs.
> 
> If you just want to get the change from diff then you should use suggest instead, it is basically the same.

It's not quite what I mean, I mean after the generation, go to the chunks of code to apply them, press something like `<g>`(instead of `<A>`) and overwrite the lines directly, without having to 'choose all theirs', choose all theirs by default after pressing the `<a>` and then `<cA>`

aarnphm:
https://github.com/yetone/avante.nvim/blob/d395aa3fea0cc24bd5c15e84aebed511ce29daa0/lua/avante/config.lua#L99

CrustyCode:
I'm not sure if this is what you mean, but: In addition to <CR> to submit your query and produce the diff, you could also have <C-CR> or something which would auto apply any changes - like an "I'm Feeling Lucky" button.

`auto_apply_diff_after_generation = false`
This setting is broken for me. Most of the time it outputs a different diff from which it says in the sidebar, or it will output the LAST thing I asked for instead of the current query.

PR #556, opened 2024-09-08
feat/fix: AvanteEdit should immediately switch into insert mode

frankroeder:
Dear Maintainers,

I suggest that `AvanteEdit` should immediately switch to insert mode, as users most often want to input a prompt.

### DIFF
--- a/lua/avante/selection.lua
+++ b/lua/avante/selection.lua
@@ -318,6 +318,7 @@ function Selection:create_editing_input()
   local winid = api.nvim_open_win(bufnr, true, win_opts)
 
   self.editing_input_winid = winid
+  vim.cmd("startinsert")
 
   api.nvim_set_option_value("wrap", false, { win = winid })
   api.nvim_set_option_value("cursorline", true, { win = winid })


Trevato:
Auto-insert mode would be frustrating. If I enter the `AvanteEdit` window by mistake, I'd have to hit `<Esc>` before I could use my exit keybind. If I want to paste I have to `<Esc>` first. The idea is you enter the window, then take action. If the action you want to take is insert, then just press your insert bind.

Trevato:
Sure bad example but do you not find yourself cycling windows? 

I find myself using the `AvanteEdit` window just like any other buffer. Sometimes I want to edit my prompt. Are you also going to decide where to put my cursor when I enter the buffer and there is content?  

The majority of cases for me is default nvim behavior.

aarnphm:
I would argue that this breaks the cycling workflow (+1 from trev here)

frankroeder:
I'm not sure if we're talking about the same thing. I'm referring to the command triggered by `<Leader>ae` that allows you to "edit a selected block." This command opens a small input prompt next to the selected code, where you can input a command to edit it.

screenshot 2024-09-11 at 10 02 16

By default, this input window is empty, and I have not encountered any instances where text is already present. I apologize for any potential confusion with `AvanteEdit`. At least the code of the PR is still doing the right thing.

aarnphm:
This is AvanteEdit

frankroeder:
Just to add: It would follow the behavior of `vim.ui.input` being an input prompt.

Trevato:
Yeah I'm aware of both input methods. My opinion remains.

georgeguimaraes:
I had asked this over on Discussions as well:
https://github.com/yetone/avante.nvim/discussions/549

aarnphm:
I kinda prefer multiple steps over progressive disclosure https://github.com/yetone/avante.nvim/issues/445#issuecomment-2327311194

aarnphm (changes requested):
ehh I'm fine with this through a behaviour option (opt-in)

aarnphm:
Hmm, I will close this PR. My reasoning are stated above.

feel free to open this again if you decide to add on option for this.

Issue #559, opened 2024-09-10
bug: Sending message crashes because of calling function in lua's loop


leiserfg:
### Describe the bug

I'm using this config
```lua
return {
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false,
  opts = {
    provider = "openai",
    openai = {
      endpoint = "http://shiralad:11434/v1",
      model = "yi-code",
      temperature = 0,
      max_tokens = 4096,
      ["local"] = true,
    },
  },
  build = "make",
  dependencies = {
    "stevearc/dressing.nvim",
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
    {
      -- Make sure to set this up properly if you have lazy=true
      "MeanderingProgrammer/render-markdown.nvim",
      opts = {
        file_types = { "markdown", "Avante" },
      },
      ft = { "markdown", "Avante" },
    },
  },
}
```
then I do `<leader>aa` and call ask a question, as soon as I sent I get this error:

```
Error executing luv callback:                                                                                                                                                                                                                                                     
...g/.local/share/nvim/lazy/avante.nvim/lua/avante/path.lua:19: E5560: nvim_buf_get_name must not be called in a lua loop callback                                                                                                                                                
stack traceback:                                                                                                                                                                                                                                                                  
        [C]: in function 'nvim_buf_get_name'                                                                                                                                                                                                                                      
        ...g/.local/share/nvim/lazy/avante.nvim/lua/avante/path.lua:19: in function 'filename'                                                                                                                                                                                    
        ...g/.local/share/nvim/lazy/avante.nvim/lua/avante/path.lua:36: in function 'get'                                                                                                                                                                                         
        ...g/.local/share/nvim/lazy/avante.nvim/lua/avante/path.lua:53: in function 'save'                                                                                                                                                                                        
        ...local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:1270: in function 'on_complete'                                                                                                                                                                               
        ...fg/.local/share/nvim/lazy/avante.nvim/lua/avante/llm.lua:150: in function '_user_on_exit'                                                                                                                                                                              
        .../.local/share/nvim/lazy/plenary.nvim/lua/plenary/job.lua:241: in function '_shutdown'                                                                                                                                                                                  
[...]

PR #560, opened 2024-09-10
fix(save): schedule save within a loop

leiserfg:


leiserfg:
Fixes #559

lua/avante/path.lua
```diff
@@ -49,10 +49,10 @@ end
 -- Saves the chat history for the given buffer.
 ---@param bufnr integer
 ---@param history table
-M.save = function(bufnr, history)
```
aarnphm, line 52: I don't think this would fix the bug in #559, rather it might introduce more bugs.

`save` usually should be called synchronously.

aarnphm (changes requested):
I think let's find the actual bug instead of this hack.

leiserfg:
The error comes from the fact that save calls history.get, which then calls vim.nvim_buf_get_name which needs to be scheduled.

leiserfg:
I'm not sure why delaying safe a few milliseconds will be a problem tho.

leiserfg:
Hi @aarnphm can you please help me with this?

Lap1n:
Looks like this proposed fix is the way it was suggested in this other closed issue on [plenary repo](https://github.com/nvim-lua/plenary.nvim/issues/189#issuecomment-882424314).

yetone:
This fix looks fine.

ibanks42:
Can this get merged then? It's currently blocking me from using a local LLM.

aarnphm (approved):

Issue #582, opened 2024-09-14
feature: minimal diff


will-lynas:
### Feature request

The LLMs tend to suggest replacement of large blocks of code, even when most of the block is unchanged. This makes it harder to review the changes it wants to make. Would it be possible to generate a minimal diff and apply that inside the merge conflict markers instead?

Example: This is generated

```sql
<<<<<<< HEAD
CREATE TABLE serbian_scores (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    user_id INTEGER,
    test_id INTEGER,
    FOREIGN KEY (user_id) REFERENCES users(id)
    ON DELETE CASCADE
);
=======
CREATE TABLE serbian_scores (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    user_id INTEGER,
    test_id INTEGER,
    FOREIGN KEY (user_id) REFERENCES users(id)
        ON DELETE CASCADE,
    FOREIGN KEY (test_id) REFERENCES serbian_tests(id)
        ON DELETE CASCADE
);
>>>>>>> Snippet
```

But it could just be

```sql
<<<<<<< HEAD
=======
    FOREIGN KEY (test_id) REFERENCES serbian_tests(id)
        ON DELETE CASCADE
>>>>>>> Snippet
```

aarnphm:
This is syntatic diff you are suggesting. I think that in the context of code it is better to dump the whole thing.

What we can do is to have a semantic diff support, but seems like too much work atm :))

will-lynas:
Can I make a PR to add it as an option? Just find lines that are the same in both sections and delete them?

will-lynas:
Maybe just stripping duplicates off the ends of the block to avoid turning one block into multiple

will-lynas:
I've had a go at making it. Seems to work from my testing

```sql
CREATE TABLE serbian_scores (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    user_id INTEGER,
    test_id INTEGER,
    FOREIGN KEY (user_id) REFERENCES users(id)
<<<<<<< HEAD
    ON DELETE CASCADE
=======
            ON DELETE CASCADE,
        FOREIGN KEY (test_id) REFERENCES serbian_tests(id)
            ON DELETE CASCADE
>>>>>>> Snippet
);
```

aarnphm:
The indentation doesn't seem correct?

will-lynas:
[...]

PR #583, opened 2024-09-14
feat: minimal diff

will-lynas:
See #582

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
     auto_set_keymaps = true,
     auto_apply_diff_after_generation = false,
     support_paste_from_clipboard = false,
-    minimize_diff = false,
+    minimize_diff = true, -- Whether to remove unchanged lines when applying a code block
   },
   mappings = {
     --- @class AvanteConflictMappings
--- a/README.md
+++ b/README.md
@@ -231,6 +231,7 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
     auto_set_keymaps = true,
     auto_apply_diff_after_generation = false,
     support_paste_from_clipboard = false,
+    minimize_diff = false,
   },
   mappings = {
     --- @class AvanteConflictMappings
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -120,14 +120,14 @@ M.defaults = {
   ---                                     Note that avante will safely set these keymap. See https://github.com/yetone/avante.nvim/wiki#keymaps-and-api-i-guess for more details.
   ---3. auto_set_highlight_group        : Whether to automatically set the highlight group for the current line. Default to true.
   ---4. support_paste_from_clipboard    : Whether to support pasting image from clipboard. This will be determined automatically based whether img-clip is available or not.
-  ---5. minimize_diff    : Whether to remove unchanged lines when applying a code block
+  ---5. minimize_diff                   : Whether to remove unchanged lines when applying a code block
   behaviour = {
     auto_suggestions = false, -- Experimental stage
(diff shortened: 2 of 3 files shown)


lua/avante/config.lua
```diff
@@ -92,12 +92,14 @@ You are an excellent programming expert.
...
   ---3. auto_set_highlight_group        : Whether to automatically set the highlight group for the current line. Default to true.
   ---4. support_paste_from_clipboard    : Whether to support pasting image from clipboard. This will be determined automatically based whether img-clip is available or not.
+  ---5. minimize_diff    : Whether to remove unchanged lines when applying a code block
   behaviour = {
     auto_suggestions = false, -- Experimental stage
     auto_set_highlight_group = true,
     auto_set_keymaps = true,
     auto_apply_diff_after_generation = false,
     support_paste_from_clipboard = false,
+    minimize_diff = false,
```
yetone, line 102: It would be better to sync this option to the README, as I am concerned that some users might not check the source code.

yetone:
I encountered an error during my local testing:

image

will-lynas:
I've added a check for that. Im interested, what do you do to get it to do that though?

yetone:
I just pulled the latest version and tried it, but I can no longer trigger any diff effects.

minimize diff:


https://github.com/user-attachments/assets/72821e86-47c2-4f34-9556-f392d9dfae93

original diff:


https://github.com/user-attachments/assets/613015e8-d465-47cd-a72a-e1d88a8a6127

README.md
```diff
@@ -205,6 +205,7 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
     auto_set_keymaps = true,
     auto_apply_diff_after_generation = false,
     support_paste_from_clipboard = false,
+    minimize_diff = false,
```
aarnphm, line 208: no need to expose this on README imo

aarnphm (commented):
wondering if we can just generate the minimal diff instead of parsing the snippets

README.md
will-lynas: @yetone said to put it in

will-lynas:
> I just pulled the latest version and tried it, but I can no longer trigger any diff effects.
> 
> minimize diff:
> minimal-diff.mp4
> 
> original diff:
> normal-diff.mp4

Thats weird. Not sure why that's happening. The code that should be applied is definetly different from whats there already, right?

will-lynas:
> wondering if we can just generate the minimal diff instead of parsing the snippets

I've moved it to where the diff is created

README.md
yetone: I think it needs to be exposed in the README.

yetone:
> wondering if we can just generate the minimal diff instead of parsing the snippets

LLMs are still unable to achieve this.

will-lynas:
I think I'm actually running into a bug which is on `main`.

As far as I understand `snippet.range` is supposed to be start inclusive and end exclusive.

But there's this loop in `insert_conflict_contents` which treats the end as inclusive:
```lua
    for i = start_line, end_line do
      table.insert(result, lines[i])
    end
```

Heres a video of the ranges getting messed up:

https://github.com/user-attachments/assets/04d764fc-600f-42c0-8ad3-9d46b10d07dd

But I also have one where it works somehow?

https://github.com/user-attachments/assets/9dfca468-ec06-468c-9545-164bf36e9f83

aarnphm:
ok seems like a bug then :))

Chen-Yulin:
> I just pulled the latest version and tried it, but I can no longer trigger any diff effects.
> 
> minimize diff:
> minimal-diff.mp4
> 
> original diff:
> normal-diff.mp4

WOW, I'd like to ask what plugin do you use to highlight the conflicts

PR #585, opened 2024-09-14
Add user option for sidebar window focus mapping

valenber:
### Motivation
The default settings for switching window focus in sidebar interfered with my copilot config, making it impossible to use it for writing prompts.

### What has been changed
I kept the default bindings `<Tab>/<S-Tab>`, but added possibility for user to override it via `opts.mappings`.

Hope it's useful! :slightly_smiling_face:

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,12 +154,10 @@ require('avante').setup ({
 
 </details>
 
-
 > [!IMPORTANT]
 >
 > For `avante.tokenizers` and templates to work, make sure to call `require('avante_lib').load()` somewhere when entering the editor.
 > We will leave the users to decide where it fits to do this, as this varies among configurations. (But we do recommend running this after where you set your colorscheme)
-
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -231,6 +229,10 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
       normal = "<CR>",
       insert = "<C-s>",
     },
+    sidebar = {
+      switch_windows = "<Tab>",
+      reverse_switch_windows = "<S-Tab>",
```
yetone, line 236: These two options might be more appropriate under `windows`, or we could make a breaking change: rename `windows` to `sidebar` and rename `windows.sidebar_header` to `sidebar.header`.

yetone: Ignore me, I was mistaken. It turns out this is under `mappings`, which makes a lot of sense.

yetone (approved):

PR #593, opened 2024-09-15
Update the readme for vim-plug installment

were:


aarnphm:
Hi there, I updated the instruction in #595.

PR #622, opened 2024-09-23
move curl body to file

zeratax:
On windows I noticed I kept getting `ENAMETOOLONG` since the curl argument was so long, so I moved the body into a file.
This is a draft since I would def need to do this for the other providers too and the temp path is currently only valid for windows, but I am unsure if this is even the correct direction to go in.

Also I had to use `pwsh` instead of `powershell` otherwise the build failed with: 
```
New-TemporaryFile : The term 'New-TemporaryFile' is not recognized as the name of a cmdlet, function, script file, or  operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

### DIFF
--- a/lua/avante/llm.lua
+++ b/lua/avante/llm.lua
@@ -122,7 +122,7 @@ M.stream = function(opts)
     headers = spec.headers,
     proxy = spec.proxy,
     insecure = spec.insecure,
-    body = vim.json.encode(spec.body),
+    body = spec.body_file,
     stream = function(err, data, _)
       if err then
         completed = true
--- a/lua/avante/providers/claude.lua
+++ b/lua/avante/providers/claude.lua
@@ -81,24 +81,33 @@ M.parse_curl_args = function(provider, prompt_opts)
   if not P.env.is_local("claude") then headers["x-api-key"] = provider.parse_api_key() end
 
   local messages = M.parse_message(prompt_opts)
+  local body_content = vim.tbl_deep_extend("force", {
+    model = base.model,
+    system = {
+      {
+        type = "text",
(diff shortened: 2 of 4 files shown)


lua/avante/llm.lua
```diff
@@ -122,7 +122,7 @@ M.stream = function(opts)
     headers = spec.headers,
     proxy = spec.proxy,
     insecure = spec.insecure,
-    body = vim.json.encode(spec.body),
```
aarnphm, line 125: this is considered breaking, if the spec doesn't contains a binary body (iiuc that is what body_file is) then this will break for other provider.

lua/avante/providers/claude.lua
```diff
@@ -81,24 +81,33 @@ M.parse_curl_args = function(provider, prompt_opts)
...
+    messages = messages,
+    stream = true,
+  }, body_opts)
+
+  local id = string.gsub("xxxx4xxx", "[xy]", function(l)
+    local v = (l == "x") and math.random(0, 0xf) or math.random(0, 0xb)
+    return string.format("%x", v)
+  end)
+
+  local body_file_path = string.format("%s\\AppData\\Local\\Temp\\avante_body_%s.json", os.getenv("USERPROFILE"), id)
```
aarnphm, line 102: let's not do this, this won't work on linux.

aarnphm (changes requested):
I think let's try to investigate what leads to this error. 

This is very much a hack.

Can you provide a small reproducer?

aarnphm:
closed due to inactive. Feel free to open once you have time to tackle this again. Thanks.

PR #638, opened 2024-09-26
fix: windows build script

yetone:
‌‌‌Since I don't have Windows, I can't verify if it's correct. Could you please help me check it, @yuchanns?


yuchanns:
Confirmed.

PR #673, opened 2024-10-01
Write curl body to a temporary JSON file

### LINKED ISSUE #672: bug: plenary curl fails to spawn process on Windows due to maximum command line length
### Describe the bug

This plugin works great, but I noticed it's failing on certain files for me. After testing a little, it seems to be failing on longer files in particular.

My curl works fine, and I edited the plugin locally to dump to the body to a temporary JSON file; then using curl with that file I'm getting output as expected.

EDIT: I see this PR now https://github.com/yetone/avante.nvim/pull/622
Seems like the same issue.

For more reading: https://stackoverflow.com/questions/3205027/maximum-length-of-command-line-string
(I wasn't aware of this limitation prior today...)

### To reproduce

Standard config, on Windows

1. Open a longish file (>1000 lines)
2. Visual mode select all and ask "please review this"
3. Prints error

### Expected behavior

Longer content/files should work on Windows as well.

Since calling curl from my command line worked, I figured I'd edit avante locally to use the JSON as well such that the curl request uses ``--data "@request_body.json"`` instead of `--data-raw`.

Replacing body accordingly in `llm.lua` seems to do the trick (currently hardcoded path):

```lua
  vim.fn.writefile(vim.split(vim.json.encode(spec.body), "\n"), "c:/tmp/request_body.json")

  active_job = curl.post(spec.url, {
    headers = spec.headers,
    proxy = spec.proxy,
    insecure = spec.insecure,
    -- body = vim.json.encode(spec.body),
    body = "c:/tmp/request_body.json",
    ...
```

I'll submit a PR for your convenience.

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  opts = {
    -- add any opts here
  },
  build = "make",
  dependencies = {
    "nvim-treesitter/nvim-treesitter",
    "stevearc/dressing.nvim",
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}
```

### Environment

NVIM v0.10.1
Build type: Release
LuaJIT 2.1.1713484068
Platform: Windows

### Repro

_No response_

Huite:
Fixes #672

Should have the same effect as #622, but cross-platform since this uses a tempname() provided by neovim instead.
(Somewhat surprising to me: the plenary curl wrapper checks whether the content of `body` is a file, and if so, it prepends `@` telling curl to read the file contents. See: https://github.com/nvim-lua/plenary.nvim/blob/2d9b06177a975543726ce5c73fca176cedbffe9d/lua/plenary/curl.lua#L205)

I don't think you'd want to put the logic at the provider, since you'll have to repeat it for every provider. This will work for all providers since it's just calling curl in a slightly different way.

### DIFF
--- a/lua/avante/llm.lua
+++ b/lua/avante/llm.lua
@@ -122,11 +122,15 @@ M.stream = function(opts)
 
   local active_job
 
+  local temp_file = vim.fn.tempname() .. ".json"
+  local json_content = vim.json.encode(spec.body)
+  vim.fn.writefile(vim.split(json_content, "\n"), temp_file)
+
   active_job = curl.post(spec.url, {
     headers = spec.headers,
     proxy = spec.proxy,
     insecure = spec.insecure,
-    body = vim.json.encode(spec.body),
+    body = temp_file,
     stream = function(err, data, _)
       if err then
         completed = true
--- a/lua/avante/llm.lua
+++ b/lua/avante/llm.lua
@@ -159,10 +159,12 @@ M.stream = function(opts)
     on_error = function()
       active_job = nil
       completed = true
+      pcall(os.remove, temp_file)
       opts.on_complete(nil)
     end,
     callback = function(result)
       active_job = nil
+      pcall(os.remove, temp_file)
       if result.status >= 400 then
         if Provider.on_error then
           Provider.on_error(result)
--- a/lua/avante/llm.lua
+++ b/lua/avante/llm.lua
@@ -1,4 +1,5 @@
 local api = vim.api
+local fn = vim.fn
 
 local curl = require("plenary.curl")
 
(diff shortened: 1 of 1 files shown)


trevorprater:
Also useful for debugging.

yetone:
LGTM, but there needs to be a cleanup logic for the temp file.

Huite:
I've added `pcall(os.remove, temp_file)` to delete the temporary file immediately after use.

(`vim.fn.delete` seems to run into "Vimscript function must not be called in a lua loop callback" issues)

For debugging, it might be worthwhile to store the temp_file path and not remove it immediatelly, but that would be a separate issue IMO.

yetone:
> I've added `pcall(os.remove, temp_file)` to delete the temporary file immediately after use.
> 
> (`vim.fn.delete` seems to run into "Vimscript function must not be called in a lua loop callback" issues)
> 
> For debugging, it might be worthwhile to store the temp_file path and not remove it immediately, but that would be a separate issue IMO.

You can call `vim.fn.delete` within a `vim.schedule` to avoid this issue.

PR #683, opened 2024-10-05
Handle backticks before decoding JSON

ryul99:
Since LLM models often do not remove backticks in their suggesting responses, particularly with smaller models, I have implemented backtick handling in the code. Please let me know your thoughts.
backticks_example

### DIFF
--- a/lua/avante/suggestion.lua
+++ b/lua/avante/suggestion.lua
@@ -85,6 +85,7 @@ function Suggestion:suggest()
       vim.schedule(function()
         local cursor_row, cursor_col = Utils.get_cursor_pos()
         if cursor_row ~= doc.position.row or cursor_col ~= doc.position.col then return end
+        full_response = full_response:gsub("^`*%w*\n(.-)\n`*$", "%1")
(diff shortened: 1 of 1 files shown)


yetone:
‌Very helpful! Thank you!

lua/avante/suggestion.lua
```diff
@@ -85,6 +85,7 @@ function Suggestion:suggest()
       vim.schedule(function()
         local cursor_row, cursor_col = Utils.get_cursor_pos()
         if cursor_row ~= doc.position.row or cursor_col ~= doc.position.col then return end
+        full_response = full_response:gsub("^`*%w*\n(.-)\n`*$", "%1")
```
yetone, line 88:
```suggestion
        full_response = full_response:gsub("^```%w*\n(.-)\n```$", "%1")
```

Would such an accurate match be better? Because I would worry about the following normal content, such as multi-line strings in certain languages:

```
`firstline
secondline`
```

ryul99: I think the accurate match version is better! I'll fix it.

PR #685, opened 2024-10-06
Add other open buffer files to context. 

hesiod-au:


JunYang-tes:
It's great, I think it would be even better if, like codecompanion.nvim, it allowed users to select which buffers to insert.

dimroc:
I would love this. Although a separate keymap would be nice, say <space>+ab

yetone:
‌‌‌‌‌‌‌Can it be turned into a configurable item? And how about as a cmp mentions? For example: `@buffers`

dimroc:
Btw, I assumed this \<space>+ab would add all visible buffers in the current tab to the context. Adding ALL buffers could get sloppy as your open buffers list grows quite large over a long vim session.

AndrewGerstenslager:
I pulled this branch into my config and keep getting the same error no matter what I ask:

```
- Date
time: 2024-11-14 11:07:42

- Model: claude/claude-3-5-sonnet-20240620

> can you provide the javascript so I can move it to a different file?

I apologize, but it seems there wasn't a specific request or question in your last message. Could you please provide a clear instruction or ask a specific question about what changes you'd like to make to the code? I'm here to help with any modifications or improvements you need for the HTML file or any other part of the project.

**Generation complete!** Please review the code suggestions above.
```

If theres something I did wrong let me know. All I did was change my Lazy source to "hesiod-au/avante.nvim" and reloaded and updated my Lazy config.

I agree with other suggestions that there should be a keybind or inline command to add buffers to context. Copilot chat in VsCode has the "#" symbol which you can then explicitly add a file to the context. Something as a command in the chat window or a keybind to list and choose buffers would be a nice feature rather than all buffers.

dimroc:
FWIW the two features have eliminated my need for this:
1. adding an individual file with leader+a+c
2. The upcoming ability to add files in the quickfix with a @mention. (I haven't gotten this to work yet)  https://github.com/yetone/avante.nvim/commit/be92be6124f743c30ba4b52ded908d02cfb80a1b


Do others feel the same?

yetone:
‌‌‌‌‌‌I'll close this very old PR for now, and I'll implement it in another way.

PR #686, opened 2024-10-07
O1 parsing code replace fix

LessComplexity:
O1 models sometimes return the following format when marking lines to replaces:
```md
**Replace Lines**: 3-23
**Replace Lines: 2-23**
**Replace Lines**: `3-23`
```

This format doesn't allow completion to work correctly, so I've added a fix that clears the current line to parse the "Replace Lines" correctly on O1 models too.
Since the new code only cleans the ```**``` and ``` ` ``` characters -> it still works as expected with Claude and all other models.


yetone:
‌‌‌Hello, avante.nvim has refactored the prompt and is now using aider's prompt. It will no longer rely on line numbers returned by LLM, so it seems this issue won't occur anymore. You can try updating and testing it.

https://github.com/yetone/avante.nvim/pull/671

LessComplexity:
> ‌‌‌Hello, avante.nvim has refactored the prompt and is now using aider's prompt. It will no longer rely on line numbers returned by LLM, so it seems this issue won't occur anymore. You can try updating and testing it.
> 
> #671

Awesome I saw the new change and it's great architecture wise :)
Problem is, that it didn't take into account a situation when the messages are not streaming correctly, this works on every other model than O1.
Basically we need to add something a bit different for non-streaming messages from the models since from my tests it doesn't work at all without streaming (AKA O1 models).

To fix this, I think the best option is either to change the "on_chunk" function to handle non-streaming requests (the function fails in the openai non-stream request processor), or to separate the searching/replacing from the "on_chunk" itself, thus not interfering with model answers (searching adds latency from chunk to chunk) and allowing it to work both during streaming and non streaming - in the same place in code per se.

I will try to work on one of the options now to let the O1 models I use work with it fluently too, will update on progress :)

LessComplexity:
@yetone 
I've made it work!
My comment above is irrelevant, the code actually does suppose to work with the O1 model, the thing that made it not work is the use of this part in the code:
```lua
      if is_first_chunk then
        is_first_chunk = false
        self:update_content(content_prefix .. chunk, { stream = false, scroll = true })
        return
```

Since O1 models give a result it one line, this code was just returning the content without replacing the relevant lines for parsing. I've basically removed this as this doesn't affect the Claude and other streaming models and provides the ability to work with O1 models as they are meant too.

lua/avante/sidebar.lua
```diff
@@ -1214,11 +1218,7 @@ function Sidebar:create_input(opts)
       transformed_response = transformed.content
       prev_is_searching = transformed.is_searching
       local cur_displayed_response = generate_display_content(transformed)
-      if is_first_chunk then
-        is_first_chunk = false
-        self:update_content(content_prefix .. chunk, { stream = false, scroll = true })
```
yetone, line 1219: If this logic is removed, `**Generating response ...**\n` will always exist in the response buffer.

LessComplexity: From all my tests it did remove the `**Generating response ...**\n` line
Did I miss something?

yetone: ‌‌‌‌‌‌‌This is a bug, thank you for helping to find it. I will fix it next.

LessComplexity: The problem is that if this code is kept one way or another, then the "transformed_response" is not printed out which doesn't parse the result of O1 or non-stream models. Now that I've removed this line it does work, so there is no bug isn't there? 😂

yetone: ‌There is currently no issue, mainly because I did not update the response buffer using the stream method. This was my oversight. I will use the stream method to update the response buffer next, so the judgment of is_first_chunk should be retained.

lua/avante/sidebar.lua
```diff
@@ -282,18 +282,22 @@ local function extract_code_snippets(response_content)
   local explanation = ""
 
   for idx, line in ipairs(vim.split(response_content, "\n")) do
+    -- Remove bold formatting markers for consistent matching
+    local clean_line = line:gsub("%*%*", "")
```
yetone, line 286: The current `Replace Lines: 2-23` is generated by code, not returned by LLM, so no special handling is needed because it will definitely be `Replace Lines: 2-23`, and there won't be any other situation.

lua/avante/sidebar.lua
yetone: ‌‌‌‌‌I have finished making changes on my end, you can sync the latest code now.

LessComplexity:
I've fixed the new version of the code to work with O1 models.
The fix is pretty similar to before - use transformed content instead of chunk on first_chunk, cause no-stream responses (like those from O1) work in a single chunk either way.

The fix doesn't interfere with other functionality of different models/providers from my tests :)

lua/avante/sidebar.lua
```diff
@@ -1353,7 +1353,7 @@ function Sidebar:create_input(opts)
       local cur_displayed_response = generate_display_content(transformed)
       if is_first_chunk then
         is_first_chunk = false
-        self:update_content(content_prefix .. chunk, { scroll = true })
+        self:update_content(content_prefix .. cur_displayed_response, { stream = false, scroll = true })
```
yetone, line 1356: ‌‌‌‌‌‌‌There is no longer a "stream" parameter now. Another question is, won't this change affect other providers and models?

LessComplexity: Actually it works as it is now with O1 models the previous changes that were already merged, so I've removed the latest changes and this issue can be closed :)

Issue #717, opened 2024-10-13
feature: Auto insert mode in `Ask` buffer


gytis-ivaskevicius:
### Feature request

It should be easy to do with autocmd after playing around, but I feel like default behavior should be auto insert mode

gam-phon:
@gytis-ivaskevicius could you suggest the autocmd please and the way. It is annoying that it is not in insert mode by default.

PR #721, opened 2024-10-14
[Feature] Floating input

b0o:
Open a floating input similar to the "edit" input for the "ask" input. Enabled in config via `Config.windows.ask.floating` or by passing `{ floating = true }` to the `api.ask` method.

Includes logic to ensure the sidebar uses the correct buffer and selection if an existing sidebar is open for another code buffer.

Also refactored the `selection` module to extract the floating input logic into a new `PromptInput` class.

Not sure about the config options, happy to rename/move them.

Partially addresses #717 by defaulting to start in insert mode for floating input.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,7 +271,7 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
     },
     ask = {
       floating = false, -- Open the 'AvanteAsk' prompt in a floating window
-      start_insert = true, -- Start insert mode when opening the ask window
+      start_insert = true, -- Start insert mode when opening the ask window, only effective if floating = true.
       border = "rounded",
     },
   },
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -262,6 +262,18 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
...
+    input = {
+      prefix = "> ",
+    },
+    edit = {
+      border = "rounded",
+      start_insert = true, -- Start insert mode when opening the edit window
+    },
+    ask = {
+      floating = false, -- Open the 'AvanteAsk' prompt in a floating window
+      start_insert = true, -- Start insert mode when opening the ask window
```
yetone, line 274: ‌‌‌‌When it is not a floating input, does the `start_insert` configuration not take effect?

b0o: Correct, but it might make sense to make it work in both cases.

PR #734, opened 2024-10-18
feat(repo-map): C++ improvements

b0o:
- fixed C++ queries
- improved C++ definition extraction
- switched to BTreeMaps to ensure output order is stable
- added tests
- re-enabled tests for rust

Caveats: Private members are not omitted, and a lot of C++ syntax still isn't handled properly.

As a follow-up, it would be nice to support namespaces, but it seems difficult with the current structure of the code. I think it might be a good idea to refactor `extract_definitions` to make it more modular.


aarnphm (approved):
LGTM. cc yetone once he has time

yetone:
Yes, we need to refactor `extract_definitions`.

Issue #604, opened 2024-09-18
feature: Additional context with Open Buffers in Active Tab, Active Split Buffers, All Open Buffers


varunshoor:
### Feature request

What a commendable work you have done, I am a fan.

Would it be possible to provide additional contexts (with options for existing commands perhaps):
* Open Buffers in Active Tab: I routinely group my tasks in tabs, so this will allow me to provide the context for entire task
* Active Split Buffers: I split my code and test case, currently the context is missed when using this to generate tests.
* All Open Buffers: This is one level below Project context and useful when you want to prompt with context for all open tasks

I don’t know if you accept sponsorships but I would rather sponsor this plugin than pay for Cursor’s subscription if this can be implemented.

Thank you for everything

hesiod-au:
https://github.com/yetone/avante.nvim/pull/685

PR #738, opened 2024-10-19
option for additional nvim cmp sources

napisani:
Hi there, 

First of all, id like to thank you for creating this plugin. I use it all the time and its wonderful, please keep up the good work!

This PR adds an extra configuration parameter so that the nvim user can define extra cmp sources to be used in the sidebar / selection buffers. 

Selfishly this is helpful for another plugin that im working on but i think other users will find utility in the ability to tailor exactly which cmp sources contribute suggestions via cmp type-ahead. IE it may be helpful to have path suggestions to auto complete file names etc.

Let me know what you think,
~Nick

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,6 +270,12 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
       incoming = "DiffAdd",
     },
   },
+
+  additional_cmp_sources = {
+    { name = "path" },
+    { name = "nvim_lsp" },
+    { name = "buffer" },
+  }, -- Additional cmp sources to be used in the sidebar and selection buffers
   --- @class AvanteConflictUserConfig
   diff = {
     autojump = true,
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -182,6 +182,7 @@ Respect and use existing conventions, libraries, etc that are already present in
   hints = {
     enabled = true,
   },
+  additional_cmp_sources = {}, -- Additional cmp sources to be used in the sidebar and selection buffers
 }
 
 ---@type avante.Config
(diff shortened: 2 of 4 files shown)


aarnphm (changes requested):
I felt like this is not needed and would make the code too complex.

napisani:
@aarnphm  would you be willing to allow me to add autocmds that get fired after InsertEnter, so that the user can make their own adjustments to the buffer without the sources being controlled by avante.nvim?

aarnphm:
Can you give an example where additional_cmp_source would be helpful here? `buffer` only works within input buffer,meaning it don't have context from other buffer source (correct me if I'm wrong)

`lsp` also doesn't work afaict, given that lsp handler have to attach to the floating buffer (which doesn't contain any context for said code)

napisani:
Absolutely!
the main reason i suggested this change is because i have been working on authoring a plugin: https://github.com/napisani/context-nvim

this plugin allows neovim users to manage contextual code and it provides a nvim-cmp source to retrieving the formatted contexts in any buffer.

for example you can add 3 files to your manual context, then you can start typing `@manual_context` in any buffer and it will allow you to reference the context.

i would love to be able to use this plugin with Avante to give users the ability to tailor their context selections to the areas of the code that they know are relevant to their prompt, similar to the cursor experience. 

wdyt?

aarnphm:
cool concept, but I felt like this use case should be covered with `@codebase`?

seems like similar to manual adding context #604 

also that plugin are unrelated to the cmp change. `nvim-cmp` only applies in input buffer, which I don't really see a point of dumping the whole context here

probably should do it through `repomap` or proper workspace context. :)

PR #739, opened 2024-10-21
feat: clear key mapping; default to <C-l>

teocns:
Uses `Config.windows.ask.start_insert` to determine whether to go insert mode after clearing. Is this something we want, or should we create a separate config?

Clear mapping is set on `mappings.sidebar.clear`, but feel free to suggest a better one if this doesn't fit the sidebar context


teocns:
@yetone have my back with the linter please

yetone:
‌‌‌‌Should we confirm before clearing? Because in my opinion, deletion is always the most dangerous operation.

teocns:
`<C-l>` is a pretty standard clear command. Perhaps one doesn't just randomly press that. Usually no confirmation is prompted (hence the "shortcut" name)

aarnphm (changes requested):
I actually don't agree with this

teocns:
How about double `<C-l> <C-l>` + warning at first key stroke? 
I understand deleting is dangerous and might require some sort of confirmation, though I wouldn't see the point in having an extra confirmation step such as  pressing `Y`: it'd consume as much time as typing `/clear`, and Y is also a painful key to reach

aarnphm:
L1245

teocns:
Right, I just replicated the same behavior of `/clear` buying a safe ticket to consistency in behaviors:

https://github.com/yetone/avante.nvim/blob/a8e2b9a00c73b11d28857f0f5de79a9022281182/lua/avante/sidebar.lua#L1199-L1208

You might wanna look at the [blame](https://github.com/yetone/avante.nvim/blame/main/lua/avante/sidebar.lua#L1199-L1208) history.

teocns:
Added double keystroke for `<C-l>` + warning showing to press again before deleting

https://github.com/yetone/avante.nvim/pull/739/commits/4778e1e2da90af8ad3c9976140f36f209ca97328

This behavior lines up with my expectation, but if you don't vision it merged feel free to leave it aside or modify it at will

yetone:
‌Thank you for your contribution. Since this PR hasn't been updated for a long time, it has been closed. Sorry about that.

Issue #499, opened 2024-09-04
HTTP 401 error with gemini provider's tokenizer (gemma-2b)


meinside:
This error occurs with gemini:

```
Error executing vim.schedule lua callback:
called `Result::unwrap()` on an `Err` value: 
RequestError(Status(401, Response[status: 401, status_text: Unauthorized, url: https://huggingface.co/google/gemma-2b/resolve/main/tokenizer.json]))
```

I see codes which check for an environment variable named `HF_TOKEN`, but it is not used elsewhere.

aarnphm:
Hi there, gemma/2b is a gated models, so you must have access to the repo on huggingface in order to use the tokenizer.

meinside:
I exported my huggingface's token (with the required access to gemma models), but it seems not picked up by the tokenizer.

meinside:
I tested with my huggingface's token:

```bash
$ curl "https://huggingface.co/google/gemma-2b/resolve/main/tokenizer.json" -H "Authorization: Bearer $HF_TOKEN"
```

then I got a url which redirects to the tokenizer json file:

```
Found. Redirecting to https://some-url-to-tokenizer-json-with-many-many-query-params
```

so I think the token is valid. Is there anything else I can check?

aarnphm:
I will just gpt-4o for now, until I implement token counting will revisit this

meinside:
Ok, thanks!

PR #760, opened 2024-10-25
revert(cohere): revert to gpt-4o as tokenizer

maxbrunet:
Hello there! 👋 

Same as #506

FYI the tokenizer can be download from Cohere directly, it would avoid the user from configuring a HuggingFace account in addition to the Cohere API token: https://docs.cohere.com/v2/docs/tokens-and-tokenizers#downloading-a-tokenizer

Thank you for this wonderful project! ❤️

### DIFF
--- a/lua/avante/providers/cohere.lua
+++ b/lua/avante/providers/cohere.lua
@@ -29,7 +29,9 @@ local P = require("avante.providers")
 local M = {}
 
 M.api_key_name = "CO_API_KEY"
-M.tokenizer_id = "CohereForAI/c4ai-command-r-plus-08-2024"
+-- FIXME: The tokenizer should also be fetched from Cohere to avoid a dependency on HuggingFace
+-- https://docs.cohere.com/v2/docs/tokens-and-tokenizers#downloading-a-tokenizer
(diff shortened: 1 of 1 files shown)


aarnphm:
thanks for the PR. I will kindly close this afaik the tokenizers library in Rust from HF doesn't support this

I will think about this once I update with tool use, given that it will be closely related to tokenizers

maxbrunet:
Hey @aarnphm, thank you for looking into this, I can remove the comment if you prefer, the source of the tokenizer is not the problem, it is simply broken at the moment like Gemini even with the `HF_TOKEN` environment variable (same issue as the Gemini PR closed: #499). So it'd be nice if that could be merged, so the default config for the `cohere` provider can work out of the box.

The source of the tokenizer is merely something to consider when fixing the root of the problem

aarnphm:
I'm fixing this atm, just that for cohere gpt-4o tokenizer wasn't as good, hence the token_id from hg

aarnphm:
hi there, #765 should address this.

though for most stuff going forward, if the model is gated it is probably best to also set HF_TOKEN, given that we are depending on hf/tokenizers

PR #770, opened 2024-10-29
docs: replace wiki url

codycjy:
Original url return not found error, replace it with wiki url


yetone:
sorry, I have merged this: https://github.com/yetone/avante.nvim/pull/772

PR #773, opened 2024-10-30
feat: add configurable input window height for vertical layout

amojury:
- Updated README.md and config.lua to include a new height option for the input window in vertical layout.
- Modified sidebar.lua to use the new configurable height instead of a hard-coded value of 8.

This change allows users to customize the height of the input window when using the vertical layout, providing more flexibility in the UI.


b0o:
LGTM

PR #777, opened 2024-10-31
fix(repo-map) go vendor included in repo map causing token limit error

brewinski:
When using the `@codebase` flag, I encountered token exceeded errors. Upon investigation, I discovered that this issue arises because the Go project I'm working on uses a `vendor` folder for its dependencies. This folder was being included as part of the repo map, resulting in larger project context.

I am open to feedback and willing to alter the solution if this approach is not suitable. Please let me know if there are any suggestions or improvements that can be made.

### DIFF
--- a/lua/avante/utils/init.lua
+++ b/lua/avante/utils/init.lua
@@ -583,7 +583,7 @@ local function pattern_to_lua(pattern)
 end
 
 function M.parse_gitignore(gitignore_path)
-  local ignore_patterns = { "%.git", "%.worktree", "__pycache__", "node_modules" }
+  local ignore_patterns = { "%.git", "%.worktree", "__pycache__", "node_modules", "vendor" }
   local negate_patterns = {}
   local file = io.open(gitignore_path, "r")
   if not file then return ignore_patterns, negate_patterns end


b0o:
This seems like it would be useful to have a configuration option. We could create a new configuration sub-table for repo-map options e.g.:

```lua
{
  repo_map = {
    ignore_patterns = { "%.git", "%.worktree", "__pycache__", "node_modules" }, -- ignore files matching these patterns when building the repo map
  }
}
```

With this approach, I think it's debatable whether `vendor` should be added to the default configuration in this case. It's not as standard as `node_modules`, some users may actually want to include things inside of `vendor` directories. Users who do want to ignore it can opt-in.

If you decide to take this approach, in order to decouple `parse_gitignore` from repo-map specifics, maybe we should not set the default `ignore_patterns` directly inside of `parse_gitignore`, but extend the returned table inside of `RepoMap._build_repo_map`:

```lua
function M.parse_gitignore(gitignore_path)
  local ignore_patterns = {}
  local negate_patterns = {}
  --- ... snip ...
  return ignore_patterns, negate_patterns
end

function RepoMap._build_repo_map(project_root, file_ext)
  local output = {}
  local gitignore_path = project_root .. "/.gitignore"
  local ignore_patterns, negate_patterns = Utils.parse_gitignore(gitignore_path)
  vim.list_extend(ignore_patterns, Config.repo_map.ignore_patterns)
  -- maybe we should support configurable negate_patterns too?
  -- ... snip ...
end
```

---

Additional notes for future repo-map improvements:

- In the future, the `repo_map` configuration can be expanded with additional options, e.g.:

  ```lua
  {
    repo_map = {
      enabled = true, -- include additional context about your project with messages sent to LLMs
      ignore_patterns = { "%.git", "%.worktree", "__pycache__", "node_modules", "vendor" }, -- ignore files matching these patterns when building the repo map
      vcs_ignore = true,  -- in addition to ignore_patterns, also respect .gitignore files
    }
  }
  ```

  Also, looking at the implementation of `RepoMap._build_repo_map`, it seems that only the project's root `.gitignore` is respected. Ideally we would respect nested `.gitignore` files. We already have a dependency on Plenary, which already has support for this, so we should leverage it.
  
- Perhaps we should have a configurable size limit for the repo_map, both in terms of the size of individual files and the overall number of files to include.

brewinski:
Thanks for the feedback @b0o. Seems like a reasonable approach. I'll look into implementing this over the next couple of days.

yetone:
‌‌‌‌‌@brewinski Thank you for your PR and @b0o's suggestions! I'll merge this PR first, and then you can submit a new PR to make it configurable.

PR #788, opened 2024-11-03
feat(repo_map): add scala support

susliko:
Thanks for the amazing plugin!

This PR brings scala support to repo map


aarnphm (approved):
nice

PR #789, opened 2024-11-03
fix(sidebar): apply windows.wrap setting to the selected code and avante input

brewinski:
Small change to apply the wrap setting to the selected code and avante input windows in the sidebar. 

If the preference is to allow the user to configure each window individually, I can implement this by extending the window configuration object to include a input and selected code wrap setting.

As always, let me know if there are any suggestions or improvements that can be made.

Before:
image

After:
image

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -1368,7 +1368,9 @@ function Sidebar:create_selected_code()
         winid = self.input.winid,
       },
       buf_options = buf_options,
-      win_options = base_win_options,
+      win_options = vim.tbl_deep_extend("force", base_win_options, {
+        wrap = Config.windows.wrap,
+      }),
       position = "top",
       size = {
         height = selected_code_size + 3,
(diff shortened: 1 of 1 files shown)


aarnphm (commented):
you might also have to handle how to handle the extmark for `confirm` correctly with wrap (the reason we didn't handle it earlier was that this was a bit too hard fwiw, or idk how to do it).

brewinski:
Thanks for the feedback @aarnphm. Do you mean handling the position of `<CR>: submit` hint text? If so, I'll see what I can do about the position of this.

aarnphm:
yeah the `<CR>: submit`

brewinski:
@aarnphm I've tried a couple of different options. As you mentioned it's fairly difficult to get it to work well in all situations.

The simplest option is to move the hint into the header and have the header update based on the user's mode and input. e.t.c. The main drawback of this is that it increases the width of the header. 
image

Alternatively, using extmark / virtual text to place the hint text at the end of the line of the current cursor can work. A drawback is that the hint will overflow the window when the text is close to the end of the line until the line wraps. The benefit being that the text should never overlap the users input.

There's an option where we calculate the wrap position however I haven't tried this yet.

If you have any other ideas or have a preference for one of these options happy to hear it and please let me know.

PR #791, opened 2024-11-04
fix(copilot): refresh token before sending the request

brewinski:
I was occasionally encountering 401 error responses when using the Copilot provider. Upon investigation, I discovered that the chat/complete request was being sent before the token could be refreshed.

To address this issue, I have modified the token refresh call to be synchronous. Additionally, I have used `vim.schedule` to refresh the token on startup.


lua/avante/providers/copilot.lua
```diff
@@ -91,22 +91,24 @@ H.refresh_token = function()
...
+    local response = curl.get(H.chat_auth_url, {
       headers = {
         ["Authorization"] = "token " .. M.state.oauth_token,
         ["Accept"] = "application/json",
       },
       timeout = Config.copilot.timeout,
       proxy = Config.copilot.proxy,
       insecure = Config.copilot.allow_insecure,
-      on_error = function(err) error("Failed to get response: " .. vim.inspect(err)) end,
-      callback = function(output)
```
aarnphm, line 103: This should already start the jobs regardless

The reason why we don't want to refresh the tokens is to avoid internal GitHub rate limits, which a few users reported to

https://github.com/nvim-lua/plenary.nvim/blob/2d9b06177a975543726ce5c73fca176cedbffe9d/lua/plenary/curl.lua#L312

If we need to change this logic then we need to check the expires_at logic

lua/avante/providers/copilot.lua
```diff
@@ -161,7 +163,8 @@ M.setup = function()
       oauth_token = H.get_oauth_token(),
     }
   end
-  H.refresh_token()
+
+  vim.schedule(function() H.refresh_token() end)
```
aarnphm, line 167: hmm why do we need to schedule this?

afaik this should be called as soon as `VeryLazy` is triggered

aarnphm (commented):
few comments

lua/avante/providers/copilot.lua
brewinski: Apologies if i've misunderstood something. 

The issue I'm facing is a race condition. parse_curl_args uses an expired token due to refresh_token running an async job.

The first request after a token expires will fail. Subsequent requests succeed until the token expires again

Running refresh_token as a sync job solved the problem for me.

I didn't find any problems with the expiry logic. I don't think this will increase or change the number of requests made to the api.

If there is a better way to go about this please let me know.

lua/avante/providers/copilot.lua
brewinski: I found that making the token request synchronous increased neovim startup time.

I'm fairly new to lua/nvim lua so I understand if this isn't a great way to approach this issue.

brewinski:
> few comments

Thanks for taking the time to review @aarnphm. 

For context, I didn't understand your point regarding the token rate limits. My response might not be very helpful.

lua/avante/providers/copilot.lua
dimfeld: Is there a reason we want to refresh on editor load? IMO the proper change is:

- Remove this line completely
- Use the rest of the change in the PR above to make the refresh call finish synchronously before the 

This way we don't try to refresh the token on every editor startup, which helps with rate limits in cases where you start the editor but don't use Avante (e.g. when writing a Git commit message or editing a Markdown file), but we do refresh it  if needed before making the request.

lua/avante/providers/copilot.lua
aarnphm: What you suggest is essentially the same with the previous logic.

but if this works for you then I'm fine merging it in.

lua/avante/providers/copilot.lua
```diff
@@ -91,22 +91,24 @@ H.refresh_token = function()
...
-      on_error = function(err) error("Failed to get response: " .. vim.inspect(err)) end,
-      callback = function(output)
-        M.state.github_token = vim.json.decode(output.body)
-        local file = Path:new(copilot_path)
-        file:write(vim.json.encode(M.state.github_token), "w")
-        if not vim.g.avante_login then vim.g.avante_login = true end
-      end,
     })
+
+    if response.status >= 200 and response.status < 300 then
```
aarnphm, line 104: wait shouldn't this be 200 only?

lua/avante/providers/copilot.lua
dimfeld: I see this bug as well sometimes, and I think the timeline of the bug is something like this:

- Step away from the computer with Neovim open, and the token expires in the meantime
- Return and do something in Avante
- Because the token is expired, the refresh request starts and runs in the background. Lets say it takes 1 second to return for this example. But because it runs asynchronously, the Lua code here continues running.
- The Copilot request is created, and sent, but the new token hasn't been obtained yet, so the old expired one is used.
- The refresh request finishes and updates the state with the new token.

The issue is that `plenary.curl` runs in the background if a `callback` is passed. So with this change, `refresh_token` blocks on the request finishing, so its guaranteed that the token will be updated (assuming it succeeds) before the LLM request is submitted.

lua/avante/providers/copilot.lua
brewinski: Yes, I'll update this tonight.

lua/avante/providers/copilot.lua
brewinski: @aarnphm is this the pull request you're referring to? https://github.com/yetone/avante.nvim/pull/746/

I've been using this change for the last couple of days and haven't had any rate limit problems yet. I'll keep this in mind while I'm working so that I can address any regressions with this change.

brewinski:
@aarnphm I've updated the response status logic. Thanks again for spending your time looking into this, please let me know if you need any further changes or updates from me.

lua/avante/providers/copilot.lua
aarnphm: yes

aarnphm (approved):
lgtm thanks

PR #810, opened 2024-11-06
feat(cody) sourcegraph cody provider implementation

brewinski:
I've tested this with a free/pro and an enterprise account. It's missing some of the context and code search features but it's a good start.

If this isn't something worth adding as a supported provider, I can document this configuration in the Wiki for other users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,7 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
 
 ```lua
 {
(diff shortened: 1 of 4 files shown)


aarnphm:
I'm fine with keeping this as a external source in the wiki page. We don't have the bandwidth to maintain this atm.

mortymacs:
No plan to reconsider Cody?

brewinski:
@mortymacs if it's helpful, I've been keeping a provider plugin up to date for Sourcegraph Cody. Might be worth a shot.

https://github.com/brewinski/avante-cody.nvim

PR #813, opened 2024-11-06
fix(openai): add backward compat for `get_user_message`

BruceChen7:
opts.message is of type AvanteLLMMessage. The message.content field should be used. Otherwise, table.concat(opts.messages) will fail.

### DIFF
--- a/lua/avante/providers/openai.lua
+++ b/lua/avante/providers/openai.lua
@@ -39,7 +39,10 @@ M.role_map = {
 }
 
 ---@param opts AvantePromptOptions
-M.get_user_message = function(opts) return table.concat(opts.messages, "\n") end
+M.get_user_message = function(opts)
+  local content = vim.tbl_map(function(message) return message.content end, opts.messages)
+  return table.concat(content, "\n")
+end
 
 M.parse_messages = function(opts)
   local messages = {}
--- a/lua/avante/providers/openai.lua
+++ b/lua/avante/providers/openai.lua
@@ -44,7 +44,7 @@ M.get_user_message = function(opts)
   return table.concat(
     vim.iter(opts.messages):filter(function(_, value) return value.role == "user" end):fold({}, function(acc, value)
       acc = vim.list_extend({}, acc)
-      acc = vim.list_extend(acc, { value })
+      acc = vim.list_extend(acc, { value.content })
       return acc
     end),
     "\n"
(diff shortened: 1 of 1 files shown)


BruceChen7:
@yetone  Can you help confirm this issue? Version 0.0.8 doesn't have this problem, but version 0.0.9 introduced it.

aarnphm:
Are you using get_user_message in your custom providers? 

since I don't think we use this anywhere

lua/avante/providers/openai.lua
```diff
@@ -39,7 +39,10 @@ M.role_map = {
 }
 
 ---@param opts AvantePromptOptions
-M.get_user_message = function(opts) return table.concat(opts.messages, "\n") end
+M.get_user_message = function(opts)
+  local content = vim.tbl_map(function(message) return message.content end, opts.messages)
+  return table.concat(content, "\n")
+end
```
aarnphm, line 45: Can you test with the following?

```suggestion
---@param opts AvantePromptOptions
M.get_user_message = function(opts)
  vim.deprecate("get_user_message", "parse_messages", "0.1.0", "avante.nvim")
  return table.concat(
    vim.iter(opts.messages):filter(function(_, value) return value.role == "user" end):fold({}, function(acc, value)
      acc = vim.list_extend({}, acc)
      acc = vim.list_extend(acc, { value })
      return acc
    end),
    "\n"
  )
end
```

aarnphm (commented):
tiny

BruceChen7:
> Are you using get_user_message in your custom providers?
> 
> since I don't think we use this anywhere

Yes, I use it in custom providers.

lua/avante/providers/openai.lua
aarnphm: oh I forgot it is supposed to be value.content

lua/avante/providers/openai.lua
```diff
@@ -39,7 +39,17 @@ M.role_map = {
 }
 
 ---@param opts AvantePromptOptions
-M.get_user_message = function(opts) return table.concat(opts.messages, "\n") end
+M.get_user_message = function(opts)
+  vim.deprecate("get_user_message", "parse_messages", "0.1.0", "avante.nvim")
+  return table.concat(
+    vim.iter(opts.messages):filter(function(_, value) return value.role == "user" end):fold({}, function(acc, value)
+      acc = vim.list_extend({}, acc)
+      acc = vim.list_extend(acc, { value })
```
aarnphm, line 47:
```suggestion
      acc = vim.list_extend(acc, { value.content })
```

aarnphm (approved):

PR #818, opened 2024-11-07
fix(compat): filter out non value and not user message

BruceChen7:


BruceChen7:
hihi, @aarnphm  Sorry to bother you. I found there is a problem with the use of "fiter". I have repaired it again.

aarnphm (approved):

PR #840, opened 2024-11-13
feat: :sparkles: Added vertex AI provider for orgs using gemini

ShouryaRSharma:
Organisations or users that use the vertex AI API through google instead of the gemini API need to be able to SSO with gcloud authentication in order to use LLMs. You can see documentation for the API here, viewing the curl tab.

https://cloud.google.com/vertex-ai/generative-ai/docs/start/quickstarts/quickstart-multimodal#send-text-only-request

The vertex AI is gemini compatible (i.e it uses the same API under the hood), so we can use the same gemini parse message and response attributes. 

It would greatly help if this provider could be added, as it would allow me to use avante with models available via the vertex AI model garden.

The changes have been tested locally, and I would be happy to maintain this if the API changes as I regularly use this API and can keep track.

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -5,11 +5,10 @@ local Utils = require("avante.utils")
 
 ---@class avante.CoreConfig: avante.Config
 local M = {}
-
 ---@class avante.Config
 M.defaults = {
   debug = false,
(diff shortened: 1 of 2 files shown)


yetone:
I don't have a vertex testing environment here, but there are no issues with the code, so I merged it first.

PR #846, opened 2024-11-14
fix (sidebar) lock avante input and avante buffers to the sidebar window

brewinski:
While using telescope to open a file or `bNext` to switch buffers the user can mistakenly open a buffer in place of the avante input / avante buffers. Locking the buffers to the window seems to resolve this issue.
image

Alternatively, I can expose winfixbuf as a config option if that makes more sense. Please let me know if you'd like to see this changed.

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -584,6 +584,7 @@ local buf_options = {
 }
 
 local base_win_options = {
+  winfixbuf = true,
   spell = false,
   signcolumn = "no",
   foldcolumn = "0",


poying:
I saw this error when I tried to apply changes. it seems to be related to this pr.

> Error executing vim.schedule lua callback: ...al/share/nvim/lazy/avante.nvim/lua/avante/utils/init.lua:686: E1513: Cannot switch buffer. 'winfixbuf' is enabled

brewinski:
Thanks for the feedback. I noticed this happening when avante was trying to open a buffer in the sidebar / input windows but not on applying changes in all cases..

Happy to revert this change and find a different solution to the origin bug.

jrhoads:
First, thank you for your work on this project.  It is amazing.

Would it be possible to have winfixbuf as a config option?
From v0.0.10 (when this was released) I get an issue rendering the sidebar where it will create multiple of each of the buffers for avante.

I've gone in and disabled this in the sidebar.lua file and things seem to work even on more recent versions (v0.0.16)

It would be nice to be able to do this in the config instead of modifying the plugin.

PR #854, opened 2024-11-15
feat: add Dual Boost Mode for Enhanced LLM Responses with Multi-Provider Support

Lucklyric:
This PR introduces an experimental Dual Boost Mode to avante.nvim, allowing simultaneous response generation from two LLM providers (e.g., OpenAI and Claude). The outputs from these providers are included in a prompt as references for the default provider to generate a final, consolidated response.

Key Changes:

Added Dual Boost Mode configuration options.
Implemented response merging and error handling logic.
Updated stream management to support dual providers.
This feature enables users to leverage multiple providers for more robust responses.

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -112,13 +112,13 @@ M.defaults = {
   ---4. prompt: The prompt to generate response based on the two reference outputs.
   ---5. timeout: Timeout in milliseconds. Default to 60000.
   ---Whow it works:
-  --- When dual_boost is enabled, avante will generate two responses from the first_provider and second_provider respectively. Then use  the response from the first_provider as provider1_output and the response from the second_provider as provider2_output. Finally, avante will generate a response based on the prompt and the two reference outputs, with the default Provider as normal.
+  --- When dual_boost is enabled, avante will generate two responses from the first_provider and second_provider respectively. Then use the response from the first_provider as provider1_output and the response from the second_provider as provider2_output. Finally, avante will generate a response based on the prompt and the two reference outputs, with the default Provider as normal.
   ---Note: This is an experimental feature and may not work as expected.
   dual_boost = {
     enabled = false,
     first_provider = "openai",
     second_provider = "claude",
-    prompt = "Based on the two reference outputs below, generate a response that incorporates elements from both but reflects your own judgment and unique perspective. Do not provide any explanation, just give the response directly. Reference Output 1: [{% provider1_output %}], Reference Output 2: [{% provider2_output %}]",
+    prompt = "Based on the two reference outputs below, generate a response that incorporates elements from both but reflects your own judgment and unique perspective. Do not provide any explanation, just give the response directly. Reference Output 1: [{{provider1_output}}], Reference Output 2: [{{provider2_output}}]",
     timeout = 60000, -- Timeout in milliseconds
   },
   ---Specify the behaviour of avante.nvim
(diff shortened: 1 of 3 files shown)


Lucklyric:
I originally created such tool for my own daily use, and I found that this dual mode was beneficial. Since I am an avante user, I thought, why not integrate it directly into Neovim?

yetone:
‌‌‌‌Looks great!

lua/avante/config.lua
```diff
@@ -105,6 +105,22 @@ M.defaults = {
...
+  ---4. prompt: The prompt to generate response based on the two reference outputs.
+  ---5. timeout: Timeout in milliseconds. Default to 60000.
+  ---Whow it works:
+  --- When dual_boost is enabled, avante will generate two responses from the first_provider and second_provider respectively. Then use  the response from the first_provider as provider1_output and the response from the second_provider as provider2_output. Finally, avante will generate a response based on the prompt and the two reference outputs, with the default Provider as normal.
+  ---Note: This is an experimental feature and may not work as expected.
+  dual_boost = {
+    enabled = false,
+    first_provider = "openai",
+    second_provider = "claude",
+    prompt = "Based on the two reference outputs below, generate a response that incorporates elements from both but reflects your own judgment and unique perspective. Do not provide any explanation, just give the response directly. Reference Output 1: [{% provider1_output %}], Reference Output 2: [{% provider2_output %}]",
```
yetone, line 121: ‌‌‌‌‌How about aligning template variable placeholders with avanterules? Use `{{}}` for all: https://github.com/yetone/avante.nvim/blob/9891b0365607373553b5dab74494a5e81b01e3c2/lua/avante/templates/_context.avanterules#L7

lua/avante/config.lua
```diff
@@ -105,6 +105,22 @@ M.defaults = {
...
+  ---Specify the special dual_boost mode
+  ---1. enabled: Whether to enable dual_boost mode. Default to false.
+  ---2. first_provider: The first provider to generate response. Default to "openai".
+  ---3. second_provider: The second provider to generate response. Default to "claude".
+  ---4. prompt: The prompt to generate response based on the two reference outputs.
+  ---5. timeout: Timeout in milliseconds. Default to 60000.
+  ---Whow it works:
+  --- When dual_boost is enabled, avante will generate two responses from the first_provider and second_provider respectively. Then use  the response from the first_provider as provider1_output and the response from the second_provider as provider2_output. Finally, avante will generate a response based on the prompt and the two reference outputs, with the default Provider as normal.
+  ---Note: This is an experimental feature and may not work as expected.
+  dual_boost = {
```
yetone, line 117: ‌‌‌‌‌‌How about putting this configuration in the README?

lua/avante/config.lua
aarnphm: seems like to me this is generating multiple responses from multiple llm, so probably we can even make it `multi_response` (beyond dual mode)

Lucklyric: I completely agree with you. I conducted some tests and found that the dual-model setup offers reasonable usability without overwhelming the process, as reference outputs cost tokens and context windows. Typically, Claude and ChatGPT are the most common combination. This configuration avoids unnecessary complexity, as unbounded multi-response setups might affect stability and quality when introducing more models. Hence, I believe a dual-model approach is a good starting point for now.

However, I am open to extending this into a multi-agent version with additional recommendation comments in the doc.

I appreciate your suggestions and look forward to more ideas for the next steps.

yetone: ‌‌For now, we can avoid making such significant changes and add multi_response support when there is a second request of this kind.

Issue #861, opened 2024-11-17
bug:  msg_show.emsg E5108: Error executing lua: ...ck/lazy/opt/avante.nvim/lua/avante/providers/copilot.lua:88: internal initialization error


Pger-Y:
### Describe the bug

runtime：
LunarVim latest
neovim 0.10.2
Mac M2 15.1
Provider: copilot
Copilot is on and works find,but when i try to chat with it using avante, error happened
<img width="1773" alt="image" src="https://github.com/user-attachments/assets/59882227-4419-404d-9793-7c53497f01cc">

### Installation method

	{
		"yetone/avante.nvim", --cd ~/.local/share/lunarvim/site/pack/lazy/opt/avante.nvim && make
		event = "VeryLazy",
		lazy = false,
		version = false, -- set this if you want to always pull the latest change
		opts = {
			default = {
			},
			provider = "copilot", -- Recommend using Claude
			auto_suggestions_provider = "copilot",
			-- add any opts here
		},
		-- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
		build = "make",
		-- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
		dependencies = {
			"nvim-treesitter/nvim-treesitter",
			"stevearc/dressing.nvim",
			"nvim-lua/plenary.nvim",
			"MunifTanjim/nui.nvim",
			--- The below dependencies are optional,
			"nvim-tree/nvim-web-devicons", -- or echasnovski/mini.icons
			"zbirenbaum/copilot.lua",   -- for providers='copilot'
			{
				-- support for image pasting
				"HakonHarnes/img-clip.nvim",
				event = "VeryLazy",
				opts = {
					-- recommended settings
					default = {
						embed_image_as_base64 = false,
						prompt_for_file_name = false,
						drag_and_drop = {
							insert_mode = true,
						},
						-- required for Windows users
						use_absolute_path = true,
					},
				},
			},
			{
				-- Make sure to set this up properly if you have lazy=true
				'MeanderingProgrammer/render-markdown.nvim',
				opts = {
					file_types = { "markdown", "Avante" },
				},
				ft = { "markdown", "Avante" },
			},
		},
	}

### Environment

runtime：
LunarVim latest
neovim 0.10.2
Mac M2 15.1
Provider: copilot

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
[...]

PR #862, opened 2024-11-17
fix(copilot): setup M.state when not initialized

ethanh20009:
fixes https://github.com/yetone/avante.nvim/issues/861

Copilot was broken and wouldn't work at all for me. PR resolves it.

### DIFF
--- a/lua/avante/providers/copilot.lua
+++ b/lua/avante/providers/copilot.lua
@@ -119,6 +119,7 @@ end
 M.state = nil
 
 M.api_key_name = ""
+M.setup_without_api_key = true
 M.tokenizer_id = "gpt-4o"
 M.role_map = {
   user = "user",
--- a/lua/avante/providers/init.lua
+++ b/lua/avante/providers/init.lua
@@ -78,6 +78,7 @@ local DressingState = { winid = nil, input_winid = nil, input_bufnr = nil }
 ---@field model? string
 ---@field parse_api_key fun(): string | nil
 ---@field parse_stream_data? AvanteStreamParser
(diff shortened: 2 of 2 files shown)


lua/avante/providers/copilot.lua
```diff
@@ -85,7 +85,7 @@ H.chat_completion_url = function(base_url) return Utils.trim(base_url, { prefix
 local M = {}
 
 H.refresh_token = function()
-  if not M.state then error("internal initialization error") end
+  if not M.state then M.setup() end
```
aarnphm, line 88: ugh this should be setup before hand, not to be called here.

aarnphm (changes requested):
I don't think this fix it. let's investigate

lua/avante/providers/copilot.lua
ethanh20009: Found the root cause:
Copilot does not have an api_key_name as is setup in other plugin. 
In provider init, it checks for no api_key_name and early returns. In our case this is wrong as we need copilot setup.

So added a bool for providers that are set up without api keys.

aarnphm:
cc @yetone

errnoh:
Tested this with Copilot (gpt-4o-2024-08-06) and this fixes #861 for me. ( running avante https://github.com/yetone/avante.nvim/commit/9891b0365607373553b5dab74494a5e81b01e3c2 )

yetone:
‌‌‌‌Hello, I fixed this issue in this way because I really couldn't find a reason not to set up the provider: https://github.com/yetone/avante.nvim/pull/868

PR #864, opened 2024-11-17
fix(sidebar): fix get_sources when blink.compat is installed

rachartier:
Hello,

`blink.cmp` has a module: `blink.compat`. This module is used to work with `nvim-cmp` sources but does not "emulate" all the functions present in `nvim-cmp`.

As a result, when the autocmds in the sidebar are set up and used (like quitting the sidebar), it is checked whether `require("cmp")` exists. With `blink.compat`, it does indeed exist, but the `get_sources` function used afterward does not, which causes an error: `Error executing lua callback: ...local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:1547: attempt to call method 'get_sources' (a nil value)`

This small fix adds a check for the `get_sources` method. If it does not exist, the rest of the function is aborted.

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -1544,9 +1544,11 @@ function Sidebar:create_input(opts)
     callback = function()
       local has_cmp, cmp = pcall(require, "cmp")
       if has_cmp then
-        for _, source in ipairs(cmp.core:get_sources()) do
-          if source.name == "avante_commands" or source.name == "avante_mentions" then
-            cmp.unregister_source(source.id)
+        if type(cmp.core.get_sources) == "function" then
+          for _, source in ipairs(cmp.core:get_sources()) do
+            if source.name == "avante_commands" or source.name == "avante_mentions" then
+              cmp.unregister_source(source.id)
+            end
           end
         end
       end


lua/avante/sidebar.lua
```diff
@@ -1544,9 +1544,11 @@ function Sidebar:create_input(opts)
     callback = function()
       local has_cmp, cmp = pcall(require, "cmp")
       if has_cmp then
-        for _, source in ipairs(cmp.core:get_sources()) do
-          if source.name == "avante_commands" or source.name == "avante_mentions" then
-            cmp.unregister_source(source.id)
+        if type(cmp.core.get_sources) == "function" then
```
aarnphm, line 1547: Let's not do this. probably better to have cmp backend (we should add support to blink fwiw)

then update the callback here instead of doing this (otherwise maintenance cost are too high)

aarnphm (changes requested):
tiny

doodleEsc:
tried with `blink.compat` and it works fine with first time.

`blink.cmp` call function is_available(), but  `self.bufnr` in `cmp_avante.commands` and `cmp_avante.mentions` will never change.

```lua
function commands_source:is_available() return api.nvim_get_current_buf() == self.bufnr end
```

I change `is_available` function
```
function commands_source:is_available() return vim.bo.filetype == "AvanteInput" end
```

#https://github.com/doodleEsc/avante.nvim/commit/a5438d0f16208b7ae9e97ae354bed5ec16b4f9ed


my LazyVim.nvim config, juse add `avante_commands` and `avante_mentions` in source.compat
```lua
  {

    "saghen/blink.cmp",
    opts = {
      sources = {
        compat = { "avante_commands", "avante_mentions" },
      },
    },
  },

```

it works for me

yetone:
‌Thank you for your contribution. Since this PR hasn't been updated for a long time, it has been closed. Sorry about that.

Issue #882, opened 2024-11-21
bug: Fail to load avante_repo_map


dinggul:
### Describe the bug

```
LUA_VERSION=luajit bash ./build.sh
+ mkdir -p /home/dinggul/.local/share/nvim/lazy/avante.nvim/build
+ curl -L https://github.com/yetone/avante.nvim/releases/download/v0.0.9/avante_lib-linux-aarch64-luajit.tar.gz
+ tar -zxv -C /home/dinggul/.local/share/nvim/lazy/avante.nvim/build
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0 9714k    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0avante_repo_map.so
  9 9714k    9  908k    0     0   494k      0  0:00:19  0:00:01  0:00:18 1173kavante_templates.so
avante_tokenizers.so
100 9714k  100 9714k    0     0  4043k      0  0:00:02  0:00:02 --:--:-- 7249k
```

```
file .local/share/nvim/lazy/avante.nvim/build/avante_repo_map.so
.local/share/nvim/lazy/avante.nvim/build/avante_repo_map.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=b5fd49939426d21b29a5047270aa0c082ca47978, not stripped
```

I tried to install avante on my aarch64 machine, but this download link seems to contain x86-64 binary

This causes lib load error

```
Error executing vim.schedule lua callback: ...ocal/share/nvim/lazy/avante.nvim/lua/avante/repo_map.lua:23: Failed to load avante_repo_map
stack traceback:
        [C]: in function 'error'
        ...ocal/share/nvim/lazy/avante.nvim/lua/avante/repo_map.lua:23: in function ''
        vim/_editor.lua: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
```

### Installation method

Use lazy.nvim:
```lua
  {
    "yetone/avante.nvim",
    event = "VeryLazy",
    lazy = false,
    version = false, -- set this if you want to always pull the latest change
    opts = {
      -- add any opts here
    },
    -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
    build = "make",
[...]

PR #909, opened 2024-11-26
Pipline aarch64 fix

Moerliy:
Mentioned in issue #882 the binaries provided for linux-aarch64 are currently x86_64 architecture. 

This PR changes the GitHub CI pipeline to use QEMU to emulate aarch64 architecture to compile the right bin's.
One thing I noticed is that the pipeline step for aarch64 will now take much longer (obviously because of the emulation). 

Maybe there is a better way but I just wanted to get the ball rolling on this.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -54,14 +54,16 @@ jobs:
         config:
           - os: ubuntu-latest
             os_name: linux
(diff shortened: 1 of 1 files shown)


ghosert:
thanks the latest 0.0.12 fixed the issue.

PR #912, opened 2024-11-27
feat(context): add a ui for selecting and adding files to the sidebar as context

brewinski:
I've been playing around with some tools to allow users to add other files as context in the sidebar. So far I have something that's relatively stable and seems to work.

I'm at a point where it would be good to have some feedback on the direction I'm going with this as I'd like to make sure my work aligns with the vision of the core maintainers. As always feedback and change requests are welcome.

So far this is what I have:

Opening the context menu from the sidebar with "c" by default
image

Searching for files be pressing "s" in the context window by default. 
image

Displaying the context used in the chat and saving that as part of the chat history. 
image

Search and Replace blocks in the current file and context files.
Example Output:
~~~markdown
---

- Datetime: 2024-11-27 15:17:56

- Model: copilot/claude-3.5-sonnet

- Selected file: lua/avante/context.lua

- Context files:
  - ./lua/avante/providers/copilot.lua
  - ./lua/avante/providers/claude.lua


> Add a remove by index to context.lua. add a binary search function to the copilot.lua module and add a bubble sort function to the claude module. 
> 

I'll help add these functions. Let me make the changes one by one.

1. First, let's add the remove by index function to context.lua:

Filepath: lua/avante/context.lua
Replace lines: 294-305
```lua
---@return nil
function Context:remove_context()
  local current_string = self:get_full_string_under_cursor()
  for i, file_path in ipairs(self.context_files) do
    if vim.fs.basename(file_path) == current_string then
      table.remove(self.context_files, i)
      break
    end
  end
  self:update_file_cache()
  self:render()
end

---@param index integer
---@return nil
function Context:remove_by_index(index)
  if index > 0 and index <= #self.context_files then
    table.remove(self.context_files, index)
    self:update_file_cache()
    self:render()
  end
end
```

2. Let's add the binary search function to copilot.lua:

Filepath: lua/avante/providers/copilot.lua
Replace lines: 164-164
```lua
---@param arr number[] The sorted array to search in
---@param target number The value to find
---@return integer|nil index The found index or nil if not found
M.binary_search = function(arr, target)
  local left = 1
  local right = #arr
  
  while left <= right do
    local mid = math.floor((left + right) / 2)
    if arr[mid] == target then
      return mid
    elseif arr[mid] < target then
      left = mid + 1
    else
      right = mid - 1
    end
  end
  
  return nil
end

M.setup = function()
```

3. Let's add the bubble sort function to claude.lua:

Filepath: lua/avante/providers/claude.lua
Replace lines: 31-31
```lua
---@param arr number[] The array to sort
---@return number[] sorted The sorted array
M.bubble_sort = function(arr)
  local n = #arr
  for i = 1, n do
    for j = 1, n - i do
      if arr[j] > arr[j + 1] then
        arr[j], arr[j + 1] = arr[j + 1], arr[j]
      end
    end
  end
  return arr
end

M.parse_messages = function(opts)
```

These changes add:
1. A remove_by_index function to Context that removes a context file by its index
2. A binary search implementation to the copilot provider
3. A bubble sort implementation to the claude provider

The functions follow the existing code style and include type annotations where appropriate.

**Generation complete!** Please review the code suggestions above.
~~~


lua/avante/llm.lua
```diff
@@ -72,6 +72,13 @@ M._stream = function(opts, Provider)
     if diagnostics ~= "" then table.insert(messages, { role = "user", content = diagnostics }) end
   end
 
+  if opts.code_context ~= nil then
+    for key, ctx in pairs(opts.code_context) do
+      local pj_ctx = key .. " " .. ctx
+      table.insert(messages, { role = "user", content = "PROJECT CONTEXT: " .. pj_ctx })
```
brewinski, line 78: I'm considering creating a .avanterules file context files.

yetone:
‌‌Perhaps `@file` is a good UX.

lua/avante/llm.lua
yetone: ‌‌‌You can use this template: https://github.com/yetone/avante.nvim/blob/16c4254b14ebe8acb79f6abecb49ca36b9d96cb0/lua/avante/templates/_project.avanterules#L3

radwo:
Great work!

When we have this context, maybe we can detach chat from the single file and operate on the context? The same thing as Cursor is working right now.

CleanShot 2024-11-27 at 18 21 20@2x

yetone:
@brewinski Hello, I will refactor this PR on a large scale to implement `@files`.

brewinski:
No problem @yetone, I haven't had time to action the feedback yet but I might have some time tonight :).

I think @file is a good approach.

brewinski:
@yetone I had some time last night so I spent it getting the @file menu working in the AvanteInput buffer. Still some work to be done getting it to actually provide the file as context.

image
image

lua/avante/sidebar.lua
```diff
@@ -156,13 +159,21 @@ end
 ---@param original_content string
 ---@param result_content string
 ---@param code_lang string
+---@param context_code table
 ---@return AvanteReplacementResult
-local function transform_result_content(original_content, result_content, code_lang)
+local function transform_result_content(original_content, result_content, code_lang, context_code)
```
yetone, line 164: ‌‌‌‌‌It seems unnecessary to add a new `context_code`; using `original_content` to carry the content of different files will suffice.

lua/avante/llm.lua
yetone: ‌‌‌‌‌‌I was thinking, it would be more appropriate to place the selected files in here: https://github.com/yetone/avante.nvim/blob/57311bf8cd2f48729565d2351bcbf383b6a56907/lua/avante/templates/_context.avanterules#L2

lua/avante/llm.lua
```diff
@@ -342,6 +349,7 @@ end
 ---@field file_content string
 ---@field selected_code string | nil
 ---@field project_context string | nil
+---@field code_context table | nil
```
yetone, line 352: The options need to be refactored by removing `file_content` and `code_context`, and changing it to `selected_files`, where the type of `selected_files` is `{ path: string, content: string}[]`.

yetone:
image

This style is quite strange. Can we rely solely on markdown to render these contents? Similar to codecompanion:
image

brewinski:
Thanks for the feedback @yetone I looked at CodeCompanion and looked at driving this via the markdown but couldn't get a consistent experience.

Instead I looked at cursor for inspiration and this workflow seems to work reasonably well.
image
image
image

lua/avante/llm.lua
```diff
@@ -342,6 +359,7 @@ end
 ---@field file_content string
```
yetone, line 359: ‌‌‌‌‌‌Can the `file_content` be deleted?

yetone:
> Thanks for the feedback @yetone I looked at CodeCompanion and driving this via the markdown but couldn't get a consistent experience. This project was very helpful in working out how to get the @file callback working and integrating callbacks into mentions.lua
> 
> The best I've managed to come up with is slightly closer to the Cursor experience so hopefully this is reasonable intuitive. image image image image

Great, thank you for your work! It looks very elegant now. How do I delete a selected file in this situation?

brewinski:
> > Thanks for the feedback @yetone I looked at CodeCompanion and driving this via the markdown but couldn't get a consistent experience. This project was very helpful in working out how to get the @file callback working and integrating callbacks into mentions.lua
> > The best I've managed to come up with is slightly closer to the Cursor experience so hopefully this is reasonable intuitive. image image image image
> 
> Great, thank you for your work! It looks very elegant now. How do I delete a selected file in this situation?

Thank you! Currently if your cursor if over the file path and your press `d` it will remove the file. This could be improved by adding some hint's / ext marks to make it obvious.

lua/avante/llm.lua
brewinski: I tried removing this yesterday but it caused an error on line 60 of llm.lua. I suspect it's related to the rust crate but I haven't had a chance to dig into it yet.

I'll try and take a look at this tomorrow.

- lua/avante/llm.lua:60:25
- crates/avante-templates/src/lib.rs:48:19
- crates/avante-templates/src/lib.rs:24:5

lua/avante/llm.lua
yetone: Delete these two fields, then add `selected_files`, and if you encounter any issues during the modification process, feel free to provide feedback on Discord.

https://github.com/yetone/avante.nvim/blob/57311bf8cd2f48729565d2351bcbf383b6a56907/crates/avante-templates/src/lib.rs#L23-L24

lua/avante/llm.lua
yetone: Indeed, it is necessary to simultaneously modify the related Rust code. You can refer to a similar PR: https://github.com/yetone/avante.nvim/pull/891/files#diff-83be62c82ad7d2b73e8a65f2437760bfc972f12d3b527d78554fba74049e4601

brewinski: Thanks. I have removed file content but I had to leave it in the rust code as I'm using the value to insert a message per selected file.

This options has the benefit of making it less likely that a message will exceed the llm token limit. The downside is, a file might be excluded from the context.

```lua
  if opts.selected_files ~= nil then
    for _, selected_file in ipairs(opts.selected_files) do
      local code_context = Path.prompts.render_file(
        "_context.avanterules",
        vim.tbl_extend("force", template_opts, {
          filepath = selected_file.path,
          file_content = selected_file.content,
        })
      )
      if code_context ~= "" then table.insert(messages, { role = "user", content = code_context }) end
    end
  end
```

Is it better to change the `_context.avanterules` template to insert all context files via a for loop?

brewinski:
commit adds support for code suggestions in multiple languages.
image

aarnphm:
A small ask: can you make the choose modal as a side panel on non-float?

brewinski:
> A small ask: can you make the choose modal as a side panel on non-float?

@aarnphm happy to look into it. Do you have an example of what you're looking for? If you do, I'll see if I can make it work.

aarnphm:
oh I was thinking to just append a panel on top of the input modal.

Sth like this

image

lua/avante/sidebar.lua
```diff
@@ -1108,7 +1116,7 @@ function Sidebar:update_content(content, opts)
   end
   if opts.stream then
     local scroll_to_bottom = function()
-      local last_line = api.nvim_buf_line_count(self.result.bufnr)
+      local last_line = api.nvim_buf_line_coun(self.result.bufnr)
```
yetone, line 1119: This is a typo

lua/avante/sidebar.lua
```diff
@@ -1222,7 +1241,8 @@ function Sidebar:render_history_content(history)
       entry.model,
       entry.request or "",
       entry.selected_file,
-      entry.selected_code
+      entry.selected_code,
+      entry.context_summary
```
yetone, line 1245: ‌‌Shouldn't we avoid transmitting the context summary here and treat all selected files equally, rather than having some as summary context and others as selected files? Otherwise, if the file content passed to LLM is incomplete, how can LLM modify all selected files?

lua/avante/sidebar.lua
brewinski: Thank you for picking this one up.

lua/avante/llm.lua
brewinski: @yetone I have updated the rust template code and the _context.avanterules file to handle an array of files. Hopefully this works how you expect it to work now.

lua/avante/sidebar.lua
brewinski: @yetone that seems like a reasonable suggestion. Would it be better change `selected_file` to `selected_files` and render all of the selected files in a single list.

E.g: this

- Selected file: lua/avante/context.lua

- Selected files:
  - ./crates/avante-repo-map/src/lib.rs
  - ./lua/avante/providers/copilot.lua

would become this:
- Selected files:
  - ./lua/avante/context.lua
  - ./crates/avante-repo-map/src/lib.rs
  - ./lua/avante/providers/copilot.lua

brewinski:
> oh I was thinking to just append a panel on top of the input modal.
> 
> Sth like this
> 
> image

Thanks @aarnphm, I'm not quite sure how this would work but id be happy if you wanted to alter the input control. 

I originally went with vim.ui.select because it makes it easy to extend in the future so that users can override it with their own file picker. (telescope/fzf-lua). I have a local branch where I'm able to configure telescope's builtin file picker.

An alternative to changing the UI could be to go back to using cmp as the file picker menu as seen in [this comment](https://github.com/yetone/avante.nvim/pull/912#issuecomment-2508718524)?

lua/avante/sidebar.lua
yetone: ‌Yes, all the selected files should be at the same level in a single list, which makes more sense.

yetone:
@brewinski Hello, I just refactored and fixed some bugs. Now the code structure and UI/UX are more in line with my expectations. I'm ready to merge it now. Thank you so much for your hard work over the past two weeks, bringing such powerful features to avante.nvim!

image

dogonthehorizon:
This is great, thanks for your work @brewinski and @yetone to get this merged. I've been testing this over the last few hours and it's working as described. The pattern of adding files w/ `@` and switching to a file picker is really slick. What I'm longing for now is being able to add the currently focused buffer to the file list with a keybind.

I took a stab at implementing it here:
https://github.com/yetone/avante.nvim/pull/941

I'm not married to the key binding, if there's one that makes more sense I'm happy to make that happen!

brewinski:
Thank you @yetone for all the time you spent providing feedback. The refactor is looking and working great.

@dogonthehorizon that's a cool idea.

PR #926, opened 2024-12-05
fix: load path

yetone:


yetone:
‌‌‌This stupid bug caused this [issue](https://github.com/yetone/avante.nvim/issues/882), and it's strange that I can't reproduce it on my side.

PR #930, opened 2024-12-09
feat(AvanteClear): accept args to clear history, memory or cache

b0o:
Usage:
```
:AvanteClear [history|memory|cache]
```

- changed default behavior from clearing the cache to clearing history.
- added args to clear history, memory or cache.
- added confirm dialog before clearing cache.


yetone:
‌‌‌‌‌‌‌This is useful.

Issue #933, opened 2024-12-09
bug: Copilot Token expires on long lived sessions 


StevenMatchett:
### Describe the bug

After session has been left open for a long time github copilot token expires. Sending a new requests results in `Error: 'API request failed with status 401. Body: "unauthorized: token expired"'`. Once the request is resubmitted the request is successful.

### Installation method

Use lazy.nvim:
```lua
	{
		"yetone/avante.nvim",

		lazy = false,
		dependencies = {
			"nvim-treesitter/nvim-treesitter",
			"stevearc/dressing.nvim",
			"nvim-lua/plenary.nvim",
			"MunifTanjim/nui.nvim",
			"nvim-tree/nvim-web-devicons",
			"zbirenbaum/copilot.lua",
			{
				"MeanderingProgrammer/render-markdown.nvim",
				opts = { file_types = { "markdown", "Avante" } },
				ft = { "markdown", "Avante" },
			},
		},
		build = "make",
		config = function()
			require('avante_lib').load()
			require('avante').setup({
				provider = "copilot",
				auto_suggestions_provider = "copilot",
				behaviour = {
					-- auto_suggestions = true, -- Enable auto suggestions
				},
				mappings = {
					suggestion = {
						-- accept = "<Tab>",
						-- next = "<C-;>",
						-- prev = "<C-'>",
						-- dismiss = "<C-]>",
					},
				},
				windows = {
					position = "right",
					width = 30,
				}
			})
		end,
	},
```

### Environment

NVIM v0.10.1
Build type: Release
LuaJIT 2.1.1713484068

on latest release of avante
`  "avante.nvim": { "branch": "main", "commit": "57311bf8cd2f48729565d2351bcbf383b6a56907" },`

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

jones-sam:
Just submitted a PR fixing this 👍

PR #918, opened 2024-11-30
Fix: Get copilot refresh token asynchronously

arkuna23:
Getting copilot refresh token will freeze the process if response takes too much time.

Issue #809, opened 2024-11-05
bug: Token Expiration for Copilot Provider


traysh:
### Describe the bug

After having neovim idle for a while, asking a question to Avante configured with the Copilot provider will result in API request failed with status 401. This bug was reported before: https://github.com/yetone/avante.nvim/issues/691. I suppose the fix did not work, at least not in all cases.

My configs are:

plugins.lua:
```
  {
    'zbirenbaum/copilot.lua',
    event = 'InsertEnter',
    config = function()
      require('configs.copilot')
    end,
  },
  {
    "yetone/avante.nvim",
    event = "VeryLazy",
    lazy = false,
    version = false, -- set this if you want to always pull the latest change
    opts = require("configs.avante"),
    -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
    build = "make",
    -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
    dependencies = {
      "nvim-treesitter/nvim-treesitter",
      "stevearc/dressing.nvim",
      "nvim-lua/plenary.nvim",
      "MunifTanjim/nui.nvim",
      --- The below dependencies are optional,
      "nvim-tree/nvim-web-devicons",
      "zbirenbaum/copilot.lua",
      -- Make sure to set this up properly if you have lazy=true
      'MeanderingProgrammer/render-markdown.nvim',
    },
  },
```

avante.lua:
```
return {
  provider = "copilot",                  -- Recommend using Claude
  auto_suggestions_provider = "copilot", -- Since auto-suggestions are a high-frequency operation and therefore expensive, it is recommended to specify an inexpensive provider or even a free provider: copilot
  behaviour = {
    auto_suggestions = false, -- Experimental stage
    auto_set_highlight_group = true,
    auto_set_keymaps = true,
    auto_apply_diff_after_generation = false,
    support_paste_from_clipboard = false,
  },
  mappings = {
    --- @class AvanteConflictMappings
    diff = {
      ours = "co",
      theirs = "ct",
      all_theirs = "ca",
      both = "cb",
      cursor = "cc",
      next = "]x",
      prev = "[x",
    },
    suggestion = {
      accept = "<M-l>",
      next = "<M-]>",
      prev = "<M-[>",
      dismiss = "<C-]>",
    },
    jump = {
      next = "]]",
      prev = "[[",
    },
    submit = {
      normal = "<CR>",
      insert = "<C-s>",
    },
[...]

PR #935, opened 2024-12-10
fix(copilot): refreshing copilot tokens

jones-sam:
https://github.com/yetone/avante.nvim/issues/933

When the copilot token expired, the refresh token function would get called but it would happen async so that the completion would return a 401 since it would use still the expired token. The next request would then work because the async call to update it would then be completed. This would happen every time after the token expired, which was quite frequent since it expires every 30 min. 

This PR fixes it so it gets called synchronously, if the token is expired it will update it first before then calling the completion endpoint.

### DIFF
--- a/lua/avante/providers/copilot.lua
+++ b/lua/avante/providers/copilot.lua
@@ -35,7 +35,63 @@ local O = require("avante.providers").openai
 
 local H = {}
 
+---@class AvanteProviderFunctor
+local M = {}
+
 local copilot_path = vim.fn.stdpath("data") .. "/avante/github-copilot.json"
+local lockfile_path = vim.fn.stdpath("data") .. "/avante/copilot-timer.lock"
+
+-- Lockfile management
+local function is_process_running(pid)
+  if vim.fn.has("win32") == 1 then
+    return vim.fn.system('tasklist /FI "PID eq ' .. pid .. '" 2>NUL | find /I "' .. pid .. '"') ~= ""
+  else
+    return vim.fn.system("ps -p " .. pid .. " > /dev/null 2>&1; echo $?") == "0\n"
+  end
+end
+
+local function try_acquire_timer_lock()
+  local lockfile = Path:new(lockfile_path)
+
+  local tmp_lockfile = lockfile_path .. ".tmp." .. vim.fn.getpid()
+
+  Path:new(tmp_lockfile):write(tostring(vim.fn.getpid()), "w")
+
+  -- Check existing lock
+  if lockfile:exists() then
+    local content = lockfile:read()
+    local pid = tonumber(content)
+    if pid and is_process_running(pid) then
+      os.remove(tmp_lockfile)
+      return false -- Another instance is already managing
+    end
+  end
+
+  -- Attempt to take ownership
+  local success = os.rename(tmp_lockfile, lockfile_path)
+  if not success then
+    os.remove(tmp_lockfile)
+    return false
+  end
+
+  return true
+end
+
+local function start_manager_check_timer()
+  if M._manager_check_timer then
+    M._manager_check_timer:stop()
+    M._manager_check_timer:close()
+  end
+
+  M._manager_check_timer = vim.uv.new_timer()
+  M._manager_check_timer:start(
+    30000,
+    30000,
+    vim.schedule_wrap(function()
+      if not M._refresh_timer and try_acquire_timer_lock() then M.setup_timer() end
+    end)
+  )
+end
 
 ---@class OAuthToken
 ---@field user string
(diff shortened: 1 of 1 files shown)


b0o:
This has been going back and forth a lot. Running `plenary.curl` synchronously will lock up the UI until the request completes. I think there's a better way to structure the code so we can refresh asynchronously without causing the bug.

See:
- https://github.com/yetone/avante.nvim/issues/809
- https://github.com/yetone/avante.nvim/pull/918
- https://github.com/yetone/avante.nvim/pull/791

jones-sam:
Alright, how about this?

- Added ability to do both sync and async requests depending on the context
- Created timer to refresh token every 28 minutes to prevent ever having an outdated token
- Added lockfile so that only one instance of neovim can have a timer to refresh the token so that we don't send unneeded requests
- Added another timer to check if instance of neovim that was refreshing the token was closed, and take it over
- Added synchronous call to request token before calling the completion endpoint to prevent 401 only if the token is expired, however this should rarely happen since it should be upgraded in the background asynchronously before that ever gets triggered. Basically it is just a backup

tomaszferens:
Can we merge it? Currently, I have to run the prompts twice because the first one always fails with "token expired" error

aarnphm (commented):
I can chime in here, I think we should think of a permanent fix other than this

We probably still want to do async refresh without blocking the main UI.

jones-sam:
> I can chime in here, I think we should think of a permanent fix other than this
> 
> We probably still want to do async refresh without blocking the main UI.

This should be a permanent fix as long as github keeps their auth process the same. The token gets refreshed async in the background to make sure it never expires, as far as I am aware that doesn't block any UI or cause any other issues. If we do it async right before sending a completion like it is in the main branch now, it is always guaranteed to fail after since the async request is not finished before the completion request, meaning the new token doesn't get used, resulting in a 401. If we do a sync request before, it will always freeze up, waiting for the request to go through. With this implementation, it does it async in the background, so neither of those problems arise.

I've been using my branch of avante since I created this PR and it has been working well, no issues. I think we should merge this soon because I imagine everyone who is using copilot is running into 401s every 30 min. Are there any changes you'd like in this code? I'm not sure what other approach could be made to solve the problem other than this.

tomaszferens:
I think what you did @jones-sam is great, much better than status quo we have today.

albankurti:
I've been using this PR's implementation for a few days and it works well. It's much better than having to yank and re-paste queries.

lua/avante/providers/copilot.lua
```diff
@@ -81,35 +137,59 @@ end
...
+      if not async and M._refresh_timer then M.setup_timer() end
+
+      return true
+    else
+      error("Failed to get success response: " .. vim.inspect(response))
+      return false
+    end
+  end
+
+  if async then
```
aarnphm, line 183: why can't we just use async here? In what case where we don't use async?

lua/avante/providers/copilot.lua
```diff
@@ -162,20 +244,110 @@ M.parse_curl_args = function(provider, code_opts)
...
+  -- Cleanup file watcher
+  if M._file_watcher then
+    M._file_watcher:stop()
+    M._file_watcher = nil
+  end
+end
+
+-- Register cleanup on Neovim exit
+vim.api.nvim_create_autocmd("VimLeavePre", {
+  callback = function() M.cleanup() end,
```
aarnphm, line 350: wait why do we need to register the cleanup here?

afaik once you exited neovim all plugins should be de-registered. Usually cleanup are only meant for during neovim session.

ig we need to remove the lock file.

aarnphm (approved):
Ok I'm good with merging this for now, given that a lot of folks are using copilot.

I will probably look into this a bit more.

lua/avante/providers/copilot.lua
jones-sam: There is a check to see if the token is expired before sending a completion request, and if it is, then it will send a synchronous request to get a new token before sending the completion request. This is because if the token is expired, the completion request will always return a 401. If we send it async in this case, the completion request will also always fail since it won't complete in time for the following request. 

This shouldn't happen often since it gets refreshed in the background to always keep it up-to-date, but sometimes it will be needed if the background request fails due to the internet being down or another error.

lua/avante/providers/copilot.lua
jones-sam: I added it so the cleanup would run and remove the lockfile when closing neovim to clean things up right away. But you are right, things would still work without this, however the lockfile would remain with the pid of the closed instance. But once another instance checks the lockfile and sees that that pid is no longer running, it will overwrite it and take over. Up to you if you want to keep this or not, I believe it will still work properly with or without this.

aarnphm: Yeah I don't have a big problem with this. (I might be wrong), but from the documentations it seems like it should only be used with shada or sth

```prolog
VimLeavePre	    Before exiting Vim, just before writing the
				.shada file.  This is executed only once,
				if there is a match with the name of what
				happens to be the current buffer when exiting.
				Mostly useful with a "*" pattern. >
				   :autocmd VimLeavePre * call CleanupStuff()
<				Use |v:dying| to detect an abnormal exit.
				Use |v:exiting| to get the exit code.
				Not triggered if |v:dying| is 2 or more.

```

aarnphm (approved):

PR #971, opened 2024-12-19
feat (file_selector): Add multi select

esezen:
Allows the user to select multiple at once when using the Telescope selector. Uses the default keybind of tab to select and `<CR>` confirms the selection

multi_picker


yetone:
‌‌Looks great! Sorry for merging so late!

culda:
<Tab> doesn't select the file for me. I can't find the code for the key bindings config
<Tab> selects a line in Telescope fine

culda:
Might be related. In Telescope, Tab works in --Insert-- mode, but in the fzf selector it's --Terminal-- mode

garrettmaring:
When I tab select multiple files and hit <CR>, it single selects whatever I'm hovered over at the time

PR #974, opened 2024-12-20
feat(sidebar): support file context when buffer is outside the cwd

brewinski:
When you open a file that is outside of the current project directory/working directory, the file context for the current buffer is incorrect.

When opening a random file for example by running `:e ../../some-random-source-file.rs` the path would resolve to some-random-source-file.ts rather than the absolute path and the file context would be unable to read it.

Before:
image

After:
image

### DIFF
--- a/lua/avante/file_selector.lua
+++ b/lua/avante/file_selector.lua
@@ -267,4 +267,16 @@ end
 
 function FileSelector:get_selected_filepaths() return vim.deepcopy(self.selected_filepaths) end
 
+---@return nil
+function FileSelector:add_quickfix_files()
+  local quickfix_files = vim
+    .iter(vim.fn.getqflist({ items = 0 }).items)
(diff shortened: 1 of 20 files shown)


lua/avante/file_selector.lua
```diff
@@ -35,7 +35,9 @@ end
 function FileSelector:add_selected_file(filepath)
   if not filepath or filepath == "" then return end
 
-  local uniform_path = Utils.uniform_path(filepath)
+  -- if the file is in the current working directory then we can make it uniform the path
+  local uniform_path = Utils.file.is_in_cwd(filepath) and Utils.uniform_path(filepath) or filepath
```
yetone, line 39: ‌‌‌‌Can this logic be placed in the Utils.uniform_path function?

brewinski: Yeah I think that's fine. I've move the logic into the uniform path function and everything seems to be working.

Issue #970, opened 2024-12-19
bug: Chat with no files does not work


CyprienRicque:
### Describe the bug

When I delete the last file using `d`, the sidebar completely closes and I can't use the chat.

### To reproduce

Delete all files in the sidebar

### Expected behavior

No files are used and it works just fine.

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false, -- set this if you want to always pull the latest change
  ---@class avante.Config
  opts = {
    provider = "claude",
    auto_suggestions_provider = "copilot", -- Since auto-suggestions are a high-frequency operation and therefore expensive, it is recommended to specify an inexpensive provider or even a free provider: copilot
    ---@type AvanteSupportedProvider
    claude = {
      endpoint = "https://api.anthropic.com",
      model = "claude-3-5-sonnet-20241022",
      temperature = 0,
      max_tokens = 4096,
    },

    mappings = {
      -- Mappings for handling merge conflicts
      --- @class AvanteConflictMappings
      diff = {
        ours = "co",          -- Choose our version
        theirs = "ct",        -- Choose their version
        all_theirs = "ca",    -- Choose all their versions
        both = "cb",          -- Keep both versions
        cursor = "cc",        -- Choose version under cursor
        next = "]x",          -- Go to next conflict
        prev = "[x",          -- Go to previous conflict
      },
      -- Mappings for handling suggestions
      suggestion = {
        accept = "<M-l>",     -- Accept current suggestion
        next = "<M-k>",       -- Go to next suggestion
        prev = "<M-j>",       -- Go to previous suggestion
        dismiss = "<M-h>",    -- Dismiss current suggestion
      },
      -- Mappings for jumping between sections
      jump = {
        next = "]]",          -- Jump to next section
        prev = "[[",          -- Jump to previous section
      },
      -- Mappings for submitting content
      submit = {
        normal = "<CR>",      -- Submit in normal mode
        insert = "<M-j>",     -- Submit in insert mode
      },
      ask = "<leader>aj",     -- Ask a question
      edit = "<leader>ae",    -- Edit content
      refresh = "<leader>a-", -- Refresh content
      focus = "<leader>af",   -- Focus on content
      -- Toggle various features
      toggle = {
        default = "<leader>at",    -- Toggle default view
        debug = "<leader>ad",      -- Toggle debug mode
        hint = "<leader>ah",       -- Toggle hints
[...]

PR #975, opened 2024-12-20
feat(sidebar): allow chat without any files selected.

brewinski:
Small change to allow a user to remove all files without closing the sidebar. I've created this as a result of this issue: https://github.com/yetone/avante.nvim/issues/970.

I'm not sure if it was intended that the sidebar would close when the file context was removed as the WinClosed auto command is closing the entire sidebar.

Happy to find another way if we'd prefer to have the selected files container not visible after all files have been deleted.

Raised PR as a result of this isssue: https://github.com/yetone/avante.nvim/issues/970

After:
image

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -2054,11 +2054,6 @@ function Sidebar:create_selected_files_container()
   local render = function()
     local selected_filepaths_ = self.file_selector:get_selected_filepaths()
 
-    if #selected_filepaths_ == 0 then
-      self.selected_files_container:unmount()
-      return
-    end
-
     local selected_filepaths_with_icon = {}
     for _, filepath in ipairs(selected_filepaths_) do
       local icon = Utils.file.get_file_icon(filepath)


brewinski:
> With this change only, the prompt sent to the llm needs to be changed. 
> For instance it contains 
> ```
> [...]
> All changes to files must use this *SEARCH/REPLACE block* format.
> ONLY EVER RETURN CODE IN A *SEARCH/REPLACE BLOCK*!
> [...]
> ```
> If there are no files, it does not make sense

@CyprienRicque that's a good point. It should be fairly simple to alter the system prompt. 

Does this warrant introducing a new mode such as "chatting" with a different system prompt & avanterules file? Or should we simply omit parts of the prompt when there are not context files?

Excuse my ignorance but what would the impact of leaving those references in the system prompt?

CyprienRicque:
I think that having extra useless guidelines does in fact alter the model performance, but only from experience using these models.
I believe adding a `chatting` mode with dedicated avanterules file is a great idea.

yetone:
‌‌‌Thank you for this PR. I will implement this feature in another way, so I'll close it for now.

Issue #973, opened 2024-12-19
feature: Allow for focusing main code window while a response is being streamed


johnsaigle:
### Feature request

Currently when using Avante Ask the streaming response will "steal" focus in the main window as the model's text is populated in the buffer.

My preference would be to have the ability to navigate in the main code window while the response streams on the side. Right now this isn't possible because the cursor and focus jumps to the response buffer.

#### Example 1
While using Avante Ask, the response is streamed and each new bit of text causes the cursor to move away from the main window into the Avante Ask window.

#### Example 2
When using Avante Edit, the small prompt window steals focus. Instead of navigating the code, new keystrokes are appended to the prompt window even after the prompt has been sent.

albankurti:
fully agreed, there is very little utility to having the attention on the chat as a response is generated, even worse - you can't step out of the chat it keeps stealing the focus :)

albankurti:
Created [PR 982](https://github.com/yetone/avante.nvim/pull/982) which should allow you to have this behavior. 

The config should be updates this way:
```
  {
    "yetone/avante.nvim",
    ...
    opts = {
      ...
      behaviour = {
        auto_focus_sidebar = false,
      },
      ...
  },
```

If you would like to test it and provide feedback it would be nice :) Just switch the .local/share/nvim/lazy/avante.nvim to the branch in the PR, or wherever your packages are located

braveltd:
great work guys!

JunYang-tes:
hi @albankurti , thanks for your contribution. 
by setting auto_focus_sidebar to true, the input area won't focused when I open the sidebar.I think this is a little inconvenient because we open sidebar  usually means we need to input something. I personally prefer have a option called auto_focus_result_window to control if the result window will be focused or not. what do you think?

albankurti:
@JunYang-tes I agree, further I noticed that insert mode is not working well during generation. I will attempt to work on these two things.

PR #982, opened 2024-12-21
feat enhancement - Auto Focus Sidebar

albankurti:
Small feature addressing [Issue 973](https://github.com/yetone/avante.nvim/issues/973) to not auto-focus on the chat and be able to code or move around in the editor as the response is being generated


albankurti:
Config would look like the following:
```
  {
    "yetone/avante.nvim",
    ...
    opts = {
      ...
      behaviour = {
        auto_focus_sidebar = false,
      },
      ...
  },
```

albankurti:
I am testing this more, there seems to be a problem, when it is generating and the output is out of focus, if you open up a terminal in insert mode while the answer is being generated, the output flow into the terminal. So a proper fix for this would entail a more systematic approach rather than just a flag with the current setup I believe.

PR #1056, opened 2025-01-08
vim.highlight deprecated in Nvim 2.0.0, use vim.hl instead

zigotica:
WHen running :checkhealth a warning raises with this info:
vim.highlight deprecated in Nvim 2.0.0, use vim.hl instead


lua/avante/diff.lua
```diff
@@ -80,7 +80,7 @@ local CURRENT_HL = "AvanteConflictCurrent"
 local INCOMING_HL = "AvanteConflictIncoming"
 local CURRENT_LABEL_HL = "AvanteConflictCurrentLabel"
 local INCOMING_LABEL_HL = "AvanteConflictIncomingLabel"
-local PRIORITY = vim.highlight.priorities.user
+local PRIORITY = vim.hl.priorities.user
```
yetone, line 83: ‌‌‌Need to ensure forward compatibility.

yetone:
Thank you for your PR, but this will cause all existing versions of nvim users to be unable to work, so it has been closed for now.

PR #1072, opened 2025-01-11
fix(auto_suggestion_provider): default to claude in copilot (closes #1048)

### LINKED ISSUE #1048: bug: GITHUB suspended my Copilot due to rapid token regeneration
### Describe the bug

Last week my Copilot account was suspended because of rapid token refreshes. Github informed me that this is also affecting other Avante users and has recommended I stop using the plugin. Can someone please shed some light on this?

### To reproduce

_No response_

### Expected behavior

_No response_

### Installation method

Using Packer:
```lua
use({ 'github/copilot.vim' })
use({ 'yetone/avante.nvim' })
```

### Environment

NVIM v0.10.2
Build type: Release
LuaJIT 2.1.1727870382
MacOS Venture 13.4.1
Iterm 2 with Tmux

aarnphm:

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -263,7 +263,7 @@ M.providers = {}
 function M.setup(opts)
   vim.validate({ opts = { opts, "table", true } })
 
-  M._options = vim.tbl_deep_extend(
+  local merged = vim.tbl_deep_extend(
     "force",
     M._defaults,
     opts or {},
@@ -274,6 +274,14 @@ function M.setup(opts)
       },
     }
   )
+
+  -- Check if provider is copilot and warn user
+  if merged.auto_suggestions_provider == "copilot" then
+    Utils.warn("Warning: Copilot is not recommended as the default auto suggestion provider. Switching to Claude.")
+    merged.auto_suggestions_provider = "claude"
+  end
+
+  M._options = merged
   M.providers = vim
     .iter(M._defaults)
     :filter(function(_, value) return type(value) == "table" and value.endpoint ~= nil end)


jonaustin:
@aarnphm Can this be reverted, so that it only forces this if the copilot model is not set to claude?

My organization has Claude 3.5 Sonnet enabled for copilot, and I have my Avante config setup to use claude for copilot, so I don't see the issue?

aarnphm:
It still runs through copilot api, so i need to fix the denounce logic first

jonaustin:
> denounce

Sorry, what is the denounce logic?

jonaustin:
I poked at this a bit more and see the reason now in the linked issue.

Might be good to change the quoted readme line though ^ as it suggests using copilot, but after playing with it, it probably makes more sense to use copilot suggestions by itself (i.e. not avante for autosuggests) if one is using copilot; i.e. if copilot provider is set to claude, I believe it'll do the same thing anyway via copilot directly (I use this fwiw: https://github.com/zbirenbaum/copilot-cmp ).

fwiw; here's a branch that allows the enduser to override the debounce setting: https://github.com/yetone/avante.nvim/compare/main...jonaustin:avante.nvim:debounce-config?expand=1
(it also reverts this PR's change to disallow copilot for autosuggest, but that can be reverted easily enough).

aarnphm:
debounce

NeckBeardPrince:
Okay, but what happens when someone doesn't use Claude now?..

nguyenhoaibao:
Please revert this change. I also use Claude under Copilot provider, but this change keeps prompting that it will switch to use Claude as default auto suggestion provider.
@yetone @aarnphm

skt041959:
Please revert this change. Some enterprise copilot users don't have Claude feature enabled.
In my opinion, it's a good behavior to suggest user, but not to force change the user's setting. 
@yetone @aarnphm

menash134:
please revert this,Please revert this change. Some enterprise copilot users don't have Claude feature enabled.
In my opinion, it's a good behavior to suggest user, but not to force change the user's setting.
@yetone @aarnphm

NeckBeardPrince:
> please revert this,Please revert this change. Some enterprise copilot users don't have Claude feature enabled. In my opinion, it's a good behavior to suggest user, but not to force change the user's setting. @yetone @aarnphm

Doesn't seem like they're paying attention to the issue they have caused. Your best bet is to either fork it or see if you can set the version to a version before this PR.

aarnphm:
Hi there, if you want to fix it, then feel free to open a PR. Currently I don't have a bandwidth atm.

We are both very busy with our full time job, whereas we are currently maintaining this out of our own free time. So be mindful, thanks.

yetone:
‌‌‌‌‌‌Hello, I have made a fix in this PR for "only alert reminders, not mandatory replacement of the suggestion provider": https://github.com/yetone/avante.nvim/pull/1121

enrique-perez-parloa:
Hey here, could anyone point to example configuration of successfully using avante along with copilot?
Is it the case you can only use copilot for autocomplete suggestion and you need to use e.g. Claude for everything else? or can you use copilot for everything with a normal copilot subscription?

jonaustin:
> Hey here, could anyone point to example configuration of successfully using avante along with copilot?

I'm still trialing it, but this seems to work pretty well: https://github.com/jonaustin/dotfiles/blob/master/.config/nvim/lua/plugins/ai.lua#L105-L138

PR #1076, opened 2025-01-13
fix: typo in suggesting.avanterules

theoxo:
Change:
- Typo in suggesting.avanterules: "DON NOT" -> "DO NOT".

Was reviewing the latest changes to avante and noticed this small typo introduced in #1071, doubt it would confuse the model much anyway but figured I might as well throw in a quick fix.


theoxo:
BTW, while you fixed this for the original line I had had in mind in #1077, you actually introduced the same typo into another line in the same commit. Updated the diff to reflect this.

PR #1147, opened 2025-01-30
refactor: use xml format to make suggestions

yetone:


walker-tx:
howdy - anything that you need to get this over the line? I've been encountering some decoding errors while using Gemini for suggest, and I think this PR would fix them.

Issue #1149, opened 2025-01-30
feature: add first-class ollama support


svenstaro:
### Feature request

ollama currently doesn't work very well when users just follow the officially documented way: https://github.com/yetone/avante.nvim/wiki/Custom-providers#ollama

It is currently treated as an OpenAI-compatible provider but as various bug reports show, this isn't really the case and ollama won't work as expected with avante: https://github.com/yetone/avante.nvim/issues/998 https://github.com/yetone/avante.nvim/issues/1067 https://github.com/yetone/avante.nvim/issues/782 https://github.com/yetone/avante.nvim/issues/910

The truth of the matter seems to be that the OpenAI-compatible endpoint isn't too useful for avante's use case, primarily because [it doesn't support setting the context size](https://github.com/ollama/ollama/blob/main/docs/openai.md#setting-the-context-size).

To make things work somehow smoothly, you currently have to do [this](https://github.com/yetone/avante.nvim/issues/1067#issuecomment-2585550870) which is far from a turn-key solution.

So my feature suggestion is: Add a dedicated ollama provider to make ollama _just work_ without going through hoops and adding 50 lines of config. It should use the [ollama chat API](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-chat-completion).

### Motivation

Using ollama with avante should be smooth, easy, and obvious. It currently isn't.

jtabke:
agreed. currently working on a pull request

jtabke:
got denied. but anyways maybe can get the wiki updated with proper instruction on implementing ollama api instead of showing as openai compatible which is causing most of the issues.

```lua
-- Ollama API Documentation https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-completion
local role_map = {
	user = "user",
	assistant = "assistant",
	system = "system",
	tool = "tool",
}

---@param opts AvantePromptOptions
local parse_messages = function(self, opts)
	local messages = {}
	local has_images = opts.image_paths and #opts.image_paths > 0
	-- Ensure opts.messages is always a table
	local msg_list = opts.messages or {}
	-- Convert Avante messages to Ollama format
	for _, msg in ipairs(msg_list) do
		local role = role_map[msg.role] or "assistant"
		local content = msg.content or "" -- Default content to empty string
		-- Handle multimodal content if images are present
		-- *Experimental* not tested
		if has_images and role == "user" then
			local message_content = {
				role = role,
				content = content,
				images = {},
			}
[...]

PR #1150, opened 2025-01-31
feat(providers): add ollama as supported provider

jtabke:
*This implementation is only working with `stream = true`*
- Uses the actual ollama api and allows for passing additional options
- Properly passes the system prompt to api
- Only requires model and endpoint definition in users config

Use ollama as provider in opts like this:
```lua
opts = {
        provider = "ollama",
        ollama = {
                api_key_name = "", -- Optional 
                endpoint = "http://127.0.0.1:11434", -- Required 
                model = "qwen2.5-coder:latest", -- Required
                options = {
                        num_ctx = 32768, -- Optional
                        temperature = 0, -- Optional see https://github.com/ollama/ollama/blob/main/docs/api.md for all options
                },
                stream = true, -- Optional 
        },
```

Or as simply as:
```lua
opts = {
        provider = "ollama",
        ollama = {
                endpoint = "http://127.0.0.1:11434", -- Required 
                model = "qwen2.5-coder:latest", -- Required
        },
```

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -8,7 +8,7 @@ local M = {}
 ---@class avante.Config
 M._defaults = {
   debug = false,
-  ---@alias Provider "claude" | "openai" | "azure" | "gemini" | "vertex" | "cohere" | "copilot" | string
+  ---@alias Provider "claude" | "openai" | "azure" | "gemini" | "vertex" | "cohere" | "copilot" | "ollama" | string
   provider = "claude", -- Only recommend using Claude
   -- WARNING: Since auto-suggestions are a high-frequency operation and therefore expensive,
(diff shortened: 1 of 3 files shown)


aarnphm:
reasoning: https://github.com/yetone/avante.nvim/wiki/Custom-providers

svenstaro:
@aarnphm I disagree in this case, please see https://github.com/yetone/avante.nvim/issues/1149. If you look at [how much code it takes to make this work properly](https://github.com/yetone/avante.nvim/issues/1149#issuecomment-2629226723) then I think users shouldn't have to copy paste (and therefore maintain the whole bunch).

I think given Ollama's popularity, it shouldn't be a "Custom provider" but a first-class provider instead.

PR #1169, opened 2025-02-03
Allow single or multi selection with mini.pick.

theherk:
Since #1153, mini.pick has not worked as a file selector. This both fixes that and enable multi-selection. Thanks to @echasnovski in [this suggestion](https://github.com/echasnovski/mini.nvim/discussions/1547#discussioncomment-12041625).


theherk:
Actually, this is still an improvement. I just noticed it because of the issues, but the improvement is still warranted. It allows multi-select, where the default picker implementation doesn't.

yetone:
‌‌‌‌Okay, I'll test it locally. Thank you for the improvement.

PR #1180, opened 2025-02-05
feat: tools

yetone:


tinkrtailor:
This is interesting. How does one use the web_search_engine

PR #1207, opened 2025-02-07
feat: bedrock temporary credentials support

dtabuenc:
Adds optional suport for including AWS session tokens, so that we can use temporary AWS credentials to access bedrock models.

### DIFF
--- a/README.md
+++ b/README.md
@@ -471,8 +471,10 @@ Given its early stage, `avante.nvim` currently supports the following basic func
 > For Amazon Bedrock:
 >
 > ```sh
-> export BEDROCK_KEYS=aws_access_key_id,aws_secret_access_key,aws_region
+> export BEDROCK_KEYS=aws_access_key_id,aws_secret_access_key,aws_region[,aws_session_token]
+>
 > ```
+> Note: The aws_session_token is optional and only needed when using temporary AWS credentials
 
 1. Open a code file in Neovim.
 2. Use the `:AvanteAsk` command to query the AI about the code.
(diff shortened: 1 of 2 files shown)


tomleavy:
@dtabuenc for some reason I'm running into an issue where I get 403 http code back from bedrock when I use the temporary credential. Have you successfully gotten this to work? Maybe it is a formatting issue I have with the session key?

dtabuenc:
> @dtabuenc for some reason I'm running into an issue where I get 403 http code back from bedrock when I use the temporary credential. Have you successfully gotten this to work? Maybe it is a formatting issue I have with the session key?

Yes.... this is the script I use to generate my env variable:

```sh
#!/bin/sh

# Get AWS credentials and parse them using jq
CREDS=$(aws configure export-credentials)

# Extract the required fields
AWS_ACCESS_KEY=$(echo "$CREDS" | jq -r '.AccessKeyId')
AWS_SECRET_KEY=$(echo "$CREDS" | jq -r '.SecretAccessKey')
AWS_SESSION_TOKEN=$(echo "$CREDS" | jq -r '.SessionToken')
AWS_REGION=$(aws configure get region)

if [ -z "$AWS_ACCESS_KEY" ] || [ -z "$AWS_SECRET_KEY" ] || [ -z "$AWS_REGION" ]; then
  echo "Error: Could not retrieve AWS credentials or region"
  return 1
fi

# Export the BEDROCK_KEYS environment variable with session token if it exists
if [ -n "$AWS_SESSION_TOKEN" ] && [ "$AWS_SESSION_TOKEN" != "null" ]; then
  export BEDROCK_KEYS="$AWS_ACCESS_KEY,$AWS_SECRET_KEY,$AWS_REGION,$AWS_SESSION_TOKEN"
else
  export BEDROCK_KEYS="$AWS_ACCESS_KEY,$AWS_SECRET_KEY,$AWS_REGION"
fi

echo "BEDROCK_KEYS has been set successfully"

```

tomleavy:
Ok, it turns out actually my curl version was the issue. The next thing though is should we make an issue to automatically refresh the keys somehow? After a couple of hours my temp creds are expired and I have to rerun a script to make new ones.

PR #1287, opened 2025-02-16
fix: correct typo in answer field name in tavily format_response_body…

kyrisu:
There is a typo in `format_response_body` that prevents wavily from working.


yetone:
LGTM

PR #1295, opened 2025-02-18
feat(llm_tools): add git diff and commit functions

yuchanns:
Added two new functions to support git operations:

- git_diff: Get git diff for generating commit message (both staged and unstaged changes)

- git_commit: Commit changes with commit message and optional signing

These functions also support:

- GPG signing if available and configured

- Automatic staging of files if scope is provided

- User confirmation before committing

- Signed-off-by line with git user info


yuchanns:
The CI Rule is damn freaking 😅

yetone:
‌‌‌‌‌‌‌I think this should be within the scope of avante llm tools. Perhaps we need to provide an llm tools spec next, allowing users to customize llm tools.

yetone (approved):

PR #1299, opened 2025-02-18
fix: do not have bg

yetone:


academo:
thanks for the quick fix!

PR #1300, opened 2025-02-18
allow disabling tools for claude & copilot

Pereira-Lucas:
While tinkering with my config, I saw that adding `disable_tools = true` did nothing to my claude config.

So I grabbed the code from openai and added it to claude & copilot

### DIFF
--- a/lua/avante/providers/claude.lua
+++ b/lua/avante/providers/claude.lua
@@ -227,6 +227,7 @@ end
 ---@return table
 M.parse_curl_args = function(provider, prompt_opts)
   local provider_conf, request_body = P.parse_config(provider)
+  local disable_tools = provider_conf.disable_tools or false
 
   local headers = {
     ["Content-Type"] = "application/json",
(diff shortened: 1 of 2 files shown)


yetone:
‌‌‌‌Under what circumstances is it necessary to disable the tools of these two providers? Previously, the function to disable tools was added for the OpenAI provider because many models compatible with the OpenAI API do not support tools.

Pereira-Lucas:
I had some problems with my Windows config, so I tried to modify my config. When I looked at the readme, I noticed it mentions we can disable tools for Claude, but when I tried this feature, it didn't work.

https://github.com/yetone/avante.nvim/blob/main/README.md?plain=1#L634

My problems were not related to the tools and I did make it work at the end, so I understand it might not be useful to merge, but maybe remove that line from the readme ?

yetone:
> I had some problems with my Windows config, so I tried to modify my config. When I looked at the readme, I noticed it mentions we can disable tools for Claude, but when I tried this feature, it didn't work.
> 
> [`main`/README.md?plain=1#L634](https://github.com/yetone/avante.nvim/blob/main/README.md?plain=1&rgh-link-date=2025-02-19T14%3A09%3A11Z#L634)
> 
> My problems were not related to the tools and I did make it work at the end, so I understand it might not be useful to merge, but maybe remove that line from the readme ?

I understand, it might be that some LLM tools are only targeted at the Linux operating system and not Windows. I will try to fix this issue.

yetone:
This PR can be merged, thank you!

PR #1323, opened 2025-02-20
Improve responsiveness while streaming generations.

hanspinckaers:
This remove stopinsert from lock_buf. 
This makes my neovim usable while we're receiving results. Otherwise it would lag and be generally unusable.


yetone:
‌‌‌This PR will cause behavior inconsistent with expectations, so it has been reverted for now. A better solution to this performance issue will be found in the future.

PR #1345, opened 2025-02-22
fix: load avante_lib synchronously when not using lazy

ojaskavathe:
Avante doesn't load its libs when using home-manager on nix (or any other plugin manager really), since it can't find lazy.


yetone:
‌Perhaps you need to update the README, and also ensure prevention of duplicate loading.

ojaskavathe:
`load()` seems to already prevent duplicate loading.

https://github.com/yetone/avante.nvim/blob/27d3f294a4cdb372ad9248ff4eb03dbf50bc3855/lua/avante_lib.lua#L13-L18

PR #1367, opened 2025-02-23
chores: add python ci badge

yetone:


otakutyrant:
What is the point of making PRs instead of just committing immediately? No offense. I am just studying.

PR #1369, opened 2025-02-23
chore: cleanup and share tool func type

PeterCardenas:
Shares the tool function type into the builtin tools and dynamic tools. This makes the types slightly less accurate for some of the tools (allowing `string|nil|boolean` when the type should be `boolean` only), but removes a decent amount of boilerplate. Additionally, this allows for adding new parameters, and for my particular use case, I would like to allow certain tools to run asynchronously, specifically the `run_command` tool.

In future PRs, I would also like to make the `on_log` parameter more powerful, such as showing a floating terminal window to see terminal output. This might be too tailored to this tool, but would be a much better user experience.

### DIFF
--- a/lua/avante/llm_tools.lua
+++ b/lua/avante/llm_tools.lua
@@ -34,9 +34,7 @@ local function has_permission_to_access(abs_path)
 end
 
 ---@param opts { rel_path: string, max_depth?: integer }
----@param on_log? fun(log: string): nil
----@return string files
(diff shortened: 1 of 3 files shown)


PeterCardenas:
@yetone fyi I had to add `| nil` to the `on_log` parameter even though it was marked as an optional type. looks like the language server wasn't smart enough to infer it

PR #1377, opened 2025-02-24
feat: run_command is run asynchronously

PeterCardenas:
For commands that take more than a couple seconds, the `run_command`
tool is very disruptive since it blocks the user interface. This commit
allows for the tool to be run asynchronously, and also supports other
tools to be run asynchronously as well.


lua/avante/types.lua
```diff
@@ -323,7 +323,11 @@ vim.g.avante_login = vim.g.avante_login
 ---@field on_stop AvanteLLMStopCallback
 ---@field on_tool_log? function(tool_name: string, log: string): nil
 ---
----@alias AvanteLLMToolFunc<T> fun(input: T, on_log?: (fun(log: string): nil) | nil): (boolean | string | nil, string | nil)
+---@alias AvanteLLMToolFunc<T> fun(
+---  input: T,
+---  on_log?: (fun(log: string): nil) | nil,
+---  on_complete?: (fun(result: boolean | string | nil, error: string | nil): nil) | nil)
+---  : (boolean | string | nil, string | nil)
```
PeterCardenas, line 330: rest assured the typing is still safe here, github's syntax highlighting is just not as good as treesitter + semantic token highlighting from lua_ls 😆

lua/avante/utils/init.lua
```diff
@@ -89,17 +87,47 @@ M.shell_run = function(input_cmd, shell_cmd)
...
+  local output = {}
+  fn.jobstart(cmd, {
+    on_stdout = function(_, data)
+      if not data then return end
+      vim.list_extend(output, data)
+    end,
+    on_stderr = function(_, data)
+      if not data then return end
+      vim.list_extend(output, data)
+    end,
```
PeterCardenas, line 126: i made it like this so that we can stream output in the future, which i would like to do for the `run_command` tool

PeterCardenas:
I made it so that we can legacy synchronous path in case that is intended some time in the future. I know that asynchronous code in lua can be slower than synchronous in some cases, so it might be intended to run some too synchronously

yetone:
‌‌‌Great PR! You might need to resolve the code conflict.

PeterCardenas:
i sensed that my other pr would cause merge conflicts, but did not realize it would be me own 🤣  just fixed

PR #1379, opened 2025-02-25
fix: python tool message formatting and check for docker

PeterCardenas:
I encountered a couple of different issues with the python tool:
1. There was a typo in the confirm message and it is a bit unclear.
2. When docker is not installed, the tool fails and the response does not finish.

I have fixed the first issue by fixing the typo, listing out the container name, and making sure that the code starts on a new line.
The second issue was fixed by first checking if docker is installed, and if not, sending back a message to the LLM to make it clear that the Python tool should not be used. I had a bit of prompt engineering here since a simple error message here made the LLM try the python tool again with simpler python scripts.

### DIFF
--- a/lua/avante/llm_tools.lua
+++ b/lua/avante/llm_tools.lua
@@ -507,17 +507,19 @@ function M.python(opts, on_log)
   if not Path:new(abs_path):exists() then return nil, "Path not found: " .. abs_path end
   if on_log then on_log("cwd: " .. abs_path) end
   if on_log then on_log("code: " .. opts.code) end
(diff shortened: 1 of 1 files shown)


PeterCardenas:
additionally, the confirm message seems quite jank after putting the newline, maybe we should consider some different UI instead of `vim.fn.confirm`

yetone:
> additionally, the confirm message seems quite jank after putting the newline, maybe we should consider some different UI instead of `vim.fn.confirm`

I am using a float window to replace vim.fn.confirm in another branch.

PR #1381, opened 2025-02-25
chore: prefer not to use function assignment

PeterCardenas:
the `x = function` way of declaring functions makes it so that there are two definitions when doing a "go to definition" action. This new syntax fixes that issue in addition to making the function definition more succinct.


PeterCardenas:
i did a cursory look and didn't find any linters that check for this. sorry for any merge conflicts this causes, feel free to close if this isn't warranted

PR #1391, opened 2025-02-25
feat: add Windows support for git directory scanning

yuchanns:
feat: add Windows support for git directory scanning

Add PowerShell command for scanning git directories on Windows systems while
maintaining the existing bash command for Unix-like systems. This allows the
directory scanner to work properly across different platforms.

Hanchin Hsieh <me@yuchanns.xyz>


yetone:
cool!

PR #1413, opened 2025-02-27
docs: correct type to class in readme opts example

t-dub:


README.md
```diff
@@ -382,7 +382,7 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
     },
   },
   highlights = {
-    ---@type AvanteConflictHighlights
+    ---@class AvanteConflictHighlights
```
yetone, line 385: ‌‌‌‌‌‌`@class` is for defining a type, `@type` is for referencing a type, so using `@type` here is correct because it references this type: https://github.com/yetone/avante.nvim/blob/9c9fadd256d6138d771e17b9ca68905908e16c17/lua/avante/diff.lua#L28

Issue #1407, opened 2025-02-26
feature: Ollama RAG support


nzlov:
### Feature request

please add ollama rag support。

PR #1427, opened 2025-02-28
feat: add support for ollama RAG providers 

nzlov:
- Added provider, model and endpoint configuration options for RAG service
- Updated RAG service to support Ollama providers
- Added Ollama embedding support and dependencies
- Improved environment variable handling for RAG service configuration

#1407


lua/avante/rag_service.lua
```diff
@@ -63,13 +64,14 @@ function M.launch_rag_service(cb)
     Utils.debug(string.format("container %s not found, starting...", container_name))
   end
   local cmd_ = string.format(
-    "docker run -d -p %d:8000 --name %s -v %s:/data -v /:/host -e DATA_DIR=/data -e OPENAI_API_KEY=%s -e OPENAI_API_BASE=%s -e OPENAI_EMBED_MODEL=%s %s",
+    "docker run -d -p %d:8000 --name %s -v %s:/data -v /:/host -e DATA_DIR=/data -e RAG_PROVIDER=%s -e API_KEY=%s -e API_BASE=%s -e RAG_EMBED_MODEL=%s %s",
```
yetone, line 67: ‌‌‌It is recommended not to use such vague environment variable names: `API_KEY` `API_BASE`

nzlov: ok

nzlov:
LLM is still needed for parsing, so this should also be configurable.

yetone:
‌‌‌‌LGTM, you might need to fix the py lint error:

```bash
pre-commit run --color=always --files $(find ./py -type f -name "*.py")
```

yetone:
‌‌‌‌Thank you for the contribution! We should synchronize the documents and inform everyone on how to customize the RAG provider. Adding a few examples would be even better.

Additionally, will switching providers abruptly render the previous provider's embedding invalid? We should address this issue.

nzlov:
Tell users in the documentation that they need to delete previous data when changing providers?

yetone:
> Tell users in the documentation that they need to delete previous data when changing providers?

Can we do this automatically?

nzlov:
Due to `pre-commit` update [version](https://github.com/microsoft/pyright/releases/tag/1.1.396), need to wait for the third-party package to update and fix the error

yetone:
‌‌‌‌Can you use `pyright: ignore` to temporarily ignore the error on this line?

nzlov:
OK, thanks for your help

PR #1428, opened 2025-02-28
refactor: rename run_command llm tool to bash

yetone:


PeterCardenas:
@yetone any particular reason for this change? how would you propose i support particular shells?

yetone:
‌For LLMs, a flexible shell is not an advantage because tools like fish/zsh are designed for human readability and ease of writing shell scripts. However, this is not an issue for LLMs as they are more familiar with bash. If a tool is named bash, the LLM will know to write a suitable bash command. But if a tool is named run_command, it becomes hesitant due to its flexibility and doesn't know which shell command to write. This mental burden can actually reduce the quality of how the LLM uses tools. @PeterCardenas

yetone:
‌‌Unless we have a command called `run_in_terminal`, the $SHELL information is crucial for llm.

PeterCardenas:
I'm ok with the LLM being worse at generating arbitrary commands, I'm more concerned about the LLM not being able to execute my functions without the output being verbose or the LLM being confused about a command not being available. How do you propose I solve this? One thing I was thinking of was to allow disabling and adding custom tools so that I can take ownership of my potentially niche use case, but if you think prefixing commands can be more applicable (such as executing commands in a docker container which is something that cursor supports) I can also try doing something like run_in_terminal. Just let me know what's a better fit for this project :)

PeterCardenas:
also i've experienced the tool do `cd` instead of passing the `rel_path` argument, which possibly means that `bash` is making it call the command incorrectly? maybe we just need to prompt engineer the `run_command` more instead of making this change

yetone:
> I'm more concerned about the LLM not being able to execute my functions without the output being verbose or the LLM being confused about a command not being available.

Would the previous `run_command` solve this problem? Why would this solve the problem? This is very confusing to me.

Sorry, maybe I didn't express myself clearly. My point is that LLM tends to execute more explicit instructions. You can think of it like this: the `bash` tool here corresponds to the `python` tool, just as we shouldn't have a `run_code` to replace the `python` tool, we also shouldn't have a `run_command` to replace the `bash` tool.

lua/avante/llm_tools.lua
```diff
@@ -264,13 +264,13 @@ function M.run_command(opts, on_log, on_complete)
     return output, nil
   end
   if on_complete then
-    Utils.shell_run_async(opts.command, Config.run_command.shell_cmd, function(output, exit_code)
+    Utils.shell_run_async(opts.command, "bash -c", function(output, exit_code)
```
ianks, line 267: Running `bash -c` means the command is invoked non-interactively, which means `.bashrc` et al are not sourced. This becomes toilsome quickly since commands that Just Work in a regular shell do not work for the LLM.

PR #1433, opened 2025-02-28
fix: rag server stop deadlock

nzlov:
Fix the problem that nvim is stuck when the container stops

### DIFF
--- a/lua/avante/rag_service.lua
+++ b/lua/avante/rag_service.lua
@@ -41,7 +41,7 @@ function M.launch_rag_service(cb)
   local port = M.get_rag_service_port()
   local image = M.get_rag_service_image()
   local data_path = M.get_data_path()
-  local cmd = string.format("docker ps -a | grep '%s'", container_name)
+  local cmd = string.format("docker ps | grep '%s'", container_name)
   local result = vim.fn.system(cmd)
   if result ~= "" then
     Utils.debug(string.format("container %s already running", container_name))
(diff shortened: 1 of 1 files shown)


lua/avante/rag_service.lua
```diff
@@ -62,6 +62,12 @@ function M.launch_rag_service(cb)
   else
     Utils.debug(string.format("container %s not found, starting...", container_name))
   end
+
+  -- check stop container
+  cmd = string.format("docker ps -a | grep '%s'", container_name)
+  result = vim.fn.system(cmd)
+  if result ~= "" then M.stop_rag_service() end
```
yetone, line 69: ‌‌‌Could you please help explain the difference between doing it this way and before?

nzlov: Previously, we fetched all Docker containers (including stopped ones). Since there was no check for running containers, subsequent connections would cause Neovim to freeze. Now we first check for running containers. If none exist, we then check for stopped containers. If stopped containers are found, they are removed before launching a new container.

yetone: ‌‌‌‌‌‌‌But `docker ps -a` returns all containers, so why would it miss running containers?

nzlov: oh, no. did not check to stopped containers.

nzlov: Since there is no check to stop the docker container, the process proceeds normally, causing nvim to freeze.

nzlov: It should be that there is no check whether the container is stopped. Because `docker ps -a` returns all containers, including those that have been stopped. However, the subsequent logic does not determine whether the container is stopped, resulting in a loop waiting for the service to start.

nzlov:
#1448 `-rm` is good

lua/avante/rag_service.lua
ftassi: docker ps allows for filtering. Something like `docker ps -a --filter "Status=running"`  should just return running container

PR #1446, opened 2025-03-01
feat(repo-map): add swift, java and php support

gaelph:
Adds repo map support for the following languages:
- Swift
- Java
- PHP


mikedfunk:
I don't know what the repo map feature is as it doesn't show up in the help doc, but I get a message that it is empty or not supported in a PHP project FYI.

gaelph:
@mikedfunk , you are right. It can help the model pickup the files that are the most useful for the task at hand, without using a RAG (which tend to miss information due to the chuncking of files). A repo map is basically like an outline view of every symbol exported by every files.

PR #1448, opened 2025-03-01
Improve avante-rag-service container execution

ftassi:
Mounting the whole filesystem may expose the user to security risks. It is something I was worried about when I noticed it.

This PR mounts only the user home directory in the container, to mitigate the security risks. The user home directory is mounted in read only mode to even reduce the risks of accidental or malicious modifications.

Mounting the whole $HOME should allow the user to have multiple neovim instances runinng at the same time and sharing the same rag_service.

Also the container is started with the --rm flag to remove it after it stops.

If we like the idea of improving how the container is started, we could go even further, but this should be a nice improvement as is.

What do you think?


nzlov:
Just mounting in the home directory is not enough, if the code is not in the home directory it will fail.
Perhaps should specify the directory to mount in.

ftassi:
@nzlov I am aware of it. Just thought that using neovim with avante for something outside the home dir would be rather unusual. 
Shall we add a config parameter inside the rag_service dic to let the user define what to mount? would $HOME be a decent default?

ftassi:
@nzlov the mounted path is now configurable. Would this be ok?

ftassi:
@yetone  the mounted path is now configurable. How do you feel about this?

PeterCardenas:
I think it makes more sense to use `vim.fn.getcwd()` as the mount path. configuration is nice, but probably is overkill

lua/avante/utils/filter.lua
```diff
@@ -0,0 +1,137 @@
...
+    if file:match(pattern) then return true end
+  end
+  return false
+end
+
+-- Checks if a file should be filtered from RAG service
+---@param file string File path
+---@param project_root string Project root path
+---@return boolean true if file should be filtered
+function M.should_filter_file(file, project_root)
```
yetone, line 97: ‌‌‌‌‌‌Has this function never been used?

ftassi: Sorry, my bad: the whole file was committed accidentally

ftassi:
> I think it makes more sense to use `vim.fn.getcwd()` as the mount path. configuration is nice, but probably is overkill


That was my starting point, but the issue here is that we have 1 rag-service container for multiple (potencially) instances of  neovim running. If you just mount the current project and start another neovim in another project, or even close the first one, start a second one, without killing the container first, you will get a container with the wrong mount, hence no rag-service.

Mounting the whole `/` solved this because the container would be able to access everything. Mounting `/` to a container I don't fully understand or know is too much  for me as a user. Hence the idea of restricting the mount point

mounting whole `$HOME` is a tradeoff between usability (I want to just start neovim and have rag-service working) and security.

Being a configurable parameter allows the user to cover every edge case he might want to cover.

The *best* solution would be to have a dedicated container for each neovim instance maybe, so we could mount just `vim.fn.getcwd()` but this would require allocating a dynamic port instead of 20250 for the rag service. I would like to work on that if we like such idea, but I think ti will require some work and some testing, considering it should work on different OS. 

In the meantime this allows the user to _restrict_ the scope of the mount point for the container, improving the current situation

yetone:
‌‌In fact, there is no security issue with mounting the entire root because my purpose of using Docker is to better handle Python dependency issues. If I don't use Docker and directly start a Python process on the host machine, the files that can be seen are actually the same, right?

ftassi:
@yetone, there is a security issue when mounting the entire filesystem because the container runs as root. This effectively gives the container root access to your whole filesystem, whereas running Python on your system typically executes as a user process.

The concern is not only what Python can do with the mounted filesystem but also what any other software running in the container can do. For example, if you run:

`docker run -it --rm -v /:/host bash:4.4`

once inside the container, you should be able to run cat /host/etc/shadow.

If malware were to infiltrate the image, it could gain access to the user’s entire filesystem, as root. Additionally, a bug in the Avante Python server could lead to serious consequences.

Mounting as read-only reduces these risks. Restricting the mount point is even better, and updating the Dockerfile to create and use an application-specific user would be ideal (although this might introduce some permission issues). We should follow the principle of least privilege, don't you think?

Also, as a Docker user, I must say that seeing a container mount / as a volume looks suspicious to me. My first reaction is to inspect the Dockerfile before starting the container. Even mounting the entire home directory is risky, as it gives the container access to my ~/.ssh directory, something I am not willing to do unless I fully trust the container (for example, distrobox requires you to mount the entire $HOME).

yetone:
> Also, as a Docker user, I must say that seeing a container mount / as a volume looks suspicious to me.

This sentence successfully convinced me.

lua/avante/rag_service.lua
```diff
@@ -111,16 +113,19 @@ function M.to_container_uri(uri)
...
 end
 
 function M.to_local_uri(uri)
   local scheme = M.get_scheme(uri)
   if scheme == "file" then
-    local path = uri:match("^file://host(.*)$")
-    uri = string.format("file://%s", path)
+    local path = uri:match("^file:///host(.*)$")
+    local host_dir = Config.rag_service.host_mount
+    uri = string.format("file://%s", host_dir .. path)
```
yetone, line 128: ‌‌‌Would using `Path:new():joinpath` be a bit safer?

ftassi:
@yetone fetched latest changes from main. Conflict solution should be ok.
I have used `Path:new():joinpath` as you suggested and even added some tests for the conversion functions.

xeon826:
Is it expected behavior that the rag container should run at all times and occasionally put the CPU usage up to around 100%? I'm also getting charged by openai about 5 usd every couple hours. I feel like this isn't intended. I had this set as my rag_service config previously
```
			rag_service = {
				enabled = true, 
			},
```
But I noticed it was consuming a lot of CPU and RAM so I changed it to 
```
			rag_service = {
				enabled = true, -- Enables the RAG service
				host_mount = os.getenv("HOME"), -- Host mount path for the rag service
				provider = "openai", -- The provider to use for RAG service (e.g. openai or ollama)
				llm_model = "", -- The LLM model to use for RAG service
				embed_model = "", -- The embedding model to use for RAG service
				endpoint = "https://api.openai.com/v1", -- The API endpoint for RAG service
			},
```
I've only been using this config few a minutes and I haven't noticed it spike the CPU on the same scale as the previous config, both cases took place when I was on a commit after this PR was merged, can someone please tell me have I set this up correctly? Furthermore do I add an llm_model/embed_model or are they given default values if I just set them to ""?

PR #1453, opened 2025-03-01
Add o3-mini provider for copilot

StevenMatchett:
I am very new to this codebase, but when I was debugging the `parse_curl_args` and see that it is in fact passing `o3-mini` to the api, but that seemed to easy. If I did something wrong if you could pass me in the right direction

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -244,6 +244,14 @@ M._defaults = {
       temperature = 0,
       max_tokens = 8000,
     },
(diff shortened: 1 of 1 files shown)


omarcresp:
Nothing wrong but it is a configuration option, no need to add a built in custom provider for that
you should be abble to just point to o3-mini in your avante setup options like:
```lua
  {
    provider = "copilot",
    copilot = {
      model = "o3-mini"
    },
  }
```

Issue #1447, opened 2025-03-01
feature: disable specific tools


eyalk11:
### Feature request

Allow to disable specific available tools. 

### Motivation

I don't want it to ask or try to use docker

PR #1459, opened 2025-03-02
feat: disable tools with config

PeterCardenas:
Since the claude endpoint really likes to call the python tool, I've preferred to just completely disable it since it's not proved very useful for me. I've added a config to disable tools by name, which will allow me to disable the python tool. I also imagine a workflow where you add overrides to disable particular tools to be focused on particular tasks that are known, but good for LLMs to handle, e.g. refactoring across many files only really needs the rag tool, maybe some of the file search tools and the write_file tool.

closes: #1447

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -408,6 +408,7 @@ M._defaults = {
     debounce = 600,
     throttle = 600,
   },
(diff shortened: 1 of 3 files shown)


eyalk11:
https://github.com/yetone/avante.nvim/issues/1447

PR #1460, opened 2025-03-02
fix: claude bedrock content_block initiation

IsomerX:
The bedrock requests were failing saying: 
Error executing vim.schedule lua callback: ...re/nvim/lazy/avante.nvim/lua/avante/providers/claude.lua:182: attempt to index local 'content_block' (a nil value)


yetone:
‌‌‌‌‌Is there no `content_block_start` event in bedrock?

IsomerX:
there is, but its not triggering for some reason. I was very very frustrated. This is just a fix I feel not a solution. Would love your input.

yetone:
fixed in this PR: https://github.com/yetone/avante.nvim/pull/1475

PR #1461, opened 2025-03-02
feat: add custom tools config

PeterCardenas:
With [this PR](https://github.com/yetone/avante.nvim/pull/1428), I've found that the `bash` command does not fit my use case. I need to be able to run some fish functions, and also docker exec before running a command. I have a decent amount of what I feel are reasonable concerns in my comments in the PR, but to save time for the maintainer, would also be open to maintaining my own custom tool, as I have enabled in this PR. A couple of changes to the config and LLM types here are so that the type checking is helpful when calling the `setup` function.

Please do let me know if you prefer another way to support my use case.

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -17,6 +17,7 @@ local Utils = require("avante.utils")
 ---@class avante.CoreConfig: avante.Config
 local M = {}
 ---@class avante.Config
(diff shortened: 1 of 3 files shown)


PeterCardenas:
fixed conflicts with my own pr again :P

yetone:
‌‌‌‌‌LGTM, we should add a document.

PR #1465, opened 2025-03-02
Ignore git crypt files

ftassi:
I am working on something to ensure no sensitive files are sent to remote servers. Ignoring files from `.gitignore` is ok but I have some projects were  I use [git-crypt](https://github.com/AGWA/git-crypt/tree/master) to encrypt local files. 
Such files contains secrets, are plain text on my working directory and are not git ignored by design.

This PR update the python server to also retrieve them from git and ignore them.

Some more testing required maybe, but at least I would like  a feedback about the idea. 
_Not sure about the logging_
_This should also require a bump in the image version number._


yetone:
‌‌‌‌‌‌Hello, please fix the lint error.

ftassi:
@yetone I'm encountering issues with the linter options here. I want to call git as a subprocess to query for git-crypt files, but I'm getting an error whether I use Shell=true or Shell=false. I haven't found a better solution than silencing S603, although I'm not entirely convinced it's the right choice.

Alternatively, I could avoid using a subprocess by directly parsing the .gitattributes file, but I'm not sure I like that approach. I prefer to use git itself to retrieve the encrypted files rather than manually parsing a file (what if the syntax changes?).

I've also implemented a small check to ensure this function works even in a monorepo context. If a user is working within a monorepo, they're likely to launch Neovim from a subdirectory rather than the git root, and the function now accounts for that scenario.

PR #1480, opened 2025-03-03
feat: add Nix as Rag service runner option

PJalv:
Wanted to create a way to have the Rag service not use docker, so I created a nix shell, and made sure it can mingle well with the existing files (e.g. Dockerfile). 

Uses uv for python packages. 

More than able to answer any questions.


yetone:
Hello, please fix the lint error.

PR #1494, opened 2025-03-05
feat: add key mappings to close sidebar in input_container

brookhong:
The PR makes below changes

* add key mappings to close sidebar in input_container
* add autocmd to enter normal mode on leaving input_container
* add autocmd to enter insert mode on entering input_container

so that user can close sidebar in input_container directly without switching to result_container.

And the auto commands help to avoid below annoyances in result_container.
image

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -148,6 +148,12 @@ function Sidebar:close(opts)
   vim.cmd("wincmd =")
 end
 
+function Sidebar:shutdown()
+  Llm.cancel_inflight_request()
+  self:close()
+  vim.cmd("stopinsert")
+end
+
 ---@return boolean
 function Sidebar:focus()
   if self:is_open() then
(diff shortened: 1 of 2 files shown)


yetone:
‌‌‌Some people use `<C-c>` as `<esc>` to exit insert mode

brookhong:
Any preference for the `close` key in insert mode?  Or how about to make `Config.mappings.sidebar.close` a dictionary like `Config.mappings.submit`, which will be a breaking change?

brookhong:
Added `Config.mappings.sidebar.close_from_input`, with it's default value as

      {
        normal = "<Esc>",
        insert = "<C-d>",
      }

`<C-d>` to close side bar in insert mode, user still can use `<C-c>` to exit insert mode, and I can use below config

        opts = {
            mappings = {
                sidebar = {
                    close_from_input = {
                      normal = "<Esc>",
                      insert = "<C-c>",
                    },
                },
            },
        },

lua/avante/sidebar.lua
```diff
@@ -2810,6 +2818,18 @@ function Sidebar:create_input_container(opts)
     callback = function() close_hint() end,
   })
 
+  api.nvim_create_autocmd("BufEnter", {
+    group = self.augroup,
+    buffer = self.input_container.bufnr,
+    callback = function() vim.cmd("startinsert") end,
```
yetone, line 2824: ‌‌‌‌‌‌Need to respect this setting: https://github.com/yetone/avante.nvim/blob/2b0e7e09aedacb19955485cdf6a30138c09bcbd1/lua/avante/config.lua#L407

brookhong: Resolved

PR #1526, opened 2025-03-08
fix: Always handle the extra response outside of stream, such as the exceptions from bedrock

brookhong:
Just found bedrock exception handling is broken, we need to check stream flag from server response(`spec.body.stream`) also, rather than just depend on the provider config.

### DIFF
--- a/lua/avante/llm.lua
+++ b/lua/avante/llm.lua
@@ -246,7 +246,9 @@ function M.curl(opts)
   end
 
   local function parse_response_without_stream(data)
-    provider:parse_response_without_stream(data, current_event_state, handler_opts)
+    if provider.parse_response_without_stream then
+      provider:parse_response_without_stream(data, current_event_state, handler_opts)
+    end
   end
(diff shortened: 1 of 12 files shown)


yetone:
The stream parameter for Gemini Prover is always nil.

brookhong:
Then it seems that `is_disable_stream` is totally not needed, as stream has been set here https://github.com/yetone/avante.nvim/blob/main/lua/avante/llm.lua#L277.

PS: spec.body.stream is always nil for bedrock in either normal cases or exceptions. Removing `is_disable_stream` works totally well for bedrock.

brookhong:
@yetone Do you have any chance to verify whether this `is_disable_stream` is necessary?


if ~~provider:is_disable_stream() and~~ result.status == 200 then

yetone:
‌‌‌‌‌‌I feel it can be deleted.

brookhong:
Deleted it.

PR #1536, opened 2025-03-09
fix: remove `sidebar.close_from_input` default keybinding

jaywonchung:
`<C-d>` should not be an insert mode default binding.
- `<C-d>` acts as the Delete key (remove a single letter after ther cursor) in shells including Bash and Zsh.
- After all, `<C-d>` (scroll half page) is such a commonly used key combination, and silently binding this key to something else leads to unexpected behavior. For instance, when jumping back and forth between code and Avante, users are prone to press `<C-d>` to scroll either code or the Avante generated response. When the cursor happens to jump to the input text window, `<C-d>` will close Avante and the input text users were currently drafting will be lost.

I've been bit by this like three times today. I think I won't be alone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,8 +349,14 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
     sidebar = {
       apply_all = "A",
       apply_cursor = "a",
+      retry_user_request = "r",
+      edit_user_request = "e",
       switch_windows = "<Tab>",
       reverse_switch_windows = "<S-Tab>",
+      remove_file = "d",
+      add_file = "@",
+      close = { "<Esc>", "q" },
+      close_from_input = nil, -- e.g., { normal = "<Esc>", insert = "<C-d>" }
     },
   },
   hints = { enabled = true },
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -383,10 +383,7 @@ M._defaults = {
       remove_file = "d",
       add_file = "@",
       close = { "<Esc>", "q" },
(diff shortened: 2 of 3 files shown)


yetone:
These two lines also need to be modified in response: https://github.com/yetone/avante.nvim/blob/0f11e2cbf9fd11ef873a16bacb0021b1c41d6f9c/lua/avante/sidebar.lua#L2713

yetone:
cc @brookhong

jaywonchung:
Alternatives:
1. Remove `normal` and `insert` variants, making `close_from_input`'s type consistent with all other keybinding actions.
2. Default `insert` to `nil` and only call `map` if `insert ~= nil`.
3. Allow both type `string` and `table` for `close_from_input` and perform type check before calling `map`(s).

I honestly feel like this part of the code is getting too unstructured and dirty (especially so if we go with 2 or 3). I can personally live fine with just pressing `<Esc>` twice. So my preference is 1, but I don't feel very strongly about this. I'm amenable to the maintainers' decision.

brookhong:
How about make `close_from_input` default nil, and make changes below?

    if Config.mappings.sidebar.close_from_input ~= nil then
        self.input_container:map("n", Config.mappings.sidebar.close_from_input.normal, function() self:shutdown() end)
        self.input_container:map("i", Config.mappings.sidebar.close_from_input.insert, function() self:shutdown() end)
    end

jaywonchung:
Done. @yetone @brookhong

brookhong:
Looks good to me, thanks

PR #1544, opened 2025-03-09
Feat - prevent file scanning hang on large codebases & fix Repo map is empty

SouhailBlmn:
This commit was able to fix the pesky Repo map is empty once and for all. it was also tested on multiple prod codeabses in different languages, and it was able to explain the codebase for me fine

In this commit, the following was done : 

- Add max_depth limit (100) to prevent deep directory recursion
- Use fast external tools (rg/fd) for file scanning with Lua fallback (way faster than Lua and proved better in large code files that are in the thousands)
- Add timeout protection for external commands
- Improve directory scanning to handle large projects efficiently

Previously, scanning large codebases would hang when trying to access all files recursively. Now, the system efficiently scans directories using ripgrep/fd with depth limits, making it work reliably even on large projects with deep directory structures.

### DIFF
--- a/lua/avante/utils/init.lua
+++ b/lua/avante/utils/init.lua
@@ -731,9 +731,7 @@ function M.scan_directory(options)
       if type == "file" then
         table.insert(files, path)
       elseif type == "directory" and name ~= ".git" and name ~= "node_modules" then
-        if options.add_dirs then
-          table.insert(files, path)
-        end
+        if options.add_dirs then table.insert(files, path) end
         if not depth or depth > 1 then
           local subfiles = scan_directory_lua(path, depth and depth - 1)
           vim.list_extend(files, subfiles)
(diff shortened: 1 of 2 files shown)


lua/avante/utils/init.lua
```diff
@@ -716,97 +716,90 @@ end
 ---@param options { directory: string, add_dirs?: boolean, max_depth?: integer }
 ---@return string[]
 function M.scan_directory(options)
-  local cmd_supports_max_depth = true
-  local cmd = (function()
+  -- Fallback function using Lua's native file operations
+  local function scan_directory_lua(dir, depth)
+    if depth and depth < 1 then return {} end
+    local files = {}
+    local handle = vim.loop.fs_scandir(dir)
```
yetone, line 723: ‌‌‌‌‌This doesn't respect the gitignore, does it?

yetone: ‌‌‌‌‌‌In addition, this method is very slow, so I just refactored this function a while ago: https://github.com/yetone/avante.nvim/pull/1362/files#diff-7ae8be0a3aeefac22b7db21bfa364ae531c260181497299a95326035c20d47b5L698

PR #1547, opened 2025-03-10
fix: skip more binary files than just audio/video in RAG scan

Tom94:
This could potentially be relaxed in the future if PDFs (and similar) documents could be converted to text form via something like pandoc in the future. But until then, I think it's good to be liberal in what gets skipped. My understanding is that the RAG service currently can deal only with text.

### DIFF
--- a/py/rag-service/src/main.py
+++ b/py/rag-service/src/main.py
@@ -660,21 +660,89 @@ def scan_directory(directory: Path) -> list[str]:
     """Scan directory and return a list of matched files."""
     spec = get_pathspec(directory)
 
-    audio_video_exts = [
+    binary_extensions = [
+        # Images
+        ".png",
+        ".jpg",
+        ".jpeg",
+        ".gif",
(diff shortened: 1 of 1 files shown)


yetone:
‌‌I am wondering if we can think in reverse and only support the extensions of some programming languages that we support.

yetone:
like this: https://gist.github.com/ppisarczyk/43962d06686722d26d176fad46879d41

Tom94:
It's a fair point -- the main argument against whitelisting specific programming languages that I can see is that some projects use either esoteric languages, new ones (e.g. I was using the Slang shading language at work before it got publicly released) or use non-standard file extensions.

Another less strong argument: the number of binary file extensions I can think of is much smaller then the list of programming languages you linked. I knew there were a lot, but holy moly, I didn't expect such a long list 😅

yetone:
‌‌‌There may be some truth to it. 🤔

PR #1559, opened 2025-03-11
feat: Add claude tools to vertex claude provider

miguelosana:
Adds the ability to use tools when using the vertex_claude provider

### DIFF
--- a/lua/avante/providers/vertex_claude.lua
+++ b/lua/avante/providers/vertex_claude.lua
@@ -1,5 +1,28 @@
 local P = require("avante.providers")
 local Vertex = require("avante.providers.vertex")
+---
+---@param tool AvanteLLMTool
+---@return AvanteClaudeTool
+local function transform_tool(tool)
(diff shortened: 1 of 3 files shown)


lua/avante/providers/vertex_claude.lua
```diff
@@ -1,5 +1,28 @@
 local P = require("avante.providers")
 local Vertex = require("avante.providers.vertex")
+---
+---@param tool AvanteLLMTool
+---@return AvanteClaudeTool
+local function transform_tool(tool)
```
yetone, line 6: Can this function be placed in claude.lua and then reused?

miguelosana: It is the same function as the one in claud.lua, but because it is defined as local I was not able to use it. Is it ok to remove the local keyword? in claude.lua?

yetone: ‌It's okay to export this method.

PR #1569, opened 2025-03-12
Revert "fix: Always handle the extra response outside of stream, such…

yetone:
… as the exceptions from bedrock (#1526)"

This reverts commit f9ab6934d23b6b5226bd3cd7254984470b8144a1.

image

@brookhong

### DIFF
--- a/lua/avante/llm.lua
+++ b/lua/avante/llm.lua
@@ -246,9 +246,7 @@ function M.curl(opts)
   end
 
   local function parse_response_without_stream(data)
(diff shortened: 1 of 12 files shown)


brookhong:
Sorry about that, I don't have openai access, but I think the proper action is to fix `parse_response_without_stream` of openai, rather than revert.

yetone:
‌‌‌I believe the problem lies in our use of `parse_response_without_stream` to handle data returned by `stream`, which is clearly a misuse of this method, so I reverted it.

PR #1589, opened 2025-03-15
added optional docker args for rag service

erahhal:
I'm using a corporate LLM proxy that runs as it's own docker container, and exposes a port localhost. the hostname is fixed and points to 127.0.0.1, which does not work from inside the avante rag docker container, so the host needs to be mapped in, e.g.:

`--add-host corporate.proxy.lan:docker-proxy-name`

There may be other custom docker configs that aren't accounted for as well, such as additional environment variables, e.g.

`-e ANONYMIZED_TELEMTRY=False`

This new option makes this future-proof.


retarpt:
I need this change - thank you!

Issue #1577, opened 2025-03-13
bug: Can't control height in horizontal layout, and width of input window


toupeira:
### Describe the bug

I'd like to use the horizontal layout with `position = 'bottom'` or `position = 'smart'`, but can't seem to control the height. It always opens at 50-60%:

![Image](https://github.com/user-attachments/assets/959a06d9-147e-41aa-9a7d-94f5248611fa)

I noticed that it does work correctly when no file is opened:

![Image](https://github.com/user-attachments/assets/98f83b00-b2df-4dc8-a816-60a441bdd144)

The `width` setting also works fine for the vertical layout.

I'd also like to make the input window wider, I see there's a setting `windows.input.height` but no `.width`

### To reproduce

1. `nvim -u repro.lua` (see bottom of the issue)
2. Open a file
3. Open Avante with `\aa`

### Expected behavior

1. Bug: `windows.height` should always work correctly
2. Feature request: `windows.input.width` option

### Installation method

Using lazy.nvim and latest commit 931a364

### Environment

`nvim -v`:

```
NVIM v0.10.4
Build type: Release
LuaJIT 2.1.1713484068
Run "nvim -V1 -v" for more info
```

`uname -a`:

```
Linux ocelot 6.12.17-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.17-1 (2025-03-01) x86_64 GNU/Linux
```

Distribution: Debian sid

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    {
      'yetone/avante.nvim',
      event = 'VeryLazy',
      build = 'make',

      dependencies = {
        'nvim-treesitter/nvim-treesitter',
        'stevearc/dressing.nvim',
        'nvim-lua/plenary.nvim',
        'MunifTanjim/nui.nvim',
      },

      opts = {
        windows = {
          position = 'bottom',
          height = 20,
        },
      },
    }
  },
})
```

toupeira:
Updated the issue with a proper `repro.lua`, I noticed it does work correctly when no file is opened so I wanted to check if another plugin is interfering, but I can reproduce it with the minimal config too.

I tried debugging the code but didn't find the problem yet. I traced all the calls to `nvim_win_set_height` like so:

```lua
orig = vim.api.nvim_win_set_height
[...]

PR #1590, opened 2025-03-15
Fixes to adjusting window sizes

toupeira:
Fixes https://github.com/yetone/avante.nvim/issues/1577

First commit drops the `wincmd =` calls and seems straightforward to me, I'm not sure what the reasoning was to add them previously in https://github.com/yetone/avante.nvim/pull/160 and https://github.com/yetone/avante.nvim/pull/198.

Second commit drops the `nvim_win_set_width` call in `adjust_result_container_layout()`, at first I was unsure about this but after testing a bit more and trying different approaches I think it makes sense too: As I understand it the plugin wants to adjust the size of the results container in response to changes to the selected code and files, and this only seems necessary in vertical layout where we want to adjust the height. In horizontal layout the width seems like it should be static so we don't need to mess with it.

See repro steps in https://github.com/yetone/avante.nvim/issues/1577, with these changes the plugin properly respect the `windows.width` setting for me, and the input window width stays at the hard-coded 40%.

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -130,8 +130,6 @@ function Sidebar:open(opts)
     vim.g.avante_login = true
   end
 
-  vim.cmd("wincmd =")
-
   return self
 end
 
@@ -198,8 +196,6 @@ function Sidebar:close(opts)
   end
 
   self:recover_code_winhl()
-
-  vim.cmd("wincmd =")
 end
 
 function Sidebar:shutdown()
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -3168,10 +3168,8 @@ function Sidebar:get_result_container_width()
 end
 
 function Sidebar:adjust_result_container_layout()
-  local width = self:get_result_container_width()
   local height = self:get_result_container_height()
 
-  api.nvim_win_set_width(self.result_container.winid, width)
   api.nvim_win_set_height(self.result_container.winid, height)
 end
 


toupeira:
@aarnphm please review and test, I'm still pretty new to Avante so might have missed some breakage :grinning:

toupeira:
Rebased and resolved conflicts.

@aarnphm or @yetone, could you please review? :grinning:

toupeira:
> This PR is stale because it has been open 14 days with no activity. Remove stale label or comment or this will be closed in 10 days.

bumping...

lua/avante/sidebar.lua
```diff
@@ -3172,10 +3168,8 @@ function Sidebar:get_result_container_width()
 end
 
 function Sidebar:adjust_result_container_layout()
-  local width = self:get_result_container_width()
   local height = self:get_result_container_height()
 
-  api.nvim_win_set_width(self.result_container.winid, width)
```
yetone, line 3178: ‌‌‌Why is this?

toupeira: @yetone thanks for merging! See commit message:

> When using horizontal layout this stretches the result container too far
> and displaces the selected files and input containers, which were
> previously set up with a hard-coded width of 40%.
> 
> It looks like the result container width only needs to be adjusted in
> vertical layout (changing the height based on selected code/files),
> and we can stop adjusting the width without any noticeable problems.

huix9527:
nvim stops adjusting window sizes when opening avante.

PR #1598, opened 2025-03-16
feat: support tools in bedrock

brookhong:
Tested with bedrock, haven't tested with claude which should have been tested, but I don't have access to claud api.

Please make sure it won't break claude.

image

### DIFF
--- a/lua/avante/providers/bedrock/claude.lua
+++ b/lua/avante/providers/bedrock/claude.lua
@@ -6,6 +6,7 @@
 ---@field role "user" | "assistant"
 ---@field content [AvanteBedrockClaudeTextMessage][]
 
+local P = require("avante.providers")
 local Claude = require("avante.providers.claude")
 
 ---@class AvanteBedrockModelHandler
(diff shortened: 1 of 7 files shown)


lua/avante/providers/bedrock/claude.lua
```diff
@@ -20,19 +20,30 @@ M.role_map = {
...
 M.parse_messages = Claude.parse_messages
 M.parse_response = Claude.parse_response
+M.transform_tool = Claude.transform_tool
 
 ---@param provider AvanteProviderFunctor
 ---@param prompt_opts AvantePromptOptions
 ---@param request_body table
 ---@return table
-function M.build_bedrock_payload(provider, prompt_opts, request_body)
+function M.build_bedrock_payload(provider, provider_conf, prompt_opts, request_body)
```
yetone, line 29: ‌‌‌provider_conf can be obtained from the provider

brookhong: Thanks, use `P.parse_config(provider)` instead.

PR #1604, opened 2025-03-16
feat: update openai/azure params

kernitus:
Realising that a lot of this could be done through user configuration, but I feel it's better for it to be the default.

Update OpenAI and Azure parameters following new docs:

- https://learn.microsoft.com/en-us/azure/ai-services/openai/api-version-deprecation
- https://platform.openai.com/docs/api-reference/chat/create

* Update deprecated `max_tokens` to `max_completion_tokens`
* Add `reasoning_effort` for reasoning models, ignored otherwise
* Use `developer` message instead of `system` for reasoning models
* Detect whether model is reasoning (o-series) and use correct set of parameters
* Reduce duplication between OpenAI and Azure providers
* Tested with both Azure and OpenAI with both reasoning and non-reasoning models

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ For building binary if you wish to build from source, then `cargo` is required. 
       timeout = 30000, -- Timeout in milliseconds, increase this for reasoning models
       temperature = 0,
       max_completion_tokens = 8192, -- Increase this to include reasoning tokens (for reasoning models)
-      --reasoning_effort = 'medium', -- low|medium|high, only used for reasoning models
+      --reasoning_effort = "medium", -- low|medium|high, only used for reasoning models
     },
   },
   -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
--- a/README.md
+++ b/README.md
@@ -65,10 +65,10 @@ For building binary if you wish to build from source, then `cargo` is required. 
     openai = {
       endpoint = "https://api.openai.com/v1",
       model = "gpt-4o", -- your desired model (or use gpt-4o, etc.)
-      timeout = 30000, -- timeout in milliseconds
-      temperature = 0, -- adjust if needed
-      max_tokens = 4096,
-      -- reasoning_effort = "high" -- only supported for reasoning models (o1, etc.)
+      timeout = 30000, -- Timeout in milliseconds, increase this for reasoning models
+      temperature = 0,
+      max_completion_tokens = 8192, -- Increase this to include reasoning tokens (for reasoning models)
+      --reasoning_effort = 'medium', -- low|medium|high, only used for reasoning models
     },
   },
   -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
--- a/cursor-planning-mode.md
+++ b/cursor-planning-mode.md
@@ -35,7 +35,7 @@ Then enable it in avante.nvim:
             api_key_name = 'GROQ_API_KEY',
             endpoint = 'https://api.groq.com/openai/v1/',
             model = 'llama-3.3-70b-versatile',
-            max_tokens = 32768, -- remember to increase this value, otherwise it will stop generating halfway
+            max_completion_tokens = 32768, -- remember to increase this value, otherwise it will stop generating halfway
         },
     },
     --- ... existing configurations
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -217,7 +217,7 @@ M._defaults = {
     temperature = 0,
     max_completion_tokens = 8192, -- Increase this to include reasoning tokens (for reasoning models)
     reasoning_effort = "medium", -- low|medium|high, only used for reasoning models
-    max_tokens = 8192,
+    max_completion_tokens = 8192,
   },
   ---@type AvanteSupportedProvider
   claude = {
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -193,9 +193,10 @@ M._defaults = {
   openai = {
     endpoint = "https://api.openai.com/v1",
     model = "gpt-4o",
-    timeout = 30000, -- Timeout in milliseconds
+    timeout = 30000, -- Timeout in milliseconds, increase this for reasoning models
     temperature = 0,
-    max_tokens = 4096,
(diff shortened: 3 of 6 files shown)


yetone:
‌‌You need to resolve the code conflicts

kernitus:
> ‌‌You need to resolve the code conflicts

I see. I synced it back with main now

README.md
```diff
@@ -65,10 +65,10 @@ For building binary if you wish to build from source, then `cargo` is required.
     openai = {
       endpoint = "https://api.openai.com/v1",
       model = "gpt-4o", -- your desired model (or use gpt-4o, etc.)
-      timeout = 30000, -- timeout in milliseconds
-      temperature = 0, -- adjust if needed
-      max_tokens = 4096,
```
yetone, line 70: ‌‌‌‌‌Will the deletion of this parameter affect other OpenAI compatible APIs?

kyboi: OpenAI compatible APIs should be using the OpenAI API - which has deprecated `max_tokens` a while back in favour of `max_completion_tokens`. For example, I am currently using the Groq API with the OpenAI interface:

```lua
            vendors = {
                groq = {
                    __inherited_from = "openai",
                    api_key_name = "GROQ_API_KEY",
                    endpoint = 'https://api.groq.com/openai/v1/',
                    model = 'llama-3.3-70b-versatile',
                    max_completion_tokens = 32768,
                }
            },
```

Who specify [on their docs](https://console.groq.com/docs/api-reference#chat-create) `max_tokens` is deprecated in favour of `max_completion_tokens`.

I do believe if someone needed it, they could still configure max_tokens to a value and max_completion_tokens to nil in their config.

kernitus: OpenAI compatible APIs should be using the OpenAI API - which has deprecated `max_tokens` a while back in favour of `max_completion_tokens`. For example, I am currently using the Groq API with the OpenAI interface:

```lua
            vendors = {
                groq = {
                    __inherited_from = "openai",
                    api_key_name = "GROQ_API_KEY",
                    endpoint = 'https://api.groq.com/openai/v1/',
                    model = 'llama-3.3-70b-versatile',
                    max_completion_tokens = 32768,
                }
            },
```

Who specify [on their docs](https://console.groq.com/docs/api-reference#chat-create) `max_tokens` is deprecated in favour of `max_completion_tokens`.

I do believe if someone needed it, they could still configure max_tokens to a value and max_completion_tokens to nil in their config.

jcsmithf22:
This breaks using mistral as mistral throws an error when receiving max_completions_token, even if max_tokens is set and max_completions_token is set to nil.

kernitus:
> This breaks using mistral as mistral throws an error when receiving max_completions_token, even if max_tokens is set and max_completions_token is set to nil.

Can you provide the config you have for mistral?

jcsmithf22:
> > This breaks using mistral as mistral throws an error when receiving max_completions_token, even if max_tokens is set and max_completions_token is set to nil.
> 
> Can you provide the config you have for mistral?

```
mistral = {
  __inherited_from = "openai",
  api_key_name = "MISTRAL_API_KEY",
  endpoint = "https://api.mistral.ai/v1/",
  model = "mistral-large-latest",
  -- max_tokens = 20480,
  -- max_completion_tokens = nil,
},
```

Any configuration of uncommenting max_tokens and max_completion_tokens results in an api error because the mistral api doesn't accept max_completion_tokens.

kernitus:
> > > This breaks using mistral as mistral throws an error when receiving max_completions_token, even if max_tokens is set and max_completions_token is set to nil.
> > 
> > 
> > Can you provide the config you have for mistral?
> 
> ```
> mistral = {
>   __inherited_from = "openai",
>   api_key_name = "MISTRAL_API_KEY",
>   endpoint = "https://api.mistral.ai/v1/",
>   model = "mistral-large-latest",
>   -- max_tokens = 20480,
>   -- max_completion_tokens = nil,
> },
> ```
> 
> Any configuration of uncommenting max_tokens and max_completion_tokens results in an api error because the mistral api doesn't accept max_completion_tokens.

It seems inheriting from openai causes Lua to overwrite the max_completion_tokens field with the default. We could either try to prevent this logic, or add a check in the openai class to not set max_completion_tokens if max_tokens is set. I can open a PR for that now

jcsmithf22:
> It seems inheriting from openai causes Lua to overwrite the max_completion_tokens field with the default. We could either try to prevent this logic, or add a check in the openai class to not set max_completion_tokens if max_tokens is set. I can open a PR for that now

Thank you, I appreciate it.

kernitus:
> > It seems inheriting from openai causes Lua to overwrite the max_completion_tokens field with the default. We could either try to prevent this logic, or add a check in the openai class to not set max_completion_tokens if max_tokens is set. I can open a PR for that now
> 
> Thank you, I appreciate it.

I don't have mistral to be able to test but it seems to work with Azure and OpenAI configs #1633

jcsmithf22:
> > > It seems inheriting from openai causes Lua to overwrite the max_completion_tokens field with the default. We could either try to prevent this logic, or add a check in the openai class to not set max_completion_tokens if max_tokens is set. I can open a PR for that now
> 
> > 
> 
> > Thank you, I appreciate it.
> 
> 
> 
> I don't have mistral to be able to test but it seems to work with Azure and OpenAI configs #1633

Ok it works now.

yetone:
image

Feel the need to revert this PR.

kernitus:
> image image
> 
> Feel the need to revert this PR.

Isn't that just telling you the default `max_completion_tokens` value you set as the default is higher than that provider allows?

yetone:
My goal is to ensure that there is no friction when any user uses the OpenAI compatible API, for example, they only need to configure it like this:

image

yetone:
> > My goal is to ensure that there is no friction when any user uses the OpenAI compatible API, for example, they only need to configure it like this:
> > image
> 
> That's understandable, but surely the error you showed was telling you you exceeded the max tokens limit rather than there being a problem with the parameter name?

Sorry, I misunderstood and have reverted it back. https://github.com/yetone/avante.nvim/pull/1839

PR #1612, opened 2025-03-17
fix: parse empty code block in markdown

yetone:


lua/avante/sidebar.lua
```diff
@@ -575,7 +575,8 @@ local function tree_sitter_markdown_parse_code_blocks(source)
     [[ (fenced_code_block
       (info_string
         (language) @language)?
-      (code_fence_content) @code) ]]
+      (block_continuation) @code_start
+      (fenced_code_block_delimiter) @code_end) ]]
```
haarisr, line 579: I believe this change should also be reflected in `extract_cursor_planning_code_snippets_map`. Might be related to #1618

PR #1617, opened 2025-03-17
fix: header text in case of no nvim-web-devicons setup and show curre…

brookhong:
This PR is to fix header text in case of no nvim-web-devicons setup and show current provider/model in it.
Before
image

After
image


lua/avante/sidebar.lua
```diff
@@ -1439,7 +1441,11 @@ function Sidebar:render_result()
   then
     return
   end
-  local header_text = Utils.icon("󰭻 ") .. "Avante"
+  local header_text = Utils.icon("󰭻 ")
+    .. "Avante - "
+    .. Config.provider
+    .. "/"
+    .. Config.get_provider_config(Config.provider).model
```
yetone, line 1448: ‌‌‌‌‌‌I feel it's not visually appealing to put model names in the header, as sometimes model names can be very long, and some providers even use random strings as model names (like Volcano Engine)

brookhong: But for ollama users,  to display `model` makes more sense than `provider`. I changed it to width-wise, so that `model` is appended only when there is enough room for it.

image

lua/avante/sidebar.lua
```diff
@@ -1438,7 +1440,10 @@ function Sidebar:render_result()
   then
     return
   end
-  local header_text = Utils.icon("󰭻 ") .. "Avante"
+  local header_text = Utils.icon("󰭻 ") .. "Avante - " .. Config.provider
```
yetone, line 1443: Let's first remove the provider and model from the header, and later I'll think of a better place to display this information.

PR #1669, opened 2025-03-21
fix: sidebar border

yetone:


TikiCat7:
I'm not sure if it's intended but currently getting white borders around avante, previously it was not happening.

PR #1682, opened 2025-03-22
fix: make sure to get the final highlight values rather than link names

mehalter:
If any of the avante highlight groups are configured as links to different highlight groups, then the internals of retrieving highlights and detecting background fallback stuff breaks. This makes sure to get the final highlight values of `fg/bg` rather than just `link = "x"`

### DIFF
--- a/lua/avante/highlights.lua
+++ b/lua/avante/highlights.lua
@@ -90,12 +90,12 @@ function M.setup_conflict_highlights()
 
   ---@return number | nil
   local function get_bg(hl_name)
-    local hl = api.nvim_get_hl(0, { name = hl_name })
+    local hl = api.nvim_get_hl(0, { name = hl_name, link = false })
     return hl.bg
   end
 
   local function get_bold(hl_name)
-    local hl = api.nvim_get_hl(0, { name = hl_name })
+    local hl = api.nvim_get_hl(0, { name = hl_name, link = false })
     return hl.bold
   end
 
--- a/lua/avante/highlights.lua
+++ b/lua/avante/highlights.lua
@@ -53,7 +53,7 @@ local H = {}
 local M = {}
 
 local function has_set_colors(hl_group)
-  local hl = api.nvim_get_hl(0, { name = hl_group })
+  local hl = api.nvim_get_hl(0, { name = hl_group, link = false })
   return next(hl) ~= nil
 end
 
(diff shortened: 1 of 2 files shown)


lua/avante/highlights.lua
```diff
@@ -53,7 +53,7 @@ local H = {}
 local M = {}
 
 local function has_set_colors(hl_group)
-  local hl = api.nvim_get_hl(0, { name = hl_group })
+  local hl = api.nvim_get_hl(0, { name = hl_group, link = false })
```
jackblackevo, line 56: I applied this modification locally for testing, but the warnings for `AvanteConflictCurrent` and `AvanteConflictIncoming` still appeared.

However, when I applied `link = false` to the [`vim.api.nvim_get_hl`](https://neovim.io/doc/user/api.html#nvim_get_hl()) in the `get_bg` function, the warning notifications no longer appeared:

https://github.com/yetone/avante.nvim/blob/4749e4ea1aff653a9e6f6e542d390e3c7560d161/lua/avante/highlights.lua#L91-L95

```diff
  ---@return number | nil
  local function get_bg(hl_name)
-   local hl = api.nvim_get_hl(0, { name = hl_name })
+   local hl = api.nvim_get_hl(0, { name = hl_name, link = false })
    return hl.bg
  end
```

jackblackevo, line 56: I applied this modification locally for testing, but the warnings for `AvanteConflictCurrent` and `AvanteConflictIncoming` still appeared.

However, when I applied `link = false` to the [`vim.api.nvim_get_hl`](https://neovim.io/doc/user/api.html#nvim_get_hl()) in the `get_bg` function, the warning notifications no longer appeared:

```diff
local function get_bg(hl_name)
- local hl = api.nvim_get_hl(0, { name = hl_name })
+ local hl = api.nvim_get_hl(0, { name = hl_name, link = false })
  return hl.bg
end
```

https://github.com/yetone/avante.nvim/blob/4749e4ea1aff653a9e6f6e542d390e3c7560d161/lua/avante/highlights.lua#L91-L95

mehalter: Ah I just have missed that. It should be all uses of that API

jackblackevo:
I'm not sure if these also need to add `link = false`:

https://github.com/yetone/avante.nvim/blob/4749e4ea1aff653a9e6f6e542d390e3c7560d161/lua/avante/highlights.lua#L72-L75

https://github.com/yetone/avante.nvim/blob/4749e4ea1aff653a9e6f6e542d390e3c7560d161/lua/avante/utils/init.lua#L310-L315

PR #1693, opened 2025-03-24
fix: redundant square bracket

nguyenhoaibao:


lua/avante/llm_tools/dispatch_agent.lua
```diff
@@ -19,7 +19,7 @@ Usage notes:
 1. Launch multiple agents concurrently whenever possible, to maximize performance
 2. When the agent is done, it will return a single message back to you
 3. Each agent invocation is stateless
-4. The agent's outputs should generally be trusted]]
+4. The agent's outputs should generally be trusted
```
yetone, line 22: ‌‌‌‌Why do this? `[[]]` is a valid Lua syntax, doing so will only cause a syntax error.

PR #1706, opened 2025-03-25
fix: refine tools

yetone:


lua/avante/llm_tools/init.lua
```diff
@@ -1008,33 +990,6 @@ M._tools = {
       },
     },
   },
-  {
-    name = "create_file",
```
nguyenhoaibao, line 1012: Any reason why is this tool get removed?

PR #1729, opened 2025-03-26
Draft: Allow for custom OPENAI_API_KEY in rag_service

### LINKED ISSUE #1728: feature: Allow setting api key for RAG service in lua config
### Feature request

I would like to set the API key for OpenAI in the same way I do for openai and other providers:
```lua
    openai = {
      ...
      api_key_name = "cmd:op read op://Employee/OpenAI/credential",
    },
```

### Motivation

I like to keep my secrets in 1password. However, I do not want to be prompted to log into it as soon as I start Neovim/ZSH but only when I actually use avante. This is possible with the default providers but not with the RAG service which I want to try out.

### Other

_No response_

m4r1vs:
closes #1728 

Had do modify the parser of env variables because it did not block the request which meant that the rag_service started before the API key has actually been parsed.


m4r1vs:
Also I noticed that sometimes I start typing before 1pw has loaded the creds. I now set the parser to synchronously load the variables always which fixes it for me. However, I can see that this might not be optimal for everyone.

Should we provide this (very technical) option through lua config or maybe still do it always and maybe add another pre_fetch function which populates the cache?

Uzaaft:
This seems to be already implemented. See https://github.com/yetone/avante.nvim/wiki#secrets

m4r1vs:
> This seems to be already implemented. See https://github.com/yetone/avante.nvim/wiki#secrets
haven't checked yet if there was an update but when I made the PR, it was only possible to use the cmd secrets for the "normal" API keys, not the API key used for the rag service.
However, I just rebased my branch and seems like there's been some work in regards to the RAG API key, will check and close this PR if so

PR #1738, opened 2025-03-27
Fix README.md

msdone-lwt:
Error field "disabled_tools"


yetone:
‌‌‌The document is correct.

https://github.com/yetone/avante.nvim/blob/a06bb97db6e522f51de1427b72762a1ad0794279/lua/avante/config.lua#L478

PR #1752, opened 2025-03-27
Resolve AWS credentials using default credentials provider chain for Bedrock

msvechla:
This switches to resolving AWS credentials using the default AWS
credentials provider chain.

This has the advantage that AWS credentials are resolved in the same way
as other AWS tools do. It also supports more mechanisms now like AWS
profiles, AWS authentication via single-sign on and more.

It is a breaking change for bedrock users, as the old `BEDROCK_KEYS` env
var will no longer work, however it gives much more flexibility in terms
of authentication.

I have used this in my custom provider for quite some time successfully: https://gitlab.com/msvechla/avante-provider-bedrock.nvim

Let me know if you have some thoughts on this or other improvements ideas.

Thanks a lot!

### DIFF
--- a/lua/avante/providers/bedrock.lua
+++ b/lua/avante/providers/bedrock.lua
@@ -1,6 +1,5 @@
 local Utils = require("avante.utils")
 local P = require("avante.providers")
-local Job = require("plenary.job")
 
 ---@class AvanteBedrockProviderFunctor
 local M = {}
@@ -101,7 +100,7 @@ end
 function M:parse_curl_args(prompt_opts)
   local provider_conf, request_body = P.parse_config(self)
 
-  local access_key_id, secret_access_key, session_token, region = "", "", "", ""
+  local access_key_id, secret_access_key, session_token, region
 
   -- try to parse credentials from api key
   local api_key = self.parse_api_key()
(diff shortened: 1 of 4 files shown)


lua/avante/providers/bedrock.lua
```diff
@@ -71,34 +100,33 @@ end
...
 function M:parse_curl_args(prompt_opts)
   local provider_conf, request_body = P.parse_config(self)
+  ---@diagnostic disable-next-line: undefined-field
+  local region = provider_conf.aws_region
+  ---@diagnostic disable-next-line: undefined-field
+  local profile = provider_conf.aws_profile
+
+  local awsCreds = M:get_aws_credentials(region, profile)
 
-  local api_key = self.parse_api_key()
```
yetone, line 75: ‌‌‌Will this cause the old setup method to become invalid? This is a disruptive update for users. Can we be compatible with the old method?

msvechla: fixed, I kept the existing functionality with `BEDROCK_KEYS` as default and falling back to the new aws default credentials provider chain as alternative.

msvechla:
@yetone the `Lint Lua` CI findings appear to be false-positives. Is there anything else I should work on to get this merged? Thanks a lot!

yetone:
‌‌‌It doesn't seem to be a false alarm. I'll find some time to fix these two lint errors.

lua/avante/providers/bedrock.lua
```diff
@@ -1,18 +1,47 @@
 local Utils = require("avante.utils")
 local P = require("avante.providers")
+local Job = require("plenary.job")
```
yetone, line 3: ‌‌‌‌Can we not use plenary? Avante is gradually removing its dependency on plenary.

msvechla: do you have a preferred alternative for running commands and parsing their output?

msvechla: @yetone I removed the dependency to plenary.job and used the `vim.loop` api instead. Let me know what you think!

msvechla:
@yetone looks like all checks are passing now and I removed the dependency as requested. Let me know if there is something else I should check!

uthmanmoh:
Hey, this seems like an awesome enhancement but it doesn't seem to be working for me. I've validated that 
```sh
aws configure export-credentials --profile default --region us-east-1
```
works for me and returns the expected 3 fields as JSON, and I know this account has access to the model because when I use the same values in a BEDROCK_KEYS format like so:
```sh
        AWS_ACCESS_KEY_ID=$(aws configure get aws_access_key_id)
        AWS_SECRET_ACCESS_KEY=$(aws configure get aws_secret_access_key)
        AWS_SESSION_TOKEN=$(aws configure get aws_session_token)
        AWS_REGION="us-east-1"
        export BEDROCK_KEYS="$AWS_ACCESS_KEY_ID,$AWS_SECRET_ACCESS_KEY,$AWS_REGION,$AWS_SESSION_TOKEN"
```
the plugin works as expected. However without setting BEDROCK_KEYS the plugin doesn't work. It still prompts me to enter a value for BEDROCK_KEYS. Is that expected?

msvechla:
I am using this setup daily, so there should be no issues.

Can you paste your avante.nvim config? Also what version of avante are you using and what version of `curl` do you have setup on your machine?

uthmanmoh:
https://github.com/user-attachments/assets/bc6b3131-13d1-4688-8db9-11dbd0e74b4d

Hmm interestingly it's behaving a bit different today. It seems like it actually ends up working after I ignore the input for BEDROCK_KEYS. Yesterday I was getting an error when making a request too.

So seems like it always expects BEDROCK_KEYS to be set?

tixwho:
Hello,

Thanks for the nice addition!

First, I can confirm that ignoring the input for key, instead of blindly hitting enter, would allow nvim to read the config through credential provider normally. A bit counterintuitive but not too bad.

Cross-region inference seems to be a problem though. Probably this is a US thing, but Amazon has been ramping up with its requirements of cross-region inference for their top models, so that their top models (Claude Sonnet 4/3.7, Deepseek R1...) can *only* be called when an inference profile ID and inference profile ARN are provided. ([Link to the description](https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference.html)).

For example, the following error will arise without providing the cross-regeion inference parameters:
```
- Datetime: 2025-06-04 15:42:41

- Model: bedrock/anthropic.claude-3-7-sonnet-20250219-v1:0

- Selected files:
  - src/components/PostRendering.vue

> Who are you?


Error: API request failed with status 400. Body: '{"message":"Invocation of model ID anthropic.claude-3-7-sonnet-20250219-v1:0 with on-demand throughput isn’t supported. Retry your request with the ID or ARN of an inference profile that contains this model."}'
```

Is their a plan to add support of `Inference Profile ID` and `Inference Profile ARN` as two optional parameters in the bedrock config?

msvechla:
Inference profiles should already work, you just have to use the correct model identifier. E.g.

```
			providers = {
				bedrock = {
					model = "us.anthropic.claude-sonnet-4-20250514-v1:0",
					extra_request_body = {
						max_tokens = 32000,
						temperature = 0,
					},
					aws_profile = "bedrock",
					aws_region = "us-east-1",
				},
			},
```

Should use cross-region inference for the US (see the `us.`) at the model arn.
Or are you talking about some other inference?

PR #1764, opened 2025-03-29
feat: accept all

yetone:


zigotica:
so, is this the reason why the answers are automatically applied now @yetone ? if so, is there an option to avoid this invasive behaviour?

Issue #1758, opened 2025-03-28
bug: ChatGPT API failed by setting both 'max_tokens' and 'max_completion_tokens' at the same time


wwingyou:
### Describe the bug

I'v been using avante.nvim pretty well since yesterday, but suddenly It show the same error for all reqeust.

Here's the output

```
- Datetime: 2025-03-28 14:04:24

- Model: openai/gpt-4o

- Selected files:
  - src/main/java/com/goolbitg/api/v1/service/ChallengeGroupServiceImpl.java

> helle??



Error: "API request failed with status 400. Body: \"{\\n  \\\"error\\\": {\\n    \\\"message\\\": \\\"Setting 'max_tokens' and 'max_completion_tokens' at the same time is not supported.\\\",\\n    \\\"type\\\": \\\"invalid_request_error\\\",\\n    \\\"param\\\": \\\"max_tokens\\\",\\n    \\\"code\\\": \\\"invalid_parameter_combination\\\"\\n  }\\n}\""
```

### To reproduce

Error showed suddenly. I don't know why. I tried avante on different project and it worked. But for the project I'm currently working, every request show error. I selected different files, cleared avante, turn off and on neovim, and all of them has no effect.

My project is java/spring project.

### Expected behavior

api respond with no error

### Installation method

return {
  "yetone/avante.nvim",
  event = "VeryLazy",
  version = false, -- Never set this value to "*"! Never!
  opts = {
    -- add any opts here
    -- for example
    provider = "openai",
    openai = {
      endpoint = "https://api.openai.com/v1",
      model = "gpt-4o", -- your desired model (or use gpt-4o, etc.)
      timeout = 30000, -- Timeout in milliseconds, increase this for reasoning models
      temperature = 0,
      max_completion_tokens = 8192, -- Increase this to include reasoning tokens (for reasoning models)
      --reasoning_effort = "medium", -- low|medium|high, only used for reasoning models
      disable_tools = true,
    },
  },
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  build = "make",
  -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
  dependencies = {
    "nvim-treesitter/nvim-treesitter",
    "stevearc/dressing.nvim",
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
    --- The below dependencies are optional,
[...]

PR #1817, opened 2025-04-05
fix: set 'max_completion_tokens' as default parameter for openai (#1758)

wwingyou:


yetone:
fixed in this pr: https://github.com/yetone/avante.nvim/pull/1819

PR #1841, opened 2025-04-09
Add ability to fetch ollama models from provider itself

### LINKED ISSUE #1581: feature: Display Ollama models in `AvanteModels` selection popup
### Feature request

Add Ollama models to `AvanteModels` popup for quick switching without restarting Neovim. As for now there is only one model that which is currently set in the `provider` configuration.

Image

Possible ways of solution:
- Having a list of `n` models in `provider` config.
- Providing all models from Ollama API.
- Providing all models from Ollama API with some filtering using `regex`.

### Motivation

I find myself wanting to try different Ollama models depending on what I'm working on, but it's a pain having to close Neovim, edit my config files (I'm the `nixvim` user, so have double pain 😄 ), and restart everything just to switch models. Would be so much nicer to just pick a different model from the popup that's already there (see screenshot).

Think that this feature might be also helpful and for others.

### Other

_No response_

chrisands:
Should fix #1581

### DIFF
--- a/lua/avante/model_selector.lua
+++ b/lua/avante/model_selector.lua
@@ -16,11 +16,56 @@ local function create_model_entry(provider_name, cfg)
     }
 end
 
-function M.open()
+local spinner_frames = { "⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏" }
+local spinner_index = 1
+local spinner_timer = nil
+local spinner_msg_id = nil
+
+local function start_spinner()
+  spinner_msg_id = vim.notify("Fetching available models... " .. spinner_frames[spinner_index], vim.log.levels.INFO, {
+    title = "Avante",
+    replace = spinner_msg_id,
+  })
+
+  spinner_index = (spinner_index % #spinner_frames) + 1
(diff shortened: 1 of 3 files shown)


chrisands:
This is early draft of functionality. I think it still needs to be done:
- [ ] Spinner probably should be used from package or separate file.
- [ ] Add OpenAI, Claude, Gemini models 
- [ ] Cleanup code
 
@yetone what do you think?

chrisands:
Not stale, waiting for comment from @yetone

repparw:
Could this be generalized? See https://github.com/yetone/avante.nvim/issues/1933#issuecomment-2906955545

yetone:
‌‌Sorry for my late reply. I recently merged a [PR](https://github.com/yetone/avante.nvim/pull/2106) that dynamically retrieves copilot models, and it seems these two PRs are closely related. Could you continue to implement this PR to support OpenAI/Anthropic/Gemini, etc.? If you don't have the time, please let me know so I can take over the subsequent development of this PR.

yetone:
I'm very sorry, I will refer to your design to implement the model fetch with a loading effect.

PR #1844, opened 2025-04-10
fix(lint): make neovim 0.11 typecheck happy

yetone:


lua/avante/ui/confirm.lua
```diff
@@ -359,6 +361,7 @@ function M:close()
     self._popup = nil
     return true
   end
+  if self._ns_id then vim.api.nvim_del_namespace(self._ns_id) end
```
wookayin, line 364: This API doesn't exist!

PR #1860, opened 2025-04-11
feat: add Bedrock Claude 3.7 Sonnet model configuration

ShiweiGe1999:
### Description
Bedrock Claude 3.7 use the same anthropic version. The tricky part is model id should use `us.anthropic.claude-3-7-sonnet-20250219-v1:0` in order for it to work. Hopefully this will solve the pain for bedrock users

```
{
  "modelId": "anthropic.claude-3-7-sonnet-20250219-v1:0",
  "contentType": "application/json",
  "accept": "application/json",
  "body": {
    "anthropic_version": "bedrock-2023-05-31",
    "max_tokens": 200,
    "top_k": 250,
    "stop_sequences": [],
    "temperature": 1,
    "top_p": 0.999,
    "messages": [
      {
        "role": "user",
        "content": [
          {
            "type": "text",
            "text": "hello world"
          }
        ]
      }
    ]
  }
}
```

### Test
- `<leader>a?` should be able to select the model
- be able to use the model in configuration too
```
opts = {
    -- add any opts here
    -- for example
    provider = "bedrock",
    bedrock = {
      model = "us.anthropic.claude-3-7-sonnet-20250219-v1:0"
    }
  }
```

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -334,6 +334,11 @@ M._defaults = {
       model = "claude-3-7-sonnet-20250219",
       api_key_name = "AIHUBMIX_API_KEY",
     },
+    ["bedrock-claude-3.7-sonnet"] = {
+      __inherited_from = "bedrock",
+      model = "us.anthropic.claude-3-7-sonnet-20250219-v1:0",
+      max_tokens = 4096,
+    },
   },
   ---Specify the special dual_boost mode
   ---1. enabled: Whether to enable dual_boost mode. Default to false.


lua/avante/config.lua
```diff
@@ -334,6 +334,11 @@ M._defaults = {
       model = "claude-3-7-sonnet-20250219",
       api_key_name = "AIHUBMIX_API_KEY",
     },
+    ["bedrock-claude-3.7-sonnet"] = {
+      __inherited_from = "bedrock",
+      model = "us.anthropic.claude-3-7-sonnet-20250219-v1:0",
+      max_tokens = 4096,
```
jpmcb, line 340: Looking at Anthropic's model comparison page: https://docs.anthropic.com/en/docs/about-claude/models/all-models#model-comparison - Claude 3.7 sonnet has a max tokens of 64k

Curious why you chose such a small number?

jpmcb (approved):
Great work!

lua/avante/config.lua
ShiweiGe1999: It can be defined by consumer in terms of max_tokens. In my configuration, we just defined it to 4096 for a test. As indicated here: https://github.com/yetone/avante.nvim/blob/e295fe82f0714188615a524604bdaccd266ced35/lua/avante/providers/bedrock/claude.lua#L33

PR #1874, opened 2025-04-15
Fix get_opened_buffer_by_filepath comparison

jcarlos7121:
# Pull Request Description
## Summary
This pull request addresses an issue with the buffer path matching logic in the codebase. The previous implementation used the `M.join_paths` function to construct a path by combining the `project_root` and the buffer name (`bufname`). However, this approach did not always produce the correct full path to the buffer. The `absolute_path` variable, on the other hand, contained the correct file path.

## Changes Made
- **Updated Buffer Matching Logic**: 
  - Replaced the exact path match using `M.join_paths` with a more flexible substring search.
  - The new logic checks if the `bufname` is a non-empty string and is included within the `absolute_path`.
  - This change ensures that the buffer is correctly identified by matching the buffer name against the correct file path.
## Code Changes
```lua
for _, buf in ipairs(api.nvim_list_bufs()) do
  - if M.join_paths(project_root, fn.bufname(buf)) == absolute_path then return buf end
  + if fn.bufname(buf) ~= "" and string.find(absolute_path, fn.bufname(buf)) then return buf end
end
```
---

### DIFF
--- a/lua/avante/utils/init.lua
+++ b/lua/avante/utils/init.lua
@@ -630,7 +630,7 @@ function M.winline(winid)
   return line
 end
 
-function M.get_project_root() return M.root.get() end
+function M.get_project_root() return vim.fn.getcwd() end
 
 function M.is_same_file_ext(target_ext, filepath)
   local ext = fn.fnamemodify(filepath, ":e")
--- a/README.md
+++ b/README.md
@@ -448,60 +448,26 @@ or you can use [Kaiser-Yang/blink-cmp-avante](https://github.com/Kaiser-Yang/bli
   <summary>Lua</summary>
 
 ```lua
-      file_selector = {
-        --- @alias FileSelectorProvider "native" | "fzf" | "mini.pick" | "snacks" | "telescope" | string | fun(params: avante.file_selector.IParams|nil): nil
+      selector = {
+        --- @alias avante.SelectorProvider "native" | "fzf_lua" | "mini_pick" | "snacks" | "telescope" | fun(selector: avante.ui.Selector): nil
         provider = "fzf",
         -- Options override for custom providers
         provider_opts = {},
       }
 ```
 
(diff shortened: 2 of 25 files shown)


lua/avante/utils/init.lua
```diff
@@ -910,7 +910,7 @@ local function get_opened_buffer_by_filepath(filepath)
   local project_root = M.get_project_root()
   local absolute_path = M.join_paths(project_root, filepath)
   for _, buf in ipairs(api.nvim_list_bufs()) do
-    if M.join_paths(project_root, fn.bufname(buf)) == absolute_path then return buf end
+    if fn.bufname(buf) ~= "" and string.find(absolute_path, fn.bufname(buf)) then return buf end
```
yetone, line 913: ‌‌‌‌‌‌Using `string.find` is very inaccurate.

jcarlos7121:
@yetone 

I left the method as is but modified the project_root lookup instead, now it's using the nvim api to fetch the current working directory, that way it will adapt to which ever way the user wants to open the project (Telescope projects, nvim call on folder, etc).

Let me know what you think of this, this works great to also fix the issue

yetone:
‌It looks like your branch carries too many other commits, you need to rebase it.

lua/avante/utils/init.lua
```diff
@@ -630,7 +630,7 @@ function M.winline(winid)
   return line
 end
 
-function M.get_project_root() return M.root.get() end
+function M.get_project_root() return vim.fn.getcwd() end
```
yetone, line 633: ‌‌Your modification is incorrect. It is recommended to make the relevant settings through this configuration option: https://github.com/yetone/avante.nvim/blob/b61e6571f967038a093b838a0aa434fb2d93bf7d/lua/avante/config.lua#L379

PR #1897, opened 2025-04-18
opts: allow `false` to be passed as ask flag

AntoineBalaine:
Hello,

I found this issue:
Passing `ask = false` to the `prompt_opts` results in a true. This should only be the case if `ask = null`.

Otherwise, even though `:AvanteChat` sets `ask = false`, the prompt defaults back to true.

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -2172,9 +2172,12 @@ function Sidebar:get_generate_prompts_options(request, cb)
 
   local selected_filepaths = self.file_selector.selected_filepaths or {}
 
+  local ask = self.ask_opts.ask
+  if ask == nil then ask = true end
+
   ---@type AvanteGeneratePromptsOptions
   local prompts_opts = {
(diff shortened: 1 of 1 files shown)


yetone:
‌Thank you, I need you to resolve the code conflicts.

AntoineBalaine:
Done, I made the update.

lua/avante/sidebar.lua
```diff
@@ -2174,7 +2174,7 @@ function Sidebar:get_generate_prompts_options(request, cb)
 
   ---@type AvanteGeneratePromptsOptions
   local prompts_opts = {
-    ask = self.ask_opts.ask or true,
+    ask = (self.ask_opts.ask == nil) and true or self.ask_opts.ask,
```
yetone, line 2177: ‌‌‌‌‌‌‌Perhaps having `ask` as a separate variable will make the type checker a bit happier.

AntoineBalaine: I pushed a more detailed inline. It’s annoying but it works:
```lua
ask = (self.ask_opts.ask == nil) and true or self.ask_opts.ask and true or false
```

Issue #1890, opened 2025-04-17
bug: The function name 'python' is reserved for use by this model


jfpedroza:
### Describe the bug

When using the new model o4-mini the following error appears:

```
Error: "API request failed with status 400. Body: \"{\\n  \\\"error\\\": {\\n    \\\"message\\\": \\\"The function name 'python' is reserved for use by this model. Please use a different function name.\\\",\\n    \\\"type\\\": \\\"invalid_request_error\\\",\\n    \\\"param\\\": null,\\n    \\\"code\\\": null\\n  }\\n}\""
```

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false, -- set this if you want to always pull the latest change
  opts = {
    -- add any opts here
  },
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  build = "make",
  -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
  dependencies = {
    "nvim-treesitter/nvim-treesitter",
    "stevearc/dressing.nvim",
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}
```

### Environment

neovim version 0.11.0

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

Klojer:
@jfpedroza try to use value `gpt-4o-mini`

Like this:

```
    ...
    openai = {
      model = "gpt-4o-mini", 
      ....
    },
   ...
```

mmainz:
@Klojer that's a bit besides the point. 4o and o4 are different models (yes, the naming on OpenAI's side sucks a bit). And seeing as o4 is supposed to be good at coding, that's the model we want to use.

andreasthoelke:
Hi! I'm running into the same issue using `openai/o3`:

```
- Datetime: 2025-04-29 17:53:52

- Model: openai/o3

- Selected files:
  - src/react/src/App.tsx

> the first names should start with "C".



[...]

PR #1901, opened 2025-04-20
Fix tool name conflict with o‑series models by renaming python to run_python (closes #1890)

rgurov:
When you call an o‑series model such as o4‑mini, the name python is already taken by the model’s built‑in code interpreter. Because Avante also registers a tool called python, the API returns a 400 error: “The function name 'python' is reserved for use by this model.” Issue #1890 tracks this problem.

This patch changes the tool metadata in lua/avante/llm_tools/init.lua:
- name = "python" -> name = "run_python"

To keep existing configs and scripts working, the code now adds a one‑line alias right after the original function:
```
M.run_python = M.python
```

### DIFF
--- a/lua/avante/llm_tools/init.lua
+++ b/lua/avante/llm_tools/init.lua
@@ -629,7 +629,7 @@ M._tools = {
     },
   },
   {
-    name = "python",
+    name = "run_python",
     description = "Run python code in current project scope. Can't use it to read files or modify files.",
     param = {
       type = "table",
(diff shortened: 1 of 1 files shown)


S-Spektrum-M:
I just ran a simple test  and it works great with o4-mini; looking forward to seeing this merged into main.

PanchoBubble:
This stops throwing the error for `o4-mini`, but testing this on my side suggestions were empty. Probably issue with o4-mini itself as this works perfectly fine with the change for `gpt-4.1` with this configuration

PR #1914, opened 2025-04-22
feature: Allow users to define custom parse_api_key functions.

### LINKED ISSUE #1782: feature: Support api_key function
### Feature request

There are too many problems with `api_key_name` cmd. Can we just provide a method for user to use their own lua function?

Like:

```lua
api_key = function()
  -- Do anythings you like to get your api key.
end
```

### Motivation

`api_key_name = "cmd:bw get item llmproxy | jq -r .login.password"`

This return error

```
failed to get key: (error code1)
error: unknown option '-r'
```

Using string parsing is indeed too fragile and lacks scalability.

https://github.com/yetone/avante.nvim/blob/e16111ba341511bb6aff30444bf00e9aa85461a1/lua/avante/providers/init.lua#L126

Ref: https://github.com/yetone/avante.nvim/issues/305#issuecomment-2326020931

### Other

_No response_

gitsang:
Close #1782

Allow users to define custom `parse_api_key` functions.

```
opts = {
  vendors = {
    ["myvendor"] = {
      __inherited_from = "openai",
      endpoint = "https://openai-proxy.com/",
      model = "claude-3.7-sonnet",
      parse_api_key = function()
        return require("env").apikey
      end,
    },                                                                                                                                                                                                                                                                         
  },                                      
}
```

I haven't added additional conditional checks (such as `is_parse_api_key_set()`) for now because I found it involves too many elements. This should be the most minimal modification at present.

### DIFF
--- a/lua/avante/providers/init.lua
+++ b/lua/avante/providers/init.lua
@@ -183,16 +183,12 @@ M = setmetatable(M, {
       t[k] = Utils.deep_extend_with_metatable("keep", provider_config, module)
     end
 
-    if t[k].is_env_set == nil then t[k].is_env_set = function() return E.parse_envvar(t[k]) ~= nil end end
-
-    if t[k].parse_api_key == nil then
-      t[k].parse_api_key = function() return E.parse_envvar(t[k]) end
-    else
-      t[k].is_env_set = function() return t[k].parse_api_key() ~= nil end
-    end
+    if t[k].parse_api_key == nil then t[k].parse_api_key = function() return E.parse_envvar(t[k]) end end
 
     -- default to gpt-4o as tokenizer
     if t[k].tokenizer_id == nil then t[k].tokenizer_id = "gpt-4o" end
+
+    if t[k].is_env_set == nil then t[k].is_env_set = function() return t[k].parse_api_key() ~= nil end end
 
     if t[k].setup == nil then
       local provider_conf = M.parse_config(t[k])
--- a/lua/avante/providers/init.lua
+++ b/lua/avante/providers/init.lua
@@ -183,12 +183,16 @@ M = setmetatable(M, {
       t[k] = Utils.deep_extend_with_metatable("keep", provider_config, module)
     end
 
-    t[k].parse_api_key = function() return E.parse_envvar(t[k]) end
+    if t[k].is_env_set == nil then t[k].is_env_set = function() return E.parse_envvar(t[k]) ~= nil end end
+
+    if t[k].parse_api_key == nil then
+      t[k].parse_api_key = function() return E.parse_envvar(t[k]) end
+    else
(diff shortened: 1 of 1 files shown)


lua/avante/providers/init.lua
```diff
@@ -183,12 +183,19 @@ M = setmetatable(M, {
...
+    if t[k].parse_api_key == nil then t[k].parse_api_key = function() return E.parse_envvar(t[k]) end end
 
     -- default to gpt-4o as tokenizer
     if t[k].tokenizer_id == nil then t[k].tokenizer_id = "gpt-4o" end
 
-    if t[k].is_env_set == nil then t[k].is_env_set = function() return E.parse_envvar(t[k]) ~= nil end end
+    if t[k].is_env_set == nil then
+      t[k].is_env_set = function()
+        -- If parse_api_key is customized (not the default one), use it to determine if API is set
+        if t[k].parse_api_key ~= E.parse_envvar then return t[k].parse_api_key() ~= nil end
```
yetone, line 194: ‌‌‌‌‌‌This judgment is always false, right? `t[k].parse_api_key ~= E.parse_envvar`

gitsang: No, because in line 186:

```lua
if t[k].parse_api_key == nil then t[k].parse_api_key = function() return E.parse_envvar(t[k]) end end
```

If the user does not set `parse_api_key`, it will use the default `parse_envvar` as the API Key parsing method. If the user sets `parse_api_key`, it will use the user's method.

The logic for `t[k].is_env_set` below is:
- If the user sets `parse_api_key`, then `is_env_set` is `true` (here, it determines whether `parse_api_key` is set by the user based on whether it is set to `E.parse_envvar`) (perhaps it could be changed to `if t[k].parse_api_key ~= E.parse_envvar then return true end`, because line 186 will always set `t[k].parse_api_key` to a non-null value, but I think this is a closure and should ensure its own robustness).
- If the user does not set `parse_api_key`, then `is_env_set` follows the original logic, checking whether the env is set.

Therefore: `t[k].parse_api_key ~= E.parse_envvar` is used to determine whether `parse_api_key` is set by the user, and it is not always false.

However, you reminded me that I think the following pattern would be more readable:

```go
if t[k].is_env_set == nil then t[k].is_env_set = function() return E.parse_envvar(t[k]) ~= nil end end

if t[k].parse_api_key == nil then
  t[k].parse_api_key = function() return E.parse_envvar(t[k]) end
else
  t[k].is_env_set = function() return true end
end
```

What do you think? If you think the below is better, I will modify my PR.

yetone: It is precisely because of L186 that `t[k].parse_api_key` is always a newly created anonymous function, and it will never be equal to `E.parse_envvar`.

gitsang: My mistake, I didn't notice this, I have revised the PR.

PR #1928, opened 2025-04-27
fix: support rag search on macos

PeterCardenas:
rag_search does not work on macos for the following reasons:
- `--network=host` doesn't work reliably on macos to expose ports
  - I fixed this by specifying the ports to expose in the `docker run` command
- the container somehow does not have `git` installed
  - this might've been because i was building for macos, but the change to install it in the `Dockerfile` didn't work
- the `build-image` step builds for the current platform, even though the `docker run` command runs for linux
  - I added the `--platform linux/amd64` flag to the `docker build` command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,5 +106,8 @@ lua-typecheck:
 
 .PHONY: build-image
 build-image:
-	docker build -t $(RAG_SERVICE_IMAGE) -f py/rag-service/Dockerfile py/rag-service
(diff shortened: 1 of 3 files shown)


xiaomi388:
Hello, I think this change causes the rag service not able to access the local ollama server:

```
root@a267b96c6d23:/app# curl http://localhost:11434
curl: (7) Failed to connect to localhost port 11434 after 0 ms: Couldn't connect to server
```

As a workaround, I had to use the docker gateway ip so that it can connect:

```
root@a267b96c6d23:/app# curl http://192.168.9.1:11434
Ollama is runningroot@a267b96c6d23:/app#
```

PR #1930, opened 2025-04-27
feat: Add ability to expire cached credentials

DanielMe:
I tried using avante with the Bedrock provider and I found it quite nice to obtain the credenitals using the command as `api_key_name` convention: It allows a config such as the following to get the credentials string from an AWS profile:

```lua
  opts = {
    provider = "bedrock",
    -- I'm not sure if auto suggestions are even active but if I do not set this to bedrock, I will still be prompted
    -- for a anthropic key for no reason
    auto_suggestions_provider = "bedrock",
    bedrock = {
      model = "anthropic.claude-3-5-sonnet-20241022-v2:0",
      api_key_name = { "sh", "-c", "REGION=$(aws configure get region --profile bedrock) && aws configure export-credentials --profile bedrock | jq -r --arg region \"$REGION\" '[.AccessKeyId, .SecretAccessKey, $region, .SessionToken] | join(\",\")'"},
    },
  },
```

The `bedrock` aws profile gives me short-lived credentials which are refreshed using the `credentials_process` using SSO. (https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-sourcing-external.html).

The only problem with this setup is that Avante caches these credentials and I have to restart nvim once they expire.

My proposal in this PR is to add an optional parameter `reevaluate_api_key_after` which invalidates the credenitals cache after a set number of seconds.

Note that there are two different caches that need to be invalidated, one in `providers/init.lua` and one in `environment.lua`. My idea was to add the expiry to one of them and a force invalidation flag to the other.
Note furthermore, that the command execution in `environment.lua` is async which is fine for startup where we do not want to block on this evaluation - however when we invalidate the credentials, we actually want to wait for the credentials to be available because the old credentials are likely no longer available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -582,6 +582,17 @@ Given its early stage, `avante.nvim` currently supports the following basic func
 > ```
 >
 > Note: The aws_session_token is optional and only needed when using temporary AWS credentials
+>
+> If you use temporary credentials which you refresh using a credentials_process in your AWS profile, you can
+> configure your api key to reference a shell command as follows:
+> ```lua
+>    bedrock = {
+>      -- ...
+>      api_key_name = { "sh", "-c", "REGION=$(aws configure get region --profile bedrock) && aws configure export-credentials --profile bedrock | jq -r --arg region \"$REGION\" '[.AccessKeyId, .SecretAccessKey, $region, .SessionToken] | join(\",\")'"},
+>      reevaluate_api_key_after = 3500, -- reevaluate the API key once per hour to get fresh credentials
+>    }
+> ```
+
 
 1. Open a code file in Neovim.
 2. Use the `:AvanteAsk` command to query the AI about the code.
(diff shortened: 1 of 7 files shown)


Qkessler:
Thanks so much for this! Just pulled the PR and was able to work with the README instructions without issues! Like the proposed approach. As a step forward, the bedrock integration could be abstracted further to work with either:

1. The current static credentials approach.
2. Abstract the integration to allow the user to configure a `profile` that auto-refreshes. Avante.nvim would then take the profile and configure itself to get up-to-date credentials, modifying the cache.

Thoughts?

cc @yetone for input on the vision for Bedrock support.

yetone:
hello, can you help resolve the code conflicts?

DanielMe:
Hey, apologies for the delay, I resolved the merge conflicts and validated that it did not break my nvim config

yetone:
hello, can you help me to fix the ci failed and code conflicts?

yetone:
Sorry, due to too many conflicts with this one, I might create a new PR referencing your implementation to complete this task. I'm very sorry.

PR #1931, opened 2025-04-27
feat(setup): make proper flake

aarnphm:
```lua
{
  "yeton/avante.nvim",
  build = "nix run .#plugin"
}
```

To startup rag, just do `nix run .#rag-service`


aarnphm:
@yetone when you have time :)

PR #1932, opened 2025-04-27
fix(rag_search): ignore global git ignore files

PeterCardenas:
I was getting errors in the rag service about not being able to read files. This was because I had files in the directory that symlinked to files outside of the host directory (specifically python in virtualenv). This was not getting properly ignored by the existing ignore scripts because I had a global ignore for a `venv` path that wasn't getting captured. I thought that in `scan_directory`, I should filter out unreadable paths. This seems like a valid change, but more of a hack than just ignoring the global git ignore file. I can add that change, but adding the global ignore seems to work better for me for now.

### DIFF
--- a/lua/avante/rag_service.lua
+++ b/lua/avante/rag_service.lua
@@ -74,7 +74,7 @@ function M.launch_rag_service(cb)
       M.stop_rag_service()
     end
     local cmd_ = string.format(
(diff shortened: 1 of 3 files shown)


PeterCardenas:
oops i just realized that the host mount is configurable so probably need to change the approach here

PeterCardenas:
@ftassi how do you feel about mounting the global git ignore unconditionally? another option could send a sanitized version of the git ignore (without comments) via network request to the service, but seems a bit overkill 🫠

PR #1941, opened 2025-04-30
fix: prevent extmark line number from being negative in render_state

momexn:
- Add boundary check to prevent negative line numbers when using sidebar
- Use math.max to ensure line number is always >= 0
- Fixes 'Invalid line: out of range' error that occurs when sidebar is invoked through advanced shortcuts configuration described in https://github.com/yetone/avante.nvim/wiki/Recipe-and-Tricks#advanced-shortcuts-for-frequently-used-queries-756

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -1835,7 +1835,8 @@ function Sidebar:render_state()
     { { string.rep(" ", padding) }, { virt_line, hl } },
   }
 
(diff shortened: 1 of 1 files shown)


momexn:
https://github.com/yetone/avante.nvim/wiki/Recipe-and-Tricks#advanced-shortcuts-for-frequently-used-queries-756

After implementing the settings from the URL above, enter Visual Mode.
For example, when entering <leader>ac, an error like shown in the image appears.

スクリーンショット 2025-04-30 124116

Issue #1872, opened 2025-04-14
bug: Popup Hint


v3ceban:
### Describe the bug

Popup hint doesn't disappear after avante window is closed or toggled off when there is anything typed inside the input box. Sometimes  it also doesn't disappear after sending a message. The hint stays floating on the side and obstructs the view. Opening and closing avante window makes it disappear. Resizing the window makes it jump to the top of the buffer.

This was first discovered in #1713 and attempted to be fixed in #1734 but the problem is still present.

![Image](https://github.com/user-attachments/assets/d0840660-5dfa-4d1d-a1e2-a2e808579680)
![Image](https://github.com/user-attachments/assets/4f20c107-d905-4975-9b8a-abdff4a07510)

### To reproduce

1. Open avante window
2. Type something and exit to normal mode
3. Toggle avante window

### Expected behavior

Closing or toggling avante window should make the popup hint disappear.

### Installation method

```lua
-- lazy
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false,
  opts = require "configs.avante",
  build = "make",
  dependencies = {
    "nvim-treesitter/nvim-treesitter",
    "stevearc/dressing.nvim",
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}

-- configs.avante
local function tokens(num)
  return num * 1024
end

local opts = {
  provider = "copilot",
  auto_suggestions_provider = "copilot",
  copilot = {
    endpoint = "https://api.githubcopilot.com", -- needs Copilot plugin authentication
    model = "claude-3.5-sonnet",
    max_tokens = tokens(64),
  },
  openai = {
    endpoint = "https://api.openai.com/v1", -- needs OPENAI_API_KEY env variable
    model = "o3-mini",
    max_completion_tokens = tokens(64),
    reasoning_effort = "high",
  },
  ollama = {
    endpoint = "http://127.0.0.1:11434",
    model = "qwen2.5-coder", -- "ollama ls" for available models
    options = {
      num_ctx = tokens(32),
    },
    stream = true,
  },
  vendors = {
    ["copilot:3.7-thought"] = {
      __inherited_from = "copilot",
      model = "claude-3.7-sonnet-thought",
      max_tokens = tokens(64),
    },
    ["copilot:3.7"] = {
      __inherited_from = "copilot",
      model = "claude-3.7-sonnet",
      max_tokens = tokens(64),
    },
[...]

PR #1950, opened 2025-04-30
feat(hints): add config to toggle submit hint popup

v3ceban:
Addresses #1872

Introduce a new `submit_hint` option in the hints config to allow enabling or disabling the submit hint popup in prompt input. Updated sidebar and prompt input logic to respect this setting, preventing the hint from showing when disabled. This provides users with more control over UI hints.


yetone:
‌Thank you for your PR. This bug has been fixed, so this PR can be closed.

PR #2000, opened 2025-05-07
fix(provider.openai) reasoning_content maybe empty string

soulfy:
in some providers developed based on OpenAI, reasoning_content may be an empty string.

### DIFF
--- a/lua/avante/providers/openai.lua
+++ b/lua/avante/providers/openai.lua
@@ -257,9 +257,7 @@ function M:parse_response(ctx, data_stream, _, opts)
     if provider_conf.model:match("o1") then delta = choice.message end
   end
   if not delta then return end
-  if delta.reasoning_content and delta.reasoning_content ~= vim.NIL
-    and choice.delta.reasoning_content ~= "" then
-
+  if delta.reasoning_content and delta.reasoning_content ~= vim.NIL and choice.delta.reasoning_content ~= "" then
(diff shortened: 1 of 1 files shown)


lua/avante/providers/openai.lua
```diff
@@ -257,7 +257,9 @@ function M:parse_response(ctx, data_stream, _, opts)
     if provider_conf.model:match("o1") then delta = choice.message end
   end
   if not delta then return end
-  if delta.reasoning_content and delta.reasoning_content ~= vim.NIL then
+  if delta.reasoning_content and delta.reasoning_content ~= vim.NIL
+    and choice.delta.reasoning_content ~= "" then
```
yetone, line 261:
```suggestion
    and delta.reasoning_content ~= "" then
```

lua/avante/providers/openai.lua
```diff
@@ -257,7 +257,7 @@ function M:parse_response(ctx, data_stream, _, opts)
     if provider_conf.model:match("o1") then delta = choice.message end
   end
   if not delta then return end
-  if delta.reasoning_content and delta.reasoning_content ~= vim.NIL then
+  if delta.reasoning_content and delta.reasoning_content ~= vim.NIL and choice.delta.reasoning_content ~= "" then
```
yetone, line 260:
```suggestion
  if delta.reasoning_content and delta.reasoning_content ~= vim.NIL and delta.reasoning_content ~= "" then
```

PR #2001, opened 2025-05-07
fix(Sidebar): fix: Hide default statusline when laststatus=0

aniaan:
## Summary

This PR improves the visual experience when using sidebar windows in Neovim with `vim.opt.laststatus = 0`. Specifically, it hides the default status line content in sidebar windows by setting their local `statusline` option to a single space character.

This change ensures a cleaner and more minimal UI, especially in setups where the global status line is disabled.

### Before & After Comparison

Below are screenshots demonstrating the effect of this change when `vim.opt.laststatus = 0`:

**Before:**

an1

**After:**

an2


yetone:
TIL, thanks!

PR #2006, opened 2025-05-07
fix(sidebar): toggle from empty selected files container

PeterCardenas:
Was trying to toggle selected file from empty and it errored. turned out to be an easy fix, nui just sets the winid to `nil` on unmount so we should just check for it.

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -2839,10 +2839,10 @@ function Sidebar:create_selected_files_container()
       return
     end
 
-    if not self.selected_files_container or not api.nvim_win_is_valid(self.selected_files_container.winid) then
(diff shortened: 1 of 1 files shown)


lua/avante/sidebar.lua
```diff
@@ -2838,7 +2838,11 @@ function Sidebar:create_selected_files_container()
       return
     end
 
-    if not self.selected_files_container or not api.nvim_win_is_valid(self.selected_files_container.winid) then
+    if
+      not self.selected_files_container
+      or not self.selected_files_container.winid
+      or not api.nvim_win_is_valid(self.selected_files_container.winid)
+    then
```
yetone, line 2845: maybe you can use `Utils.is_valid_container`

PeterCardenas:
@yetone updated 🙏

PR #2019, opened 2025-05-09
fix: always retry on 429 error

PeterCardenas:
In the history refactoring, we ended up not retrying when the response had an error message. Here we indiscriminately retry whenever a 429 error is returned.


zhouatie:
Is your provider Copilot?

PeterCardenas:
@zhouatie no i use azure openai

PR #2053, opened 2025-05-19
fix: gemini & vertex tools

### LINKED ISSUE #1949: bug: attempt to index field 'content' (a nil value) with gemini provider
### Describe the bug

When trying to run tools calling, neovim complains about attempting to index field 'content' (a nil value) when using the gemini provider

### To reproduce

Use Avante with tools calling

### Expected behavior

calling tools successfully without errors

### Installation method

Use lazy.nvim:
```lua
{
  "MunifTanjim/nui.nvim",
  "nvim-telescope/telescope-ui-select.nvim",

  {
    "nvim-telescope/telescope.nvim",
    event = "User FilePost",
    config = function()
      require "configs.telescope"
    end,
  },

  {
    "HakonHarnes/img-clip.nvim",
    event = "User FilePost",
    opts = {
      default = {
        prompt_for_file_name = false,
        drag_and_drop = {
          insert_mode = true,
        },
        use_absolute_path = true,
      },
    },
  },

  {
    opts = {
    },
  },

  {
    "yetone/avante.nvim",
    version = false,
    build = "make",
    cmd = { "AvanteAsk", "AvanteChat", "AvanteEdit", "AvanteToggle", "AvanteModels" },
    opts = {
      provider = "gemini",
    },
  },
}
```

### Environment

neovim version: 0.11.1
wsl2

### Repro

```lua

```

kernitus:
* Check if `candidate.content` and `candidate.content.parts` exist before iterating. Fixes #1949 
* Handle all Gemini stop sequence reasons
* Add tool calling for Gemini when used through Vertex (this assumes vertex provider is only used for Gemini-like models, as i see there is a separate provider for Vertex Claude)
* Refactor out common code in `parse_curl_args` between Gemini and Vertex providers
* Have only tested with Vertex provider and `gemini-2.5-pro-preview-05-06`, do not know if it will affect other models

### DIFF
--- a/lua/avante/providers/gemini.lua
+++ b/lua/avante/providers/gemini.lua
@@ -85,11 +85,18 @@ function M:parse_messages(opts)
           role = "function"
           local ok, content = pcall(vim.json.decode, item.content)
           if not ok then content = item.content end
+          -- item.name here refers to the name of the tool that was called,
+          -- which is available in the tool_result content item prepared by llm.lua
+          local tool_name = item.name
+          if not tool_name then
+            -- Fallback, though item.name should ideally always be present for tool_result
+            tool_name = tool_id_to_name[item.tool_use_id]
+          end
           table.insert(parts, {
             functionResponse = {
-              name = tool_id_to_name[item.tool_use_id],
+              name = tool_name,
               response = {
-                name = tool_id_to_name[item.tool_use_id],
+                name = tool_name, -- Gemini API requires the name in the response object as well
                 content = content,
               },
             },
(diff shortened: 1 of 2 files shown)


lua/avante/providers/gemini.lua
```diff
@@ -155,51 +216,61 @@ function M:parse_response(ctx, data_stream, _, opts)
...
-        opts.on_stop({ reason = "complete" })
+      local reason_str = candidate.finishReason
+      local stop_details = { finish_reason = reason_str }
+
+      if reason_str == "TOOL_CODE" then
+        -- Model indicates a tool-related stop.
+        -- The tool_use_list passed here is from the current chunk.
+        -- llm.lua will use its master list of accumulated tools for the turn.
+        opts.on_stop(
+          vim.tbl_deep_extend("force", { reason = "tool_use", tool_use_list = tool_use_list }, stop_details)
```
yetone, line 231: ‌‌‌‌‌There is no need to manually pass `tool_use_list` now, as it will be retrieved from the history messages.

https://github.com/yetone/avante.nvim/blob/e98bb8d3bae102662c5ad4ada1ce63a9aacac4e0/lua/avante/llm.lua#L869-L885

lua/avante/providers/gemini.lua
```diff
@@ -155,51 +216,61 @@ function M:parse_response(ctx, data_stream, _, opts)
...
+        )
+      elseif reason_str == "STOP" then
+        if #tool_use_list > 0 then
+          -- Natural stop, but tools were found in this final chunk.
+          opts.on_stop(
+            vim.tbl_deep_extend("force", { reason = "tool_use", tool_use_list = tool_use_list }, stop_details)
```
yetone, line 237: ‌‌‌‌‌‌Same as above.

lua/avante/providers/gemini.lua
kernitus: Ok have removed the tool_use_list passing

retarpt:
Thank you so much, @kernitus and @yetone! This was **much** needed.

PR #2065, opened 2025-05-22
fix: handle JSON parsing errors in OpenAI tool_use_message

matmoscicki:
I tried to add my own custom_tool, which does not accept arguments. In this situation, there was an error serializing the empty string “” to json, and another one when nil was interpreted as an empty array ( [] ), not an empty object ( {} ), causing an internal server error.


matmoscicki:
Looks like the problem is fixed

PR #2075, opened 2025-05-24
feat(prompt_logger): Save prompts to disk and add buffer navigation via keymaps

hamidi-dev:
This PR adds a prompt logging utility with two main features:

1. **Prompt Persistence**:
   - All submitted prompts are saved to disk under a configurable `log_dir`.
   - File names include timestamps (e.g., `avante_prompt_20250524_133012.log`).
   - If enabled, a random fortune message is shown after successful save.

2. **In-Buffer Prompt Navigation**:
   - Users can cycle through previous prompt logs directly inside the input buffer.
   - Configurable keybindings for normal/insert mode:
     ```lua
     prompt_logger = {
       enabled = true,
       log_dir = vim.fn.stdpath("cache") .. "/avante_logs",
       fortune_cookie_on_success = false,
       next_prompt = {
         normal = "<C-n>", insert = "<C-n>",
       },
       prev_prompt = {
         normal = "<C-p>", insert = "<C-p>",
       },
     }
     ```
   - Navigation is **clamped**:
     - `<C-n>` shows the next (newer) log but stops at the most recent one.
     - `<C-p>` goes backwards in time until the oldest log is reached.
     - No wrap-around.

Ideal for recalling, reusing, or iterating on past prompts—right from the input buffer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,6 +358,19 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
     minimize_diff = true, -- Whether to remove unchanged lines when applying a code block
     enable_token_counting = true, -- Whether to enable token counting. Default to true.
   },
+  prompt_logger = { -- logs prompts to disk (timestamped, for replay/debugging)
+    enabled = true, -- toggle logging entirely
+    log_dir = vim.fn.stdpath("cache") .. "/avante_prompts", -- directory where logs are saved
+    fortune_cookie_on_success = false, -- shows a random fortune after each logged prompt (requires `fortune` installed)
+    next_prompt = {
+      normal = "<C-n>", -- load the next (newer) prompt log in normal mode
+      insert = "<C-n>",
+    },
+    prev_prompt = {
+      normal = "<C-p>", -- load the previous (older) prompt log in normal mode
+      insert = "<C-p>",
+    },
+  },
   mappings = {
     --- @class AvanteConflictMappings
     diff = {
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -387,6 +387,11 @@ M._defaults = {
     use_cwd_as_project_root = false,
     auto_focus_on_diff_view = false,
   },
+  prompt_logger = {
+    enabled = true,
+    log_dir = vim.fn.stdpath("cache") .. "/avante_logs",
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -358,6 +358,19 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
     minimize_diff = true, -- Whether to remove unchanged lines when applying a code block
     enable_token_counting = true, -- Whether to enable token counting. Default to true.
   },
+  prompt_logger = { -- logs prompts to disk (timestamped, for replay/debugging)
+    enabled = true, -- toggle logging entirely
+    log_dir = vim.fn.stdpath("cache") .. "/avante_logs", -- directory where logs are saved
```
yetone, line 363: ‌‌It seems we need to use Utils.join_paths for Windows compatibility.

hamidi-dev: should pass now :)

yetone: ‌‌‌‌‌‌It seems that Utils.join_paths is still not being used.

lua/avante/config.lua
```diff
@@ -387,6 +387,19 @@ M._defaults = {
     use_cwd_as_project_root = false,
     auto_focus_on_diff_view = false,
   },
+  prompt_logger = { -- logs prompts to disk (timestamped, for replay/debugging)
+    enabled = true, -- toggle logging entirely
+    log_dir = vim.fn.stdpath("cache") .. "/avante_logs", -- directory where logs are saved
```
yetone, line 392: ‌‌‌‌‌‌It seems that Utils.join_paths is still not being used.

lua/avante/config.lua
```diff
@@ -387,6 +387,19 @@ M._defaults = {
     use_cwd_as_project_root = false,
     auto_focus_on_diff_view = false,
   },
+  prompt_logger = { -- logs prompts to disk (timestamped, for replay/debugging)
+    enabled = true, -- toggle logging entirely
+    log_dir = vim.fn.stdpath("cache") .. "/avante_prompts", -- directory where logs are saved
```
yetone, line 392: join_paths

mblarsen:
Oh no, going stale. The prompt logger feature would be cool.

hamidi-dev:
@yetone anything i can do on my part?

PR #2076, opened 2025-05-24
fix: opening when buffer has scheme in path

sjwall:
When opening the sidebar with a new ask from a buffer which is has as scheme for the path such as `term://` the error shown below is thrown. This fix adds a `has_scheme` check to stop adding the path to the selected files

```
error reading file: /path/to/workspace/term:/path/to/workspace/21698:/bin/bash: No such file or directory
```

### DIFF
--- a/lua/avante/file_selector.lua
+++ b/lua/avante/file_selector.lua
@@ -89,7 +89,7 @@ function FileSelector:reset()
 end
 
 function FileSelector:add_selected_file(filepath)
-  if not filepath or filepath == "" then return end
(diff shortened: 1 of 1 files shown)


sjwall:
This is still an issue, another example is with [oil.nvim](https://github.com/stevearc/oil.nvim). It uses the schema `oil://` for it's buffers. Opening from it causes the same error

yetone:
‌‌‌‌It looks like it will conflict with this PR: https://github.com/yetone/avante.nvim/pull/2077

sjwall:
Merged the latest `main`, this should resolve the merge conflict

Issue #2097, opened 2025-05-29
bug: azure  got 401


emxxjnm:
### Describe the bug

after this commit 56576c2481d82af8710d2111fef6cb2ecc57da32, azure got 401 response

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false, -- set this if you want to always pull the latest change
  opts = {
    -- add any opts here
  },
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  build = "make",
  -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
  dependencies = {
    "nvim-treesitter/nvim-treesitter",
    "stevearc/dressing.nvim",
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}
```

### Environment

-

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

emxxjnm:
Under configuration with only Azure settings as shown below, openai provider still loads first, causing azure.parse_api_key to not be nil. As a result, avante always ends up fetching OPENAI_API_KEY instead of AZURE_OPENAI_API_KEY

```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false, -- set this if you want to always pull the latest change
  opts = {
    azure = {
      endpoint = "https://<your-resource-name>.openai.azure.com",
      deployment = "gpt-4o", 
      api_version = "2025-01-01-preview",
      max_completion_tokens = 16384, 
    },
  },
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  build = "make",
  -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
  dependencies = {
    "nvim-treesitter/nvim-treesitter",
    "stevearc/dressing.nvim",
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}
```

PR #2098, opened 2025-05-29
fix: exclude parse_api_key for azure when inhert from OpenAI

emxxjnm:
fix [2097](https://github.com/yetone/avante.nvim/issues/2097)


yetone:
‌‌‌I'm wondering if there is a more scientific approach to solve this problem.

LukePeltier:
@yetone any more thoughts on this? azure is completely broken without a fix

yetone:
Thank you for your PR, this issue has been fixed in this PR: https://github.com/yetone/avante.nvim/pull/2252

Issue #1278, opened 2025-02-14
Feature Request: Replace Deprecated dressing.nvim Dependency


smarquez1:
### Feature request

[dressing.nvim](https://github.com/stevearc/dressing.nvim) is archived and the author suggests using snacks.nvim instead. 

More Info: See [this issue](https://github.com/stevearc/dressing.nvim/issues/190) for more details.

foxt451:
Would be cool to limit its usage to avante at least, because I don't like this input appearing for everything, including lsp renames

RulerOfTheQueendom:
@smarquez1 @foxt451 Hi! I took a stab at this in the PR linked above - I tested it on my own machine with my configs and it seems to work quite well, but I could use a couple other pairs of hands with different config settings to cross-confirm that the changes in the diff aren't likely to break other people's builds.

smarquez1:
> [@smarquez1](https://github.com/smarquez1) [@foxt451](https://github.com/foxt451) Hi! I took a stab at this in the PR linked above - I tested it on my own machine with my configs and it seems to work quite well, but I could use a couple other pairs of hands with different config settings to cross-confirm that the changes in the diff aren't likely to break other people's builds.

Hey that's great! please link your PR and I'll take a look when possible

PR #2101, opened 2025-05-30
Switch from using Deprecated `dressing.nvim` to `snacks.nvim`

RulerOfTheQueendom:
Should close https://github.com/yetone/avante.nvim/issues/1278 if I've done everything correctly


RulerOfTheQueendom:
(Also, thanks for this project, @yetone !)

yetone:
‌‌‌‌‌‌Thank you for your PR. In order to streamline avante.nvim dependencies, I will gradually remove dressing-related code in the future, and of course, I don't want to rely on snacks.nvim either.

PR #2106, opened 2025-05-31
feat: support multiple models in one provider

80avin:
Multiple changes:
1. Support endpoint redirection for copilot, which is used by enterprise users.
2. Support having multiple models in one provider.
   Note: You will see more models in avante than in vscode or CopilotChat.nvim because they filter out models having multiple versions. In this implementation, I am showing all the versions.

image

### DIFF
--- a/lua/avante/model_selector.lua
+++ b/lua/avante/model_selector.lua
@@ -1,4 +1,5 @@
 local Utils = require("avante.utils")
+local Providers = require("avante.providers")
 local Config = require("avante.config")
 local Selector = require("avante.ui.selector")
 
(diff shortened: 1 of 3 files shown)


yetone:
LGTM

tiagolupepic:
Thanks for taking your time and implement this! 👏

repparw:
Sorry, how do we use this?

```
copilot = {
          model = "claude-3.5-sonnet";
};
```

I had this, do I duplicate copilot entry with a different model? Do I add another model line, do i concat models name somehow?

80avin:
@repparw What happens when you run `:AvanteModels`.
Don't you see list of all models from copilot ?

repparw:
my bad, apparently my pkg still is on 0.23

PR #2161, opened 2025-06-04
feat: allow overriding provider headers

80avin:
Use case:
* Supporting custom authorization while keeping rest of logic same.
* Override `Editor-Version`, `Copilot-Integration-Id` in case neovim isn't supported ( our case. )

### DIFF
--- a/lua/avante/providers/azure.lua
+++ b/lua/avante/providers/azure.lua
@@ -58,7 +58,7 @@ function M:parse_curl_args(prompt_opts)
     ),
     proxy = provider_conf.proxy,
     insecure = provider_conf.allow_insecure,
-    headers = headers,
+    headers = Utils.tbl_override(headers, self.headers),
     body = vim.tbl_deep_extend("force", {
       messages = self:parse_messages(prompt_opts),
       stream = true,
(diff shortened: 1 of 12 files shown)


yetone:
‌‌‌‌‌‌Perhaps we should reuse the `extra_headers` configuration item, currently only openai is using it.

https://github.com/yetone/avante.nvim/blob/12bdb66ea4749f9f59861e7165a4de6cfbf4eb22/lua/avante/providers/openai.lua#L500-L504

80avin:
Okay. This can be done but one question! Is it fine to call it `extra_headers` when it can be used to override existing headers also?
I am feeling it should be `headers`, seeing that they can override the headers and aren't meant as extra headers only. But, will pick whatever is your call!

yetone:
If headers are used, users might be confused about whether they need to include Authorization and other headers, otherwise the headers would be incomplete.

80avin:
Honestly, I am not convinced as this is a very common pattern across the neovim plugins.
For example: plugin config uses `opts`, `keys`, etc, which are merged with existing `opts`, `keys`. In case, the user doesn't want them to be merged, he can provide a function.
codecompanion also uses the term `headers`, though there it can be only an object and not a function.

In any case, I have done the refactor on your word. At the end, it is only one variable name and too much debate may not be worth it.

PR #2173, opened 2025-06-05
feat: allow custom input provider, removing dressing.nvim

80avin:
Note: Though reviewed manually and confirmed to be working for all 3 input providers, the code is generated by Claude 4.0 sonnet

I wanted to use snacks.nvim instead of dressing.nvim, but seeing a previous attempt at adding snacks.nvim as input provider fail in https://github.com/yetone/avante.nvim/pull/2101 , I went with the approach of giving this flexibility to the user.


80avin:
The failing check appears to be a bug in snacks.nvim and existing issue in this repo also.
i.e. not introduced by this PR.

Issue #2172, opened 2025-06-05
bug: High CPU usage ( profiler data attached )


80avin:
### Describe the bug

I was running a small refactor of ~5 files when I noticed the laptop is almost frozen.
The neovim process is running at ~100% CPU utilization and the UI is lagging significantly.

During this, I managed to take a profiler dump using `Snacks.profiler`

![Image](https://github.com/user-attachments/assets/8b8030d4-8369-4100-a1a8-8e143a5fa92a)

It is using https://github.com/oraios/serena MCP to replace text in files since the mcphub's replace_text

Apparently, there was a memory leak also. It happened to utilize 64GB of my RAM and started using swap, which has never happened while using this large RAM with me

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false, -- set this if you want to always pull the latest change
  opts = {
    -- add any opts here
  },
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  build = "make",
  -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
  dependencies = {
    "nvim-treesitter/nvim-treesitter",
    "stevearc/dressing.nvim",
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}
```

### Environment

neovim: 0.12.0-dev ( nightly )

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

80avin:
Looks like render_state is fast enough but was called simply too many times.

Can we add some throttle/debounce to limit these many calls or is there possibility of other optimizations ?

xuyangyuio:
Same issue here.

PR #2174, opened 2025-06-06
fix: streaming diff debouncing

yetone:


80avin:
Does this fix https://github.com/yetone/avante.nvim/issues/2172 ?

I haven't been able to recreate it, but it appears to me that it should.
@yetone  ?

PR #2219, opened 2025-06-13
Dont remove indent in replace_in_file

hicder:
In JSON, removing indentation makes its match the `SEARCH` block in the wrong place


yetone:
‌‌If the indent is not removed, it will be difficult to match because LLMs usually do not return code snippets perfectly. This is especially true for models with performance inferior to claude-4 and gemini-2.5-pro, which often make many mistakes with spaces.

Issue #2094, opened 2025-05-28
bug: "No permission to access path" - replace_in_file tooling not working on windows


grichard99:
### Describe the bug

Hello, I'd love some help on this as I've been struggling for a while now and can't seem to figure it out.

Each time I use the avante chat or ask, I get this response when it tries to actually edit the file I asked it to make changes on:

 **replace_in_file  failed
│   running tool
│   path: C:/Users/guyri/Documents/Git/buggy_adventures/dragonruby-windows-amd64/mygame/app/main.rb
╰─  Error: No permission to access path: C:/Users/guyri/Documents/Git/buggy_adventures/dragonruby-windows-amd64/mygame/app/main.rb**

On a different terminal (git bash, pwsh, etc), it gives the same error just different tool, e.g:

 **insert  failed
│   running tool
│   path: C:/Users/guyri/Documents/Git/buggy_adventures/dragonruby-windows-amd64/mygame/app/main.rb
╰─  Error: No permission to access path: C:/Users/guyri/Documents/Git/buggy_adventures/dragonruby-windows-amd64/mygame/app/main.rb**

That path is just my current directory, but it happens on any file in any directory.

I'm currently using Neovim on windows 11, with lazy.nvim package manager

### I've currently tried:

1) Running as administrator
2) Running on different types of terminals
3) Changing my neovim default terminal and giving it admin permissions in init.lua
4) Changing files to full access on windows security/removing any "read-only" statuses
5) Disabling all firewall/antivirus configurations
6) Building from source instead of automatically
7) Changing the LLM (tried claude sonnet 4 and chat-gpt-4o-mini)

I can edit files just fine through neovim and other plugins, but avante can't seem to get access, so I've come to you for help.

Thank you in advance!

### To reproduce

Use AvanteAsk or AvanteChat to make edits to an open file

### Expected behavior

Expected: Be able to make edits to the desired files I'm working on

### My lazy.nvim avante plugin file is:
```
return {
  {
    'yetone/avante.nvim',
    dir = vim.fn.stdpath('data') .. '/lazy/avante.nvim', -- Point to local clone
    event = 'VeryLazy',
    build = 'powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource true', -- Windows-specific build
    dependencies = {
      'stevearc/dressing.nvim',
      'nvim-lua/plenary.nvim',
      'MunifTanjim/nui.nvim',
      'nvim-tree/nvim-web-devicons',
    },
[...]

PR #2248, opened 2025-06-17
fix(#2094): fix the path resolving in windows

guanghechen:


guanghechen:
@yetone  I have a confusion on the empty string path case, shouldn't it get `.` instead of ''? Since the normalize will always return `.` instead of ` `

image


```lua
  it("should return empty string when no paths provided", function()
    local result = utils.join_paths()
    assert.equals("", result)
  end)
```

yetone:
> @yetone I have a confusion on the empty string path case, shouldn't it get `.` instead of ''? Since the normalize will always return `.` instead of ` `
> 
> image
> 
> ```lua
>   it("should return empty string when no paths provided", function()
>     local result = utils.join_paths()
>     assert.equals("", result)
>   end)
> ```

It feels like there is no standard answer to this question.

image

guanghechen:
Yeah, but since there are already defined the behavior of the `path.relative(empty string)` to `.`, do you mind I change the test case to let the `join_paths()` return `.`?

image

brewinski:
I think this might have introduced a bug when opening Avante with an empty buffer. 

When I start Neovim and press <leader>aa, all of the files in the current directory are opened and I get the following error:
image

This is on MacOS 15. Anyone else running into this bug?

guanghechen:
will check it soon

guanghechen:
Hum, seems cannot repro it in my macos device

image

image

guanghechen:
I think I find what the issue is — is your current buffer a directory (like the cwd)? I believe that when opening Avante from a directory buffer, it shouldn't be added to the selected files list. will try to draft a pr to fix it.

guanghechen:
@brewinski can you try this fix?

https://github.com/yetone/avante.nvim/pull/2254

brewinski:
Thanks @guanghechen for looking into this so quickly. Yes, #2254 change fixed the problem for me. :)

Issue #2130, opened 2025-06-03
bug:


chyuaner:
### Describe the bug

I hosted my Ollama server and create Nginx reverse proxy.
At reversy proxy is set header auth like Open AI API Key is `Authorization: bearer <key>`.
I found commit e8c5f4f13ebdfd8706f436415dcd3aa41349f9e0 is supported.

Before you changed providers configuration format. My config is work.
But after upgrade your package. Is failed. (I updated my config move in `providers` )
Is always show: `ollama: API key not set, continuing without authentication` and get 403 error.
![Image](https://github.com/user-attachments/assets/f5911073-9098-4ec5-bd7b-51dbf12d0734)

---

After debugging, 
I found `avante.nvim/lua/avante/providers/ollama.lua` is always exec this line:
```lua
if Providers.env.require_api_key(provider_conf) then
  local api_key = self.parse_api_key() -- ← api_key is Always get nil
  if api_key and api_key ~= "" then
    headers["Authorization"] = "Bearer " .. api_key
  else -- ← Is Always execute here
    Utils.info((Config.provider or "Provider") .. ": API key not set, continuing without authentication")
  end
end
```

As a workaround, I had to manually override the `parse_api_key` function like this:
```lua
providers = {
  ollama = {
    endpoint = "http://ai.test.local",
    model = "qwen2.5-coder:7b",
    api_key_name = "OLLAMA_API_KEY",
    parse_api_key = function()
      return vim.env.OLLAMA_API_KEY
    end,
  },
},
```
With this override, everything works as expected and the Authorization header is added.

### Expected behavior

The plugin should correctly read the API key via `api_key_name` and `vim.env`, without requiring a manual override.

### Installation method

Use lazy.nvim (After update new providers format):
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  version = false, -- Never set this value to "*"! Never!
  opts = {
    providers = {
      ollama = {
        api_key_name = "OLLAMA_API_KEY",
        endpoint = "http://ai.test.local",
        model = "qwen2.5-coder:7b",
        disable_tools = true, -- Open-source models often do not support tools.
      },
    },
  },
  config = function(_, opts)
    vim.env.OLLAMA_API_KEY = "my-secret-key"
    require("avante").setup(opts)
  end,
},
```

### Environment

`nvim -v`
```
NVIM v0.11.1
Build type: RelWithDebInfo
LuaJIT 2.1.1741730670
```

OS: Manjaro Linux

[...]

PR #2252, opened 2025-06-17
fix: parse_api_key inheritance problem

rakhnin:
The current solution leads to inheritance problem. it defines `parse_api_key`, `is_env_set`, `setup` only if they are not defined at all (including parents), but  body of the functions relay on object (hard `self`) from closure.
E.g. if `azure` provider class is inherited from `openai`, `parse_api_key` will be defined ONLY once with `provider_config` for `openai`, so it reads the wrong `api_key_name`.

related issues:

* https://github.com/yetone/avante.nvim/issues/2097
* https://github.com/yetone/avante.nvim/issues/2130

related pull request:

* https://github.com/yetone/avante.nvim/pull/2098

P.S. seems this solution also is not so good as could be, but looks satisfied for me.

### DIFF
--- a/lua/avante/providers/init.lua
+++ b/lua/avante/providers/init.lua
@@ -149,12 +149,12 @@ M = setmetatable(M, {
 
     t[k] = provider_config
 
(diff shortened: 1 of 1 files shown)


yetone:
cool

PR #2258, opened 2025-06-18
fix: handle user-declined tool executions separately from errors

brewinski:
Previously, when users declined tool executions (like file edits), these were treated the same as actual tool errors. This caused user-declined operations to be incorrectly tracked as failed edits and affected history message processing.

This fix introduces an `is_user_declined` field to distinguish between:
- Actual tool execution errors
- User-declined tool executions
- Successful tool executions

Changes:
- Add `is_user_declined` field to tool result types
- Detect user decline messages in LLM streaming
- Update history processing to exclude user-declined operations from failed edit tracking
- Ensure proper context handling for AI interactions

Fixes the issue where user declining an operation would incorrectly be treated as a tool failure.

### DIFF
--- a/lua/avante/llm.lua
+++ b/lua/avante/llm.lua
@@ -682,6 +682,7 @@ function M._stream(opts)
                   tool_use_id = tool_result.tool_use_id,
                   content = tool_result.content,
                   is_error = tool_result.is_error,
+                  is_user_declined = tool_result.is_user_declined,
                 },
               },
             })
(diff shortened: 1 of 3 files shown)


khaneliman:
Thanks... was just trying to check if there was a fix to the infinite loop of trying to decline a change and it keeps trying to redo the change you just declined.

brewinski:
Yes, this PR is intended to resolve that issue. 😀

yetone:
‌‌‌‌‌Hello, is this PR ready to be merged?

PR #2316, opened 2025-06-24
fix: remove dependency on nvim-treesitter

PeterCardenas:
nvim-treesitter is moving to a new `main` branch which removes a lot of APIs that are currently on `master`. The current usage in avante is not compatible with the `main` branch. Luckily we should be able to remove the dependency all together with this change.


80avin:
@PeterCardenas I think you missed that the documentation still mentions treesitter.

PeterCardenas:
@80avin updated here: https://github.com/yetone/avante.nvim/pull/2327

Issue #1983, opened 2025-05-03
put the rust libraries in the "lua/" folder to simplify avante's loading


teto:
I had trouble using avante.nvim because it wouldn't load the rust libraries. 
https://github.com/yetone/avante.nvim/blob/main/lua/avante_lib.lua seems to massage the package.cpath to be able to find the libraries.

If you drop the libraries in `lua/`, neovim with find them automatically see `:h lua-module-load` and you dont need https://github.com/yetone/avante.nvim/blob/main/lua/avante_lib.lua anymore.

For instnace in nixpkgs this https://github.com/NixOS/nixpkgs/pull/403991 made the plugin work out of the box for me.

teto:
@yetone this should remove code (`load` can become a noop) and simplify setup.

Issue #2041, opened 2025-05-16
bug: "Make sure to build avante (missing avante_templates)" when opening AvanteAsk


dan-arnold:
### Describe the bug

When I try to open `:AvanteAsk`:
- The Avante sidebar opens
- A pane opens at the bottom showing an error `Make sure to build avante (missing avante_templates)` and a stacktrace (see reproduction section).
- Every time I dismiss the error by pressing Enter, it immediately appears again if I type a single character.

#### Error message pane appears at bottom when opening :AvanteAsk:
```
Error detected while processing ModeChanged Autocommands for "*:i":
Error executing lua callback: .../.local/share/nvim/lazy/avante.nvim/lua/avante/llm.lua:148: Make sure to build avante (missing avante_templates)
stack traceback:
        [C]: in function 'error'
        .../.local/share/nvim/lazy/avante.nvim/lua/avante/path.lua:186: in function 'get_templates_dir'
        .../.local/share/nvim/lazy/avante.nvim/lua/avante/llm.lua:148: in function 'generate_prompts'
        .../.local/share/nvim/lazy/avante.nvim/lua/avante/llm.lua:492: in function 'calculate_tokens'
        .../.local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:2090: in function 'cb'
        .../.local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:2252: in function 'get_generate_prompts_options'
        .../.local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:2089: in function 'show_input_hint'
        .../.local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:2614: in function <...local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:2612>
```

### Expected behavior

- I should be able to interact with the Avante pane without any errors, but instead I get an error message, which, when dismissed, immediately appears again when I type any character in the Avante pane.

### Installation method

Using `lazy.nvim` version `11.17.1`.

### Environment details:
Gentoo Linux in WSL2
```
$ uname -a
Linux MY_PC 5.15.167.4-microsoft-standard-WSL2 #1 SMP Tue Nov 5 00:21:55 UTC 2024 x86_64 13th Gen Intel(R) Core(TM) i7-1370P GenuineIntel GNU/Linux
```

Latest version of `yetone/avante.nvim` installed via Lazy, branch `main`, commit `d4672d7`.

```
[...]

PR #2324, opened 2025-06-24
feat: obsoletes "require("avante_lib").load()"

teto:
Fixes https://github.com/yetone/avante.nvim/issues/1983

I had trouble using avante.nvim because it wouldn't load the rust libraries. https://github.com/yetone/avante.nvim/blob/main/lua/avante_lib.lua massages the package.cpath to be able to find the libraries.

If you drop the libraries in lua/, neovim will find them automatically see :h lua-module-load. As a consequence you dont need https://github.com/yetone/avante.nvim/blob/main/lua/avante_lib.lua anymore.

For instance in nixpkgs this NixOS/nixpkgs#403991 made the plugin work out of the box for me.

I went with minimal changes for now but there are several followup possible:
- build.sh is misnamed since it does the opposite: it fetches prebuilt library instead of compiling them
- I feel like the `.tag` file should live in `stdpath('state')` or stdpath('cache') instead (under an avante.tag or a folder).
- the library checks are a bit weak, they don't try to load the library, just check the .tag. I feel it might be better to encode the version in the library name and check if such libraries exist (with the drawback one would need to clean older libraries).
- AvanteBuild without debug gives  little information to the user and is not blocking, maybe it should.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ endif
 
 LUA_VERSIONS := luajit lua51
 
-BUILD_DIR := build
+BUILD_DIR := lua
 BUILD_FROM_SOURCE ?= false
 TARGET_LIBRARY ?= all
 
@@ -68,7 +68,7 @@ $(BUILD_DIR):
 	@mkdir -p $(BUILD_DIR)
 
 clean:
-	@rm -rf $(BUILD_DIR)
+	@rm -rf $(BUILD_DIR)/*.$(EXT)
 
 luacheck:
 	@luacheck `find -name "*.lua"` --codes
--- a/build.sh
+++ b/build.sh
@@ -8,7 +8,7 @@ REPO_NAME="avante.nvim"
 SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
 
 # Set the target directory to clone the artifact
-TARGET_DIR="${SCRIPT_DIR}/build"
+TARGET_DIR="${SCRIPT_DIR}/lua"
 
 # Get the artifact download URL based on the platform and Lua version
 case "$(uname -s)" in
@@ -79,10 +79,10 @@ fi
 
 fetch_remote_tags
 latest_tag="$(git describe --tags --abbrev=0 || true)" # will be empty in clone repos
-built_tag="$(cat build/.tag 2>/dev/null || true)"
+built_tag="$(cat "${TARGET_DIR}/.tag" 2>/dev/null || true)"
 
 save_tag() {
-  echo "$latest_tag" > build/.tag
+  echo "$latest_tag" > "${TARGET_DIR}/.tag"
 }
 
 if [[ "$latest_tag" = "$built_tag" && -n "$latest_tag" ]]; then
@@ -107,6 +107,6 @@ else
   echo "No latest tag found. Building from source."
   cargo build --release --features=$LUA_VERSION
   for f in target/release/lib*.$LIB_EXT; do
-    cp "$f" "build/$(echo $f | sed 's#.*/lib##')"
+    cp "$f" "${TARGET_DIR}/$(echo $f | sed 's#.*/lib##')"
   done
 fi


teto:
Thanks awesome ! I will remove the unncessary instructions from README.

aarnphm:
@teto you probably also have to remove the `avante_lib.lua` logic to include the build path in there. Make sure that it wouldn't break on nightly neovim and nightly avante, then I'm all good.

Jonkimi:
> @teto you probably also have to remove the `avante_lib.lua` logic to include the build path in there. Make sure that it wouldn't break on nightly neovim and nightly avante, then I'm all good.

I met #2041, and I have to manually make a `build` dir, and copy dylib files to `build` dir to fix it.

I run `:lua print(package.cpath)` in my neovim, find a path for avante.nvim: `$HOME/.local/share/nvim/site/lazy/avante.nvim/lua/../build/?.dylib`

@teto `avante_lib.lua` still updates package.cpath to look for libraries in the old build/ directory.

teto:
@Jonkimi I dont understand your issue, do you mean this PR break avante for you ? your linked issue is old and mention you manipulating the runtimepath. Maybe make sure to cleanup the cache/remove .tag/libraries from old build/ folder.

@aarnphm yes that's the plan. Seems like keeping the old path might lead to avante load older libraries.

yetone:
‌‌‌‌‌When I tested it at the time, it was working fine on Linux, but today I found that it doesn't work on macOS, so I might need to revert this PR first.

teto:
@Jonkimi does this PR + https://github.com/yetone/avante.nvim/pull/2354 fix it ? (shoiuld avoid using old libraries, which is my point in https://github.com/yetone/avante.nvim/issues/2342, the library versioning is not robust)

Jonkimi:
@teto Regarding #2354, I’m still encountering the issue.

### Steps I Took

- I removed all files in avante.nvim and checked out to commit 544cc50c (from #2354).
- In Neovim, I ran `:AvanteBuild source=true`.
  - This placed the .dylib files in the lua directory as expected. 
- However, when I tried to open the Avante sidebar, I still got the following error:

```
Error detected while processing ModeChanged Autocommands for "*:i":
Error executing lua callback: ...ocal/share/nvim/site/lazy/avante.nvim/lua/avante/llm.lua:215: Make sure to build avante (missing avante_templates)
stack traceback:
        [C]: in function 'error'
        ...cal/share/nvim/site/lazy/avante.nvim/lua/avante/path.lua:215: in function 'get_templates_dir'
        ...ocal/share/nvim/site/lazy/avante.nvim/lua/avante/llm.lua:215: in function 'generate_prompts'
        ...ocal/share/nvim/site/lazy/avante.nvim/lua/avante/llm.lua:401: in function 'calculate_tokens'
        .../share/nvim/site/lazy/avante.nvim/lua/avante/sidebar.lua:2140: in function 'cb'
        .../share/nvim/site/lazy/avante.nvim/lua/avante/sidebar.lua:2584: in function 'get_generate_prompts_options'
        .../share/nvim/site/lazy/avante.nvim/lua/avante/sidebar.lua:2139: in function 'show_input_hint'
        .../share/nvim/site/lazy/avante.nvim/lua/avante/sidebar.lua:2961: in function <.../share/nvim/site/lazy/avante.nvim/lua/avante/sidebar.lua:2959>
```

### Further Investigation 

- I then checked out commit 04aeb3ca (which is before the revert commit e2d160aa) and updated avante_lib.lua as follows:


```diff
-  return dirname .. ("../build/?.%s"):format(ext)
+  return dirname .. ("./?.%s"):format(ext)
```

- After this change, the error disappeared.

### Additional Info

Here’s my current lua package.cpath in Neovim:

```
./?.so;/opt/local/lib/lua/5.1/?.so;/opt/local/lib/lua/5.1/loadall.so;/Users/jonkimi/.local/share/nvim/lazy-rocks/telescope.nvim/lib/lua/5.1/?.so;;/Users/jonkimi/.local/share/nvim
/lazy-rocks/telescope.nvim/lib64/lua/5.1/?.so;/Users/jonkimi/.local/share/nvim/site/lazy/avante.nvim/lua/./?.dylib
```

### Summary

Even with the libraries built into the lua directory, the error persists unless avante_lib.lua is updated to look for them in `lua` dir.

teto:
sry I did not update here but mentioned it at https://github.com/yetone/avante.nvim/pull/2353#issuecomment-3012835085 the problem comes from the extension. on mac it's .dylib hence it's not automatically found by neovim. Renaming those to .so should fix it. Either I do that but even better would be to have cargo builds the libraries with `.so` extension since it seems a ok extension even on darwin. Not sure how to do that though

PR #2350, opened 2025-06-27
fix: reduce the context window of copilot

yetone:


80avin:
@yetone, I'm interested in knowing how this value is derived.
I think the/models api gives the max tokens count which can be better utilised to implement such things for copilot.

The current limit, I feel, would be limiting the models having large context lengths, like Gemini, very significantly. (Unless I'm missing something)

PR #2353, opened 2025-06-27
Revert "feat: obsoletes "require("avante_lib").load()""

yetone:
Reverts yetone/avante.nvim#2324

‌‌‌‌‌When I tested it at the time, it was working fine on Linux, but today I found that it doesn't work on macOS, so I might need to revert that PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ endif
 
 LUA_VERSIONS := luajit lua51
 
-BUILD_DIR := lua
+BUILD_DIR := build
 BUILD_FROM_SOURCE ?= false
 TARGET_LIBRARY ?= all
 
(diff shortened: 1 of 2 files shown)


teto:
what is "doesn't work" so I can fix it ? I dont have a darwin machine to test

yetone:
image

Cannot read dynamic libraries from the `lua/` directory, perhaps modifying `package.cpath` could solve it, but I don't know if there is a more elegant solution.

teto:
could it be just due to the extension ? the default LUA_CPATH for interpreters look at .so files but not at `.dynlib` files which is one of mac's extension. what does `lua -e "print(require'package'.cpath)"` print on darwin ?
Looking at https://github.com/neovim/neovim/issues/21749 seems like the libraries on darwin could also use the .so extension which would solve the issue

khaneliman:
neovim still looks for `.so` on darwin. i believe i had other plugins i had to resolve this with in nixpkgs. https://github.com/NixOS/nixpkgs/pull/420358 works when I just link the `dylib` as a `so`

Issue #1939, opened 2025-04-29
bug: No permission to access path


magnusriga:
### Describe the bug

Avante complains that it does not have permission to access files:

![Image](https://github.com/user-attachments/assets/0c927413-39c0-4836-8990-43130e176ca6)

After attempting to open it, it just deleted all the code in my file and was thus not able to find the code anymore.

I have no idea what is going on here. Is this plugin working for others and used in daily coding?

### To reproduce

n/a/

### Expected behavior

Expected it to have permissions.

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false, -- set this if you want to always pull the latest change
  opts = {
    -- add any opts here
  },
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  build = "make",
  -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
  dependencies = {
    "nvim-treesitter/nvim-treesitter",
    "stevearc/dressing.nvim",
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}
```

### Environment

v0.12

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

wwnbb:
Same issue after updating

yetone:
> I do have the same error ever since updating, and the tools changed from showing the diff in the sidebar first to directly showing it in the files

You can use `mode = "legacy"` to revert to the previous mode.

magnusriga:
@yetone thank you for chiming in. Do you know if the file permission issue is fixed, or if it is a problem on my/our end?

Also, is it advisable to disable all the avante tools like described in the mcphub docs, when using the two together?

Vladipz:
@magnusriga I have the same problem and I have disabled all the servers and mcphub plugin also but still no access

MoaidHathot:
I'm also having the same issue

felipejoribeiro:
I am also having the same issue

![Image](https://github.com/user-attachments/assets/13a1e3d4-1a47-4a66-a015-41aedb5ff805)

ShangjinTang:
[...]

PR #2383, opened 2025-07-01
fix(#1939): fix the logic of ignore-checking, prefer to call the git check-ignore for git repos

guanghechen:
The old ignore logic is not robust, for example, if the .gitignore exist `out/` pattern, then all files like `src/layout/app.tsx` will be regarded as ignored.

### DIFF
--- a/lua/avante/llm_tools/helpers.lua
+++ b/lua/avante/llm_tools/helpers.lua
@@ -72,7 +72,7 @@ end
 
 ---@param abs_path string
 ---@return boolean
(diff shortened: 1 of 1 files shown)


ElliottLester:
thanks for fixing this better

ElliottLester:
Do we still need the original glob tests, or should they just be replace with tests that use use git directly?

guanghechen:
I just want to make sure it works correctly for git repo since I'm not very familiar with the pattern/regex in lua, it's hard to me to fix it. :)

PR #2448, opened 2025-07-14
File selector cleanup

dtor:
Several fixes and cleanups in the file selector module.


dtor:
Please do not use "squash merge" when pulling this but rather normal merge. I specifically split it into several logically separate changes so that they can be reviewed separately and also have a nice history in the git repository.

PR #2454, opened 2025-07-16
Split code handling message history into a separate module

dtor:
Over time utils module has become something of a kitchen sink, with a lot of unrelated functionality present in the same module. This set of changes tries to rectify this situation by moving most of the code dealing with creating, searching, and rendering messages forming the chat history into lua/avante/history module. It also tries to optimize the code where it makes sense.

This is all done in preparation for reworking history processing when composing LLM prompts/requests, as currently that code has O(n^2) complexity and is quite slow on large histories.

A word of caution: I made the code assert if messages with table content have more than one entry:

```
function M.get_tool_result_data(message)
  local content = message.message.content
  if type(content) == "table" then
    assert(#content == 1, "more than one entry in message content")
    local item = content[1]
    if item.type == "tool_result" then
      ---@cast item AvanteLLMToolResult
      return item
    end
  end
end
```

As far as I can see we don't have messages with multiple content entries, if we have multiple lines they are stored in a layer below. Please let me know if I got it wrong and I will redo the patchset. Otherwise once I clean up all code directly accessing message.message.content we can remove a level of indirection.


dtor:
Hi Yetone,

Thank you very much for honoring my request and not squashing the other pull request that I sent.

Just want to let you know that all my pull requests will have proper commit messages suitable for either normal merge or rebase merge. I would also appreciate it very much if you encourage other contributors to document their changes as well so that when working with the code base it will be easy to see why a certain piece of code was changed.

Thank you!

yetone:
> As far as I can see we don't have messages with multiple content entries, if we have multiple lines they are stored in a layer below. Please let me know if I got it wrong and I will redo the patchset. Otherwise once I clean up all code directly accessing message.message.content we can remove a level of indirection.

Sorry, I didn't understand this sentence. Do you mean to remove the `message` layer? Changing from `message.message.content` to `message.content`?

I have a separate `message` layer in the history message to distinguish between `message` and `metadata` (timestamp, visible, is_user_submission, etc.)

Also, why is `content` a list? It's to correspond with the message format of 'anthropic', but currently it seems that the type of `content` (string | table | (string | table)[]) is a bit too complex, making all code related to handling `message.content` very verbose. I've always wanted to simplify this part without disrupting existing users' historical records.

dtor:
No, I do not want to go remove "message" layer, but rather have "content" be either a string or a table, but not list of strings/tables.

We can solve history incompatibility in history reading code and keep the adapter for a year or so.

yetone:
> Hi Yetone,
> 
> Thank you very much for honoring my request and not squashing the other pull request that I sent.
> 
> Just want to let you know that all my pull requests will have proper commit messages suitable for either normal merge or rebase merge. I would also appreciate it very much if you encourage other contributors to document their changes as well so that when working with the code base it will be easy to see why a certain piece of code was changed.
> 
> Thank you!

Thank you for your recent contributions and exceptional code quality. I have added you as a Collaborator.

dtor:
> Thank you for your recent contributions and exceptional code quality. I have added you as a Collaborator.

Thank you for the recognition.

dtor:
For the record, here is the rework of history scanning and updating that I mentioned above: https://github.com/dtor/avante.nvim/tree/history-handling

It is a bit raw still so I did not include it in this pull request.

yetone:
Great, can I merge this PR now?

dtor:
Yes, I believe this PR can be merged.

PR #2477, opened 2025-07-17
fix: statusline hidden in Avante

marromlam:
Show statusbar for vim.laststatus different from zero


yetone:
Hello, has this PR reached a state where it can be reviewed?

marromlam:
yes, that is fixing the status line not appearing when laststatus != 0

PR #2497, opened 2025-07-21
History refactor - avoid poking at internals of HistoryMessage

dtor:
These patches introduce more helpers to access data in history messages so that callers do not have to be aware of exact structure of avante.HistoryMessage.


lua/avante/sidebar.lua
```diff
@@ -2098,38 +2094,6 @@ function Sidebar:add_history_messages(messages)
   end)
 end
 
----@param messages AvanteLLMMessage | AvanteLLMMessage[]
----@param options {visible?: boolean}
-function Sidebar:add_chat_history(messages, options)
```
yetone, line 2103: ‌This method is still being used by mcphub.nvim

dtor: OK. I see it also uses set_input_value()... Do you know what other external users there are? We may want to document it and ideally have a file (api.lua?) that we maintain for external users and everything else should be private/implementation details.

dtor:
Dropped the patch removing add_history_messages(), please take another look.

yetone:
‌‌Thank you for the refactoring—it made this big chunk of code much more pleasant.

PR #2504, opened 2025-07-21
feat(sidebar): allow external users identify selected code buffer

dtor:
External users such as, for example, lualine may want to identify and add special handling for Avante buffers. This is easily done for Avante result, input, selected files and todos buffers as they all set a unique file type. However the selected code buffer inherits file type from the buffer where the code came from (to have proper syntax highlighting), and so filetype can not be used to identify it as Avante buffer.

One option would be to add a custom buffer variable, such as vim.b.avante_buf_type and use it in plugins. Unfortunately lualine triggers everything on file type, so that would not work.

Solve this by setting file type of the selected code buffer to "AvanteSelectedCode" and manually activate treesitter parser with language of the source buffer if treesitter is active there. If the original code buffer uses legacy syntax highlighting then activate it. This keeps syntax highlighting active for the code while allowing external users identify Avante selected code buffer.

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -1367,8 +1367,18 @@ function Sidebar:on_mount(opts)
         Utils.lock_buf(selected_code_buf)
       end
       if self.code.bufnr and api.nvim_buf_is_valid(self.code.bufnr) then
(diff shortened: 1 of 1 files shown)


dtor:
Messed up my branches. This is a duplicate PR.

PR #2509, opened 2025-07-22
Update avante.vim

Aswan31:


dtor:
What does this do? Please provide detailed commit description of the change.

PR #2512, opened 2025-07-22
feat: implement user-defined text shortcuts

doodleEsc:
This PR introduces a shortcut system that allows users to quickly insert predefined prompt templates using the # trigger character. The feature provides intelligent completion and automatic text replacement for commonly used development tasks.
Implementation
Added shortcuts configuration with 6 default shortcuts: #refactor, #test, #document, #debug, #optimize, #security
Created ShortcutsSource completion provider for #-triggered suggestions
Integrated shortcut replacement in handle_submit function
Added AvanteShortcut type with name, description, details, and prompt fields
Benefits
Productivity: Reduces time spent writing repetitive prompts
Consistency: Ensures consistent prompt quality across team members
Customization: Allows users to define project-specific shortcuts
Integration: Seamlessly works with existing avante.nvim workflow
Usage
Type # in the input buffer to see available shortcuts. When submitted, shortcuts are automatically replaced with full prompt templates. Example: #refactor this function becomes a comprehensive refactoring request.
The system follows existing avante.nvim patterns and maintains full backward compatibility while significantly enhancing the developer experience for prompt-driven workflows.

### DIFF
--- a/lua/avante/utils/init.lua
+++ b/lua/avante/utils/init.lua
@@ -1079,7 +1079,7 @@ end
 ---@return AvanteShortcut[]
 function M.get_shortcuts()
   local Config = require("avante.config")
-  
+
   -- Built-in shortcuts
   local builtin_shortcuts = {
     {
@@ -1119,16 +1119,16 @@ function M.get_shortcuts()
       prompt = "Please perform a security review of this code, identifying potential vulnerabilities, security best practices, and recommendations for improvement.",
     },
   }
-  
+
   local user_shortcuts = Config.shortcuts or {}
   local result = {}
-  
+
   -- Create a map of builtin shortcuts by name for quick lookup
   local builtin_map = {}
   for _, shortcut in ipairs(builtin_shortcuts) do
     builtin_map[shortcut.name] = shortcut
   end
-  
+
   -- Process user shortcuts first (they take precedence)
   for _, user_shortcut in ipairs(user_shortcuts) do
     if builtin_map[user_shortcut.name] then
@@ -1140,12 +1140,12 @@ function M.get_shortcuts()
       table.insert(result, user_shortcut)
     end
   end
-  
+
   -- Add remaining builtin shortcuts that weren't overridden
   for _, builtin_shortcut in pairs(builtin_map) do
     table.insert(result, builtin_shortcut)
   end
-  
+
   return result
 end
 


doodleEsc:
[related pr](https://github.com/Kaiser-Yang/blink-cmp-avante/pull/9) in blink-cmp-avante

yetone:
COOL

80avin:
It is a great addition, something that I was personally longing for, but I have few doubts. 
Won't it be better to have this functionality as snippets ?
It would've achieved the same benefits of having easy autocompleted prompts but provided further benefits like. 
1. More transparent: The snippet would be inserted just when the user selects it instead of on submit. Leading to user knowing what exactly is being sent to the model than reading it in history to confirm it. 
2. More powerful with placeholders: Snippets can also have placeholders which can be Auto filled with computed default values or user can override them by navigating quickly using Tab key. 
3. No extra functionality required in avante or other plugins.

doodleEsc:
@80avin Thanks for the thoughtful feedback and suggestions. That's a great point.

The primary goal of this PR is to streamline the input process and maintain a clean, uncluttered interface. If the full instruction were to be displayed immediately upon selection, it could fill the input area with a large amount of text, which might compromise the elegant user experience we're aiming for.

However, your desire for snippet-like functionality is completely valid and can be fully achieved using `luasnip`'s custom snippets feature. This would give you all the benefits you mentioned (transparency, placeholders) without requiring changes to Avante or other plugins.

Here’s an example of how I have this set up in my LazyVim configuration:

```lua
-- add luasnip
{
  "L3MON4D3/LuaSnip",
  lazy = true,
  build = (not LazyVim.is_win())
      and "echo 'NOTE: jsregexp is optional, so not a big deal if it fails to build'; make install_jsregexp"
    or nil,
  dependencies = {
    {
      "rafamadriz/friendly-snippets",
      config = function()
        require("luasnip.loaders.from_vscode").lazy_load()
        -- Load custom snippets from your config directory
        require("luasnip.loaders.from_vscode").lazy_load({ paths = { vim.fn.stdpath("config") .. "/snippets" } })
      end,
    },
  },
  opts = {
    history = true,
    delete_check_events = "TextChanged",
  },
},
```

Here is the corresponding file structure for the custom snippets:
```
➜ nvim (main) ✗ tree ./snippets
./snippets
├── custom
│   └── avante.json
└── package.json

2 directories, 2 files
```

The `package.json` file links the snippets to the `AvanteInput` filetype:
```json
{
  "name": "snippets",
  "contributes": {
    "snippets": [
      {
        "language": "AvanteInput",
        "path": "./custom/avante.json"
      }
    ]
  }
}
```

And here is the custom snippet in `./snippets/custom/avante.json`:
```json
{
    "add_file_docs": {
        "prefix": "$docs",    // Using `$` as a trigger to avoid conflicts with other commands.
        "body": "Please add clear and comprehensive documentation to this code, including function descriptions, parameter explanations, and usage examples in ${0:filename}",
        "description": "Add documentation in file"
    }
}
```

With this setup, you can simply type `$docs` in the Avante input buffer, select the snippet from the completion menu, and the full prompt will be inserted. The placeholder functionality will also work perfectly, allowing you to use `Tab` to navigate.

PR #2518, opened 2025-07-23
Updated the feature to remember the last selected model

NekoNekoNiko120:


yetone:
‌Would it be better to store these configurations in `~/config/avante.nvim/config.json`?

NekoNekoNiko120:
got it

pyrightconfig.json
```diff
@@ -5,16 +5,15 @@
   "exclude": [
     "**/node_modules",
     "**/__pycache__",
-    "**/.*"
+    "**/.*",
+    "py/rag-service"
```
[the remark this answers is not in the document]
yetone: ‌‌‌‌You shouldn't exclude `py/rag-service`.

lua/avante/config.lua
```diff
@@ -684,8 +685,51 @@ M._defaults = {
...
+--- Function to save the last used model
+---@param model_name string
+function M.save_last_model(model_name, provider_name)
+  local config_dir = Utils.join_paths(vim.fn.expand("~"), ".config", "avante.nvim")
+  local storage_path = Utils.join_paths(config_dir, "config.json")
+
+  -- 确保目录存在
+  if not Utils.path_exists(config_dir) then vim.fn.mkdir(config_dir, "p") end
+
+  print("Attempting to save model and provider to:", storage_path)
```
yetone, line 697: remove this line

lua/avante/config.lua
```diff
@@ -20,6 +20,7 @@ local M = {}
 ---@class avante.Config
 M._defaults = {
   debug = false,
+  last_model = nil, -- Stores the last used model
```
yetone, line 23: ‌‌‌‌‌‌Is this field necessary?

lua/avante/config.lua
```diff
@@ -819,6 +861,15 @@ function M.setup(opts)
     }
   )
 
+  local last_model, last_provider = M.load_last_model()
+  if last_model then
+    merged.last_model = last_model
+    if last_provider then merged.provider = last_provider end
+    if merged.providers and merged.provider and merged.providers[merged.provider] then
+      merged.providers[merged.provider].model = last_model
+    end
```
yetone, line 870: ‌Add an info log to tell users which provider/model is being used, preventing complaints that changes to opts.provider didn't take effect.

PR #2529, opened 2025-07-24
fix: prevent duplicate LLM API calls with ReAct prompts

yetone:
- Add callback deduplication state management to LLM orchestrator
- Implement safe_on_stop wrapper with completion ID tracking
- Consolidate OpenAI provider callback triggers to prevent duplicates
- Add native Gemini provider methods to remove OpenAI dependencies
- Enhance ReAct parser with completion phase state machine
- Add comprehensive debug logging for callback tracking
- Fix race conditions in provider finishReason handling

This resolves the issue where use_ReAct_prompts=true caused multiple tool_use callbacks after completion, leading to unnecessary API calls.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/lua/avante/libs/ReAct_parser.lua
+++ b/lua/avante/libs/ReAct_parser.lua
@@ -358,27 +358,25 @@ function M.parse(text, parser_state)
     parser_state = {
       completion_phase = "parsing",
       tool_buffer = {},
-      last_processed_position = 0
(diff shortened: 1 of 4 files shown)


brookhong:
I also saw duplicate LLM APIs on bedrock which has no `use_ReAct_prompt` set.

image

PR #2534, opened 2025-07-26
Stop using vim.highlights and vim.loop

dtor:
These patches remove references to vim.highlights and vim.loop in favor of new vim.hl and vim.uv. Since Avante requires neovim 0.10+ doing so should be safe now. As a result luacheck is now clean: it reports no warnings.

Also rework Utils.debounce() and Utils.throttle() to use vim.defer_fn() instead of re-implementing the logic.


Steinhagen:
Shouldn't this PR be split into multiple ones? Currently the implementation does a lot more than it was originally created for. Doing everything in a single PR makes this a bit hard to parse.

PR #2536, opened 2025-07-26
feat!(selection): do not display hint immediately

dtor:
Selection hint that is displayed immediately upon entering visual mode ([<leader>aa: ask, <leader>ae: edit]) gets old pretty quickly. Add a config option to control when the hint is displayed:

  selection = {
    enabled = true,
    hint_display = "delayed",
  },

The "hint_display" option recognizes the following values:

- "immediate" results in the hint being shown immediately after entering visual mode. This is the old behavior.
- "delayed" causes the hint be displayed only if the cursor has not been moved for vim.o.updatetime milliseconds. This is the new default.
- "none" suppresses showing the hint completely.

Unfortunately "CursorHold" event is not emitted in visual mode so we have to emulate it using Utils.debounce().

This is a breaking change because selection behavior was controller by "hints" config entry which makes little sense, so the config section and associated commands were renamed to "selection".

Additionally the "hints"/"selection" was mapped to "<leader>ah", but the very same key combination was used to select from old Avante chat histories, which overrode the toggle. New selection toggle keymap is "<leader>aC".

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,7 +452,10 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
       close_from_input = nil, -- e.g., { normal = "<Esc>", insert = "<C-d>" }
     },
   },
-  hints = { enabled = true },
+  selection = {
+    enabled = true,
+    hint_display = "delayed",
+  },
   windows = {
     ---@type "right" | "left" | "top" | "bottom"
     position = "right", -- the position of the sidebar
--- a/README_zh.md
+++ b/README_zh.md
@@ -405,7 +405,10 @@ _请参见 [config.lua#L9](./lua/avante/config.lua) 以获取完整配置_
       close_from_input = nil, -- 例如，{ normal = "<Esc>", insert = "<C-d>" }
     },
   },
-  hints = { enabled = true },
+  selection = {
+    enabled = true,
+    hint_display = "delayed",
+  },
   windows = {
     ---@type "right" | "left" | "top" | "bottom"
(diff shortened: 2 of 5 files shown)


dtor:
@yetone it looks like the errors are coming from luvit-meta which is set in dependencies and which defines uv.new_timer() as returning uv_timer_t and not uv.uv_timer_t as luals does.

Why do we need this package in dependencies?

yetone:
@dtor https://www.reddit.com/r/neovim/comments/1f1q89d/why_does_kickstartnvim_import_luvitmeta/

dtor:
@yetone I see, however they luals and luvit-meta are not supposed to be used together as they conflict. Kickstart used to use libuvt-meta (but not luals 3rd library), but they switched in https://github.com/nvim-lua/kickstart.nvim/pull/1303 and lazydev made similar switch earlier.

We should pick one, but not both.

yetone:
Is this ready to be merged?

dtor:
It passes the checks but I would like to hear your opinion on the fact that it is a breaking change that will affect users who modified their configs... I do not expect many would disable the ability to select code and invoke avante.

yetone:
‌Will this affect the existing users' inability to display the hint window? I feel that the default configuration's delay is an acceptable delay.

> "delayed" causes the hint be displayed only if the cursor has not been moved for vim.o.updatetime milliseconds. This is the new default.

yetone:
LGTM

PR #2539, opened 2025-07-27
feat: support literal JSON Schemas in Claude tools

sparkleHazard:
🔧 Add Support for Literal JSON Schemas in Claude Tools

This PR introduces the ability to define **literal JSON schemas** for Claude-compatible tools in Avante by using a special opt-in flag:
```lua
input_schema.__avante_literal = true
```

✅ Why This Matters

Claude 3 and 3.5+ support tool-use with full Draft 2020-12 JSON schemas — including features like:

- `$schema`
- `additionalProperties`
- deeply nested types and enums

However, Avante currently **always rebuilds** the `input_schema` from `param.fields`, making it impossible to use manually crafted schemas. This PR fixes that by allowing devs to opt out of schema auto-generation.

🛠️ How It Works

In `transform_tool()`, this PR adds a check for the `__avante_literal` flag:

```lua
if tool.input_schema and tool.input_schema.__avante_literal then
  return {
    name = tool.name,
    description = tool.description,
    input_schema = tool.input_schema,
  }
end
```

This bypasses the schema builder and directly uses the provided literal.

🧪 Example Usage

```lua
local schema = vim.json.decode([[
  {
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "type": "object",
    "required": ["action"],
    "properties": {
      "action": {
        "type": "string",
        "enum": ["use_mcp_tool", "access_mcp_resource"]
      },
      "param": {
        "type": "object",
        "additionalProperties": true
      }
    }
  }
]])
schema.__avante_literal = true

return {
  name = "mcp",
  description = "Access MCP Hub tools and resources",
  input_schema = schema,
  param = { ... }, -- still used for UI/autocomplete
  execute = function(params)
    -- implementation
  end
}
```

🔄 Compatibility

- ✅ Backward-compatible: default schema behavior is unchanged
- ✅ Opt-in: only enabled when `__avante_literal = true`
- 🧼 Non-breaking: no external API changes or dependencies

Let me know if you'd like me to add usage docs or test helpers for this behavior. Thanks for maintaining such a great plugin! 🙌

### DIFF
--- a/lua/avante/providers/claude.lua
+++ b/lua/avante/providers/claude.lua
@@ -35,7 +35,24 @@ end
 ---@param tool AvanteLLMTool
 ---@return AvanteClaudeTool
 function M:transform_tool(tool)
-  local input_schema_properties, required = Utils.llm_tool_param_fields_to_json_schema(tool.param.fields)
+  -- ✅ Respect literal schemas (bypass generation)
+  if tool.input_schema and tool.input_schema.__avante_literal then
+    vim.notify("Avante: using literal input_schema", vim.log.levels.INFO)
+    return {
+      name = tool.name,
+      description = tool.get_description and tool.get_description() or tool.description,
+      input_schema = tool.input_schema,
+    }
+  end
+
+  -- 🟡 Fallback: auto-generate schema from param.fields
+  if not tool.param or not tool.param.fields then
+    error("Tool is missing param.fields and no literal schema provided")
+  end
+
+  local input_schema_properties, required =
+    require("avante.utils.init").llm_tool_param_fields_to_json_schema(tool.param.fields)
+
   return {
     name = tool.name,
     description = tool.get_description and tool.get_description() or tool.description,


yetone:
‌‌Thanks for the PR. Do OpenAI and Gemini also support the same JSON schema? I'm wondering if directly refactoring all of avante's tool definitions into JSON schema would be more thorough.

sparkleHazard:
OpenAI (GPT-4/3.5) uses a simplified subset of JSON Schema (roughly Draft 7), but doesn’t support $schema, additionalProperties, or full Draft 2020-12 features. Gemini’s tool inputs follow a custom format (via MakerSuite/AI Studio) and also lack full standards compliance.

So far, only Claude fully supports Draft 2020-12.

yetone:
> OpenAI (GPT-4/3.5) uses a simplified subset of JSON Schema (roughly Draft 7), but doesn’t support $schema, additionalProperties, or full Draft 2020-12 features. Gemini’s tool inputs follow a custom format (via MakerSuite/AI Studio) and also lack full standards compliance.
> 
> So far, only Claude fully supports Draft 2020-12.

So it seems the abstraction of the tool that avante built itself is quite necessary.

Then can we iterate on features based on avante’s own tool abstraction? Since in the future avante plans to implement tool use mainly through ReAct prompts, maintaining two separate abstractions for tools isn’t realistic.

https://github.com/yetone/avante.nvim/blob/daa419c79b59b92ef912b2033437f0f412721cf4/lua/avante/utils/prompts.lua#L57

PR #2542, opened 2025-07-28
fix: ReAct tool use format in few-shot

yetone:


lua/avante/llm.lua
```diff
@@ -296,6 +296,8 @@ function M.generate_prompts(opts)
 
   selected_files = vim.iter(selected_files):filter(function(file) return viewed_files[file.path] == nil end):totable()
 
+  local provider_conf = Providers.parse_config(provider)
+
   local template_opts = {
```
yetone, line 301: test

PR #2554, opened 2025-07-31
Allow running lua-typecheck and luatest locally mirroring github settings

dtor:
These changes allow running lua-typecheck and luatest on local machine in environment matching whta is used in github's workflows. lua-typecheck will install specified neovim runtime as well as specific version of luals and given plugin dependencies, and run the checks. This is the new default behavior. The ability of running lua-typecheck against "live" system is retained with "--live" flag.

luatest will download copy of plenary.nvim and ensure that local plugins and other configuration does not interfere with tests.


dtor:
Thank you for the fix to handle other OSes - I keep forgetting they exist ;)

PR #2574, opened 2025-08-02
fix: re-implement prompt logger

brookhong:
New prompt logger

1. enable user to cycle through prompt history with current user input included, for example, users can use `<C-n>` to get their input back if previous prompt was brought accidentally by user with `<C-p>`. This behaves same as shell like bash/zsh.
1. prompt history is updated instantly.
1. save all history in one file -- `~/.cache/nvim/avante_prompts.log`

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -459,8 +459,7 @@ M._defaults = {
   },
   prompt_logger = { -- logs prompts to disk (timestamped, for replay/debugging)
     enabled = true, -- toggle logging entirely
-    log_dir = Utils.join_paths(vim.fn.stdpath("cache"), "avante_prompts"), -- directory where logs are saved
-    fortune_cookie_on_success = false, -- shows a random fortune after each logged prompt (requires `fortune` installed)
+    log_dir = vim.fn.stdpath("cache"), -- directory where logs are saved
     next_prompt = {
       normal = "<C-n>", -- load the next (newer) prompt log in normal mode
       insert = "<C-n>",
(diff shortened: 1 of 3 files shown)


yetone:
LGTM

dtor:
I wonder why hooking into "TextChanged" and "TextChangedI" is required? Even though it is restricted to the input buffer this seems excessive... It is already hooked to handle_submit(). Maybe hooking to "next" and "prev" and when the sidebar is toggled would be better?

Also limiting number of entries might make sense, as well as filtering all duplicate entries, not only consecutive ones.

brookhong:
> I wonder why hooking into "TextChanged" and "TextChangedI" is required?

This is to update user input instantly, so that it can be brought back with `C-n` if users press `C-p` accidentally or intentionally to check previous input.

> Also limiting number of entries might make sense

Agree, will implement this.

> as well as filtering all duplicate entries, not only consecutive ones.

Already implemented, please see https://github.com/yetone/avante.nvim/blob/main/lua/avante/utils/promptLogger.lua#L42-L44

dtor:
> > I wonder why hooking into "TextChanged" and "TextChangedI" is required?
> 
> This is to update user input instantly, so that it can be brought back with `C-n` if users press `C-p` accidentally or intentionally to check previous input.

As I mentioned this can be handled in C-n and C-p hooks instead of doing this on each character update.

> 
> > as well as filtering all duplicate entries, not only consecutive ones.
> 
> Already implemented, please see https://github.com/yetone/avante.nvim/blob/main/lua/avante/utils/promptLogger.lua#L42-L44

Sorry, I misread the code, I initially thought you were only dropping adjacent entries. I guess rescanning entire array is fine as it will not be that large.

brookhong:
> As I mentioned this can be handled in C-n and C-p hooks instead of doing this on each character update.

Great, just missed your point.

Please see new PR on the change https://github.com/yetone/avante.nvim/pull/2581

dtor:
Thank you for doing this, it looks good.

PR #2579, opened 2025-08-03
feat: toggle code window from result container in sidebar

brookhong:
Enable user to use `x` to toggle code window, so that user can view result in sidebar with a wide window. User can also use `x` to restore the code window.

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -540,6 +540,7 @@ M._defaults = {
       edit_user_request = "e",
       switch_windows = "<Tab>",
       reverse_switch_windows = "<S-Tab>",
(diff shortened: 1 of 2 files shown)


brookhong:
@yetone Please review this, thanks.

johangp:
hey @brookhong! I found this feat really good, thanks.
I find this very useful for reading and writing prompts, especially long ones. Do you think it would be useful to allow this in the prompt windows as well? That way, you wouldn't need to jump into the result to have a wide view. However, the x key can't be used.

brookhong:
Just created a new PR for mappings in input, please see https://github.com/yetone/avante.nvim/pull/2591

PR #2580, opened 2025-08-04
fix(config): remove last used model data if it is damaged

dtor:
If file containing information about last used model contains invalid or obsolete data remove it (or rename so that it can be inspected) instead of throwing errors at user every time they open their editor.

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -683,13 +683,15 @@ M._defaults = {
 ---@diagnostic disable-next-line: missing-fields
 M._options = {}
 
+local function get_config_dir_path() return Utils.join_paths(vim.fn.expand("~"), ".config", "avante.nvim") end
+local function get_config_file_path() return Utils.join_paths(get_config_dir_path(), "config.json") end
+
 --- Function to save the last used model
 ---@param model_name string
 function M.save_last_model(model_name, provider_name)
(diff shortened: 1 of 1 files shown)


dtor:
I wonder if it would not be better to have this file stored in vim.fn.stdpath("state")/avante/ or vim.fn.stdpath("cache")/avante/ ?

from help stdpath:
cache - Cache directory: arbitrary temporary storage for plugins, etc.
data - User data directory.
state - Session state directory: storage for file drafts, swap, undo, |shada|.

Using .config directly probably is not that great option on Windows or OSX?

yetone:
I'm not sure about Windows, but on macOS, the ~/.config directory is heavily used for storing configurations.

The reason for deciding to put this configuration in ~/.config is because I've always had the idea of extracting some configurations that can be dynamically changed at runtime from lua opts and storing them as JSON files.

dtor:
> The reason for deciding to put this configuration in ~/.config is because I've always had the idea of extracting some configurations that can be dynamically changed at runtime from lua opts and storing them as JSON files.

Yes, having some persistence of state definitely makes sense and will benefit users. But I think there is crucial distinction: persistent *state* vs defined *configuration*. Configuration may be shared between environments but state is per environment.

yetone:
‌‌Makes sense, it feels more appropriate to put the last used model in `~/.state`.

PR #2591, opened 2025-08-07
feat: add mappings to toggle code window from input in sidebar

brookhong:
Enable user to toggle code window from input in sidebar, disabled by default, user need add below config to turn it on

        opts = {
            mappings = {
                sidebar = {
                    toggle_code_window_from_input = {
                      normal = "x",
                      insert = "<C-;>",
                    },
                },
            },
        },


johangp:
thanks @brookhong !

PR #2617, opened 2025-08-18
[feat] Add basic support for IBM's watsonx code assistant

SomeUserName1:
- Adds a new provider to support IBM's WCA
- fixes minor issue with error handling in sidebar.lua; `nil` and `vim.NIL` are now non-error conditions accepted from on_stop
- Add code to llm.lua to handle IBM's different API requirements (multi-part form, pre-json and base64 encoded messages from parse_curl_args)
- Adapts providers/init.lua and config.lua to include the new provider
- Adds a basic test.

Delivers #2621

### DIFF
--- a/lua/avante/providers/watsonx_code_assistant.lua
+++ b/lua/avante/providers/watsonx_code_assistant.lua
@@ -193,9 +193,7 @@ M.method_command = function(command_name)
 end
 
 local function get_iam_bearer_token(provider)
-  if M.last_iam_token_time ~= nil and os.time() - M.last_iam_token_time <= 3550 then
-    return M.iam_bearer_token
-  end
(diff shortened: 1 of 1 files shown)


SomeUserName1:
@yetone Two of the typecheck errors are because of the global scope of the language server.

It's about `parse_response_without_stream`, which has to be named like that due to the code in llm.lua when `is_disable_stream` is `true`. 

The rest should be fixed now.

SomeUserName1:
Found a workaround for the duplicate-set-field: https://github.com/LuaLS/lua-language-server/issues/2035#issuecomment-1739762563

Fixed the unit test that failed due to changing the function signature from `M.x` to `M:x`

Issue #928, opened 2024-12-07
feature: Allow Chat Sidebar to Replace the Main Window (No Split)


tonio-m:
### Feature request

I'd like to request a feature where the chat sidebar can become the main window, rather than being displayed as a split.

### Motivation

I find it frustrating that the chat window is limited to functioning as a sidebar. I can't open it as a solo buffer (or buffers - Avante and AvanteInput) in the current tab or in a new tab.

motivations:
- the sidebar size causes the model's output to wrap sometimes, which makes it harder to read.
- often I’m not actively editing a file. I’m using the chat window as the main tool instead (asking general questions or brainstorming). But currently it doesn't support this use case.
- sometimes I want to interact exclusively with the chat. Without distractions from other buffers or splits.

The split layout feels uncomfortable in these scenarios.

### Other

I've searched open and closed issues, and couldn't find this feature implemented.

Also from briefly looking at the source code. I noticed the chat is named `sidebar.lua`. Which indicated to me this feature is not implemented.

If someone has a suggestion on the proper way this could be implemented, I'd love to write a PR to solve it.

PR #2630, opened 2025-08-20
Add option custom_init to enable user to customize behaviour of sidebar

brookhong:
Add option `custom_init` to enable user to customize behavior of sidebar, for example, I can use below settings to start a full view chat with LLM

            {
                "<leader>af",
                function() require("avante.api").ask({custom_init = function(sidebar)
                    sidebar:toggle_code_window()
                    vim.api.nvim_win_set_option(sidebar.containers.result.winid, 'number', true)
                    vim.api.nvim_win_set_option(sidebar.containers.result.winid, 'relativenumber', true)
                end}) end,
                desc = "avante: ask no split",
                mode = { "n", "v" },
            },

This could also be less intrusive implementation for https://github.com/yetone/avante.nvim/issues/928


yetone:
Great, can a new opt or shortcut be created to run avante with the main window?

brookhong:
Sure, added `Config.mappings.full_view_ask` for this, default to be `<leader>am` since `<leader>af` has already been taken.

PR #2660, opened 2025-08-31
feat: supports full view in multi-window mode

yetone:


brookhong:
This does not work with me(neovim 0.11.3 on macOS), please try below steps

1. launch nvim
2. use `:split` to create another window
3. press `<leader>am` to  bring full view sidebar
4. press `<tab>` to focus in result buffer
5. press `x`, it does nothing, it should make sidebar narrow

And this also breaks `x`(toggle code window) in a single window, it works only for the first time, please try
1. launch nvim
2. press `<leader>am` to  bring full view sidebar
3. press `<tab>` to focus in result buffer
4. press `x` to make sidebar narrow
5. press `x` to make sidebar full
6. press `Esc` to close sidebar
7. press `<leader>am` to  bring full view sidebar
8. press `<tab>` to focus in result buffer
9. now `x` does not work as expected

brookhong:
@yetone Please take a look.

PR #2669, opened 2025-09-01
feat: better tool ui

yetone:


aljam3h:
#2669 #2668 #2660 #2667 #2666 #

PR #2693, opened 2025-09-06
fix: cannot bring code window back in zen mode if there are multiple windows

brookhong:
This PR is to fix the issues brought by PR #2660, which almost breaks toggle_code_window from sidebar. It also
* made toggle_code_window to fail if there are multiple windows and just show a warning message
* made update_content_with_history called as before(instead of after `sidebar_post_render`)
* removed unused fields `win_width_store`, `post_render`, `is_in_full_view` and `winids`

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -63,7 +63,6 @@ Sidebar.__index = Sidebar
 ---@field id integer
 ---@field augroup integer
 ---@field code avante.CodeState
----@field win_width_store table
 ---@field containers { result?: NuiSplit, todos?: NuiSplit, selected_code?: NuiSplit, selected_files?: NuiSplit, input?: NuiSplit }
 ---@field file_selector FileSelector
 ---@field chat_history avante.ChatHistory | nil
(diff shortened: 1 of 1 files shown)


yetone:
https://github.com/user-attachments/assets/84e87e9c-2f18-4f94-9906-ada154342920

As shown above, I can normally enter zen mode and return to the code window in a multiple window situation.

Also, this PR seems to allow me to normally enter zen mode (unable to fullscreen).

brookhong:
Please see how it works on my laptop(with neovim upgraded 0.11.4 recently), in the first half of the video it shows that `x` can not bring code window back if there are 2 windows, in the 2nd half it shows that `x` does not work properly for a single window if zen mode was entered twice.

https://github.com/user-attachments/assets/22d2875c-e30d-40bb-b7bc-502ccfb897f5

yetone:
@brookhong Thank you for providing your reproduction video, it is very helpful to me. I will try to reproduce it.

yetone:
Sorry, I've been too busy to follow up on this PR. I will catch up with it as soon as possible. My apologies.

dtor:
I do not think we need to rip out multi-window handling completely to fix these issues. I created https://github.com/yetone/avante.nvim/pull/2758 that I hope addresses the issues that you raised.

PR #2719, opened 2025-09-18
Fix: edit file and previously used model

NekoNekoNiko120:
Fixed the issues where edit_file and the function to remember the previously used model were not working.

### DIFF
--- a/lua/avante/config.lua
+++ b/lua/avante/config.lua
@@ -759,20 +759,17 @@ local function get_config_file_path() return Utils.join_paths(get_config_dir_pat
 
 --- Function to save the last used model
 ---@param model_name string
+---@param provider_name string
 function M.save_last_model(model_name, provider_name)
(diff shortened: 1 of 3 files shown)


lua/avante/config.lua
```diff
@@ -759,21 +759,16 @@ local function get_config_file_path() return Utils.join_paths(get_config_dir_pat
...
-  local Providers = require("avante.providers")
-  local provider = Providers[provider_name]
-  local provider_model = provider and provider.model
-
   local file = io.open(storage_path, "w")
   if file then
-    file:write(
-      vim.json.encode({ last_model = model_name, last_provider = provider_name, provider_model = provider_model })
-    )
+    file:write(vim.json.encode({ last_model = model_name, last_provider = provider_name }))
```
yetone, line 771: We need to save the provider_model to prevent it from becoming ineffective after the user changes the config.

Issue #2727, opened 2025-09-25
bug: Can't switch back to sider bar mode from Zen mode


ljie-PI:
### Describe the bug

I may start Zen mode first. Then I would check the code and use `avanete` in sider bar mode. But at this time, when I issue `AvanteChat`, it starts Zen mode again instead of sider bar mode.

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false, -- set this if you want to always pull the latest change
  opts = {
    -- add any opts here
  },
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  build = "make",
  -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
  dependencies = {
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}
```

### Environment

NVIM v0.11.2
Build type: Release
LuaJIT 2.1.1748459687

Linux jieliu1-laptop 5.15.167.4-microsoft-standard-WSL2 #1 SMP Tue Nov 5 00:21:55 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

ljie-PI:
Since there's already a separated command and key-binding for Zen mode. Can we use the `AvanteAsk` and `<leader>aa` specifically for sider bar mode?

ljie-PI:
> Since there's already a separated command and key-binding for Zen mode. Can we use the `AvanteAsk` and `<leader>aa` specifically for sider bar mode?

If this makes sense, please check this PR https://github.com/yetone/avante.nvim/pull/2728

ljie-PI:
Seems fixed by some other PR. Closing this

PR #2728, opened 2025-09-25
Fix: Separate zen mode and sider bar mode

ljie-PI:
Fix the issue https://github.com/yetone/avante.nvim/issues/2727


ljie-PI:
Seems this is already fixed by some other PR. Closing it now.

PR #2730, opened 2025-09-25
Add Goose as an ACP provider

ayax79:
This adds [goose](https://block.github.io/goose/) as an ACP provider. 

To get goose to work appropriately I had to:
- make sure that the previously session ids aren't passed in, because the goose acp doesn't currently support session loading
- make sure that multiple '\n' are not passed in, this causes the zed ACP library that goose uses to complain.


yetone:
LGTM

Issue #2729, opened 2025-09-25
bug: nvim-tree window becomes very narrow after toggle off Zen mode


ljie-PI:
### Describe the bug

Repo steps:
1. Open a project with nvim-tree
2. Enter Zen mode by `<leader>az`
3. Exit Zen mode by `<leader>az`
The nvim-tree window becomes very narrow.

<img width="1248" height="757" alt="Image" src="https://github.com/user-attachments/assets/eeb191aa-6e59-41df-8544-b6479217fda7" />

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false, -- set this if you want to always pull the latest change
  opts = {
    -- add any opts here
  },
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  build = "make",
  -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
  dependencies = {
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}
```

### Environment

NVIM v0.11.2
Build type: Release
LuaJIT 2.1.1748459687

Linux jieliu1-laptop 5.15.167.4-microsoft-standard-WSL2 https://github.com/yetone/avante.nvim/issues/1 SMP Tue Nov 5 00:21:55 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

ljie-PI:
<img width="677" height="353" alt="Image" src="https://github.com/user-attachments/assets/a73df854-6b07-4a0a-9044-9020247d2cc2" />

There seems to be several issues in `toggle_code_window` function
1. It seems to using `win_width == 0` to determine whether sidebar is open in Zen mode. However, in my environment (maybe with nvim-tree or other plugin), the `win_width` is `1` or `2` instead of `0` when sidebar is open in Zen mode. So I think we should use a specific `self.code.visible` to represent if code window is visible.
[...]

PR #2733, opened 2025-09-27
Fix: nvim-tree window becomes very narrow after toggle off Zen mode

ljie-PI:
Fix the issue: https://github.com/yetone/avante.nvim/issues/2729


ljie-PI:
Seems this is already fixed by some other PR. Closing it now.

Issue #2557, opened 2025-07-31
bug: `:AvanteModels` does not display ollama


mreishus:
### Describe the bug

When I use `:AvanteModels` to see the popup list of models, I don't see ollama, even though I have it configured.

My opts is:
```
  opts = {
    provider = "openrouter-sonnet",
    providers = {
      ["openrouter-sonnet"] = {
        __inherited_from = "openai", -- Required for OpenRouter compatibility
        endpoint = "https://openrouter.ai/api/v1",
        api_key_name = "AVANTE_OPENROUTER_API_KEY",
        model = "anthropic/claude-sonnet-4",
        timeout = 30000,
        -- extra_request_body = {
        --   temperature = 0.75,
        --   max_tokens = 4096,
        -- },
      },
      ["openrouter-gpt"] = {
        __inherited_from = "openai",
        endpoint = "https://openrouter.ai/api/v1",
        api_key_name = "AVANTE_OPENROUTER_API_KEY",
        model = "openai/gpt-4o",
        timeout = 30000,
      },
      ["openrouter-gemini"] = {
        __inherited_from = "openai",
        endpoint = "https://openrouter.ai/api/v1",
        api_key_name = "AVANTE_OPENROUTER_API_KEY",
        model = "google/gemini-2.5-pro",
        timeout = 60000,
      },
      ollama = {
        endpoint = "http://127.0.0.1:11434", -- Local Ollama (no /v1)
        model = "qwen3:30b-a3b-thinking-2507-q8_0", -- pull via `ollama pull`
      },
    },
```

If I type `:AvanteModels` I see everything but ollama:

<img width="751" height="204" alt="Image" src="https://github.com/user-attachments/assets/98d96dbe-0ba9-478b-8b2c-cb8e217196c5" />

I would expect to see ollama. It does work because when I `:AvanteSwitchProvider ollama` I can chat with my local qwen3.

### To reproduce

Install ollama,
run on cli: `ollama pull qwen3:30b-a3b-thinking-2507-q8_0` (downloads 30gigs, or use a model you already have)
Add
```
      ollama = {
        endpoint = "http://127.0.0.1:11434", -- Local Ollama (no /v1)
        model = "qwen3:30b-a3b-thinking-2507-q8_0", -- pull via `ollama pull`
      },
``` 
to opts,
type `:AvanteModels`.

### Expected behavior

I would expect to see ollama when typing `:AvanteModels` if it's configured and a local ollama server is running.

### Installation method

[...]

PR #2737, opened 2025-09-30
Fix: Add ollama models into model selector

ljie-PI:
Fix the issue https://github.com/yetone/avante.nvim/issues/2557


dtor (changes requested):

jekhor:
Why was it closed? The bug is still not fixed

PR #2739, opened 2025-09-30
[vertex] Allow project to be set via env var

torrancew:
This commit allows the GCP project for vertex AI to be provided via env var. As with the GCP location, the same variable that `gemini-cli` expects is honored. This allows `avante.nvim` to work predictably in environments that support multiple GCP projects, or otherwise separate the project being used for AI models from the project being used for, eg, terraform.


torrancew:
Thanks for creating this plugin!

I made this PR after attempting to use Avante.nvim at my dayjob, where we support multiple GCP projects. I can confirm that, with this change, my existing `gemini-cli` settings were picked up in `avante.nvim` and worked correctly.

PR #2760, opened 2025-10-11
feat: Allow inline buttons and popup confirmation for both ACP and normal Providers

### LINKED ISSUE #2717: feature: Make "Allow" the first option when asking for permissions instead of "Always allow"
### Feature request

Previous we had 1-key shortcuts to accept, accept all, and reject. 

Now we have 3 floating buttons that can be switched with `<Tab>` 

Image

The problem is that "Always allow" as the first option could be accepted by mistake.

Would it be possible to make "Allow" the first option? So we can just press Enter to accept once.
This would align with Claude-Code, and Gemini-cli, both show the option to accept once as the firsts option. 

And if someone really want to "Always allow" they have to take conscious actions of switching to the option, and then press enter. 

Or maybe return the 1-key per action? 

### Motivation

1. Avoid accepting all changes by mistake.
2. Align with other tools like claude-code, and gemini-cli.

### Other

_No response_

carlos-algms:
## Summary

Refactors the Confirmation Helper to allow inline buttons in the chat or the legacy floating confirmation for both normal providers and ACP providers

Fixes #2717

## Changes

- Added `confirmation_ui_style` config option ("popup" or "inline_buttons") to choose between confirmation UI styles
- Created `acp_confirm_adapter.lua` module to map ACP permission options to UI button format
- Updated `llm_tools/helpers.lua` to support inline confirmation UI with confirm_inline() function
- Changed default focus in confirm popup from "No" to "Yes" button
- Added `skip_reject_prompt` option to confirm UI to skip rejection reason input for ACP
- Extracted inline button generation logic into reusable adapter functions
- Removed unused permission_options and is_permission_confirming fields from AvanteLLMMessage type
- Updated README.md and README_zh.md with confirmation_ui_style configuration documentation
- Added icon mappings for different ACP permission option kinds (allow, allow_always, reject)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ clean:
 	@rm -rf $(BUILD_DIR)
 
 luacheck:
-	@luacheck `find -name "*.lua"` --codes
+	@luacheck `find \( -path './target' -prune \) -o -name "*.lua" -print` --codes
 
 luastylecheck:
 	@stylua --check lua/ plugin/ tests/
--- a/README.md
+++ b/README.md
@@ -500,9 +500,10 @@ _See [config.lua#L9](./lua/avante/config.lua) for the full config_
     support_paste_from_clipboard = false,
     minimize_diff = true, -- Whether to remove unchanged lines when applying a code block
     enable_token_counting = true, -- Whether to enable token counting. Default to true.
-    auto_approve_tool_permissions = false, -- Default: show permission prompts for all tools
+    auto_add_current_file = true, -- Whether to automatically add the current file when opening a new chat. Default to true.
+    auto_approve_tool_permissions = true, -- Default: auto-approve all tools (no prompts)
     -- Examples:
-    -- auto_approve_tool_permissions = true,                -- Auto-approve all tools (no prompts)
+    -- auto_approve_tool_permissions = false,                -- Show permission prompts for all tools
     -- auto_approve_tool_permissions = {"bash", "replace_in_file"}, -- Auto-approve specific tools only
     ---@type "popup" | "inline_buttons"
     confirmation_ui_style = "inline_buttons",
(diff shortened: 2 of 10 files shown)


yetone:
Haha, I specifically changed the confirmation buttons to the same style as Zed. If you prefer the previous dialog format, could it be made configurable?

carlos-algms:
Thanks for having a Look, I didn't know it was the Zed style 😅, I never used Zed. 

I removed it because your original confirmation is more feature-rich than the new buttons:
- it has a 1-key response
- a focus key-binding
- and having a unified confirmation reduces the amount of code to maintain. 

@yetone Would you consider removing the original confirmation then? And keeping only the new Zed style buttons? 
I could also try to add other features I'm missing to it, because focusing on the new buttons was my biggest usability issue.

yetone:
Personally, I prefer Zed style, and I've also heard many avante.nvim users complain that avante's confirmation dialog annoys them. However, given that there are still so many people who like the previous confirmation dialog of avante, I'm thinking of keeping both styles and allowing users to freely choose as a configuration option.

carlos-algms:
@yetone I think its ready for review again 🫡 

Both, ACP and normal providers can use any of the 2 confirmation styles:
1. `popup`: the original floating window yes,always,no
2. `inline_buttons`: the new Zed style inline Allow, Always Allow, Reject

Let me know what you think.

yetone:
LGTM! Thanks!

carlos-algms:
Thank you for merging it @yetone. 🙏🏻 😊 

I'm planning to work on the Diff and "follow the agent" features for the ACP providers. 

Is it also something you are interested in? 
Are you willing to share your contact information so we can plan this? 

Thanks!

yetone:
> Thank you for merging it @yetone. 🙏🏻 😊
> 
> I'm planning to work on the Diff and "follow the agent" features for the ACP providers.
> 
> Is it also something you are interested in? Are you willing to share your contact information so we can plan this?
> 
> Thanks!

Looking forward to it, my Discord name is `.yetone`

PR #2762, opened 2025-10-12
Stop using vim.highlights and vim.loop

dtor:
Remove references to  vim.loop in favor of new vim.uv. Since Avante requires neovim 0.10+ doing so should be safe now.


dtor:
It looks like vim.hl is only in 0.11, not 0.10...

PR #2765, opened 2025-10-13
Ollama

dtor:
Improve ollama provider:

- avoid inherinting model list from OpenAI perovider
- better handle curl errors; avoid unnecessary delays/timeouts
- provide a helper for is_env_set(), update documentation.

Note that Chinese was auto-translated, I don't speak it, sorry!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1160,14 +1160,17 @@ This approach ensures that the apply model can quickly and accurately merge your
 
 ## Ollama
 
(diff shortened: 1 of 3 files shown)


dtor:
@ljie-PI Could you please take a look at this and see if this is acceptable for you? I think the helper will make config pretty clean.

I do not have working ollama setup, so please let me know if there are issues.

PR #2773, opened 2025-10-15
fix: should not add 'add current keymap' if auto_set_keymaps is false

luweiCN:


luweiCN:
@yetone can you help to review this change

Issue #2770, opened 2025-10-14
bug: How to keep a correct fugitive horizontal split height when using avante in a vertical split in Neovim?


paulrougieux:
### Describe the bug

- When I enter `:G` in both vim and neovim, [fugitive](https://github.com/tpope/vim-fugitive) starts in a split that takes half of the screen size. There I can comfortably view changed files and stage them and write my commit message.

<img width="1308" height="849" alt="Image" src="https://github.com/user-attachments/assets/394bc6af-b3f4-4cf6-9c94-4bebefbb289a" />

- In neovim, when I start the avante side pane with leader key `aa`, then start fugitive with `:G` the height of the split is only 3 lines. This is too small to see file changes and inconvenient. 

<img width="1308" height="849" alt="Image" src="https://github.com/user-attachments/assets/71153437-2947-45ca-a7a5-d038f816ed25" />

How can I keep fugitive's split to a reasonable height when using avante? 

I can temporarily resize the split with :res +10 so that the fugitive split is high enough to be workable, but I don't want to do this each time I interact with git fugitive. 

<img width="1308" height="849" alt="Image" src="https://github.com/user-attachments/assets/21d9aa77-cd1d-4032-9869-7ce5bbf5b083" />

As you can see on the screenshot, the avante vertical split takes the right third of the screen while 2/3 remain allocated for the source code and/or text I'm viewing. Currently the fugitive split takes the whole bottom part of the screen. 

A search for avante issues related to "fugitive" doesn't correspond to this problem: https://github.com/yetone/avante.nvim/issues?q=is%3Aissue%20fugitive

### To reproduce

Install [tpope/vim-fugitive](https://github.com/tpope/vim-fugitive) and avante in neovim. start the avante side pane with leader key `aa`, then start fugitive with `:G`. 

--> The height of the split is only 3 lines.

### Expected behavior

I expect the height of the fugitive split to be at least 10 lines or to be half of the screen. The fugitive horizontal split **could be within that main code/text section, i.e. the 2/3 sized vertical split**.

<img width="1308" height="849" alt="Image" src="https://github.com/user-attachments/assets/8429186c-0c6b-449e-9a4b-e5ccd98fd88d" />

At this point I'm not sure if this is an avante issue or a fugitive issue. 

### Installation method

I use  lazy.nvim configured in  `nvim/lua/plugins/avante.lua` copied from the project's README page:
```
return {
  {
    "yetone/avante.nvim",
[...]

PR #2777, opened 2025-10-16
fix(sidebar): avoid crashing in WinNew handler

dtor:
If "WinNew" autocommand set up in Sidebar:setup_colors() fires at an inopportune time it will reference self.code.winid and error out with "Invalid window ID".

Fix this by adding check to test if code window is valid before attempting to use it. This should fix with the stack trace in #2770.

### DIFF
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -211,8 +211,8 @@ function Sidebar:setup_colors()
         if not vim.api.nvim_win_is_valid(winid) or self:is_sidebar_winid(winid) then goto continue end
         local winhl = vim.wo[winid].winhl
         if
-          winhl:find(Highlights.AVANTE_SIDEBAR_WIN_SEPARATOR)
-          and not Utils.should_hidden_border(self.code.winid, winid)
+          winhl:find("WinSeparator:" .. Highlights.AVANTE_SIDEBAR_WIN_SEPARATOR)
+          and not (vim.api.nvim_win_is_valid(self.code.winid) and Utils.should_hidden_border(self.code.winid, winid))
         then
           vim.wo[winid].winhl = self.code.old_winhl or ""
         end


dtor:
@yetone I must admit I do not quite understand what that handler (and setup_colors() in general) is trying to do. I see that on the first pass it tries to suppress the window separator for the code window if the sidebar is on right side or below the code window. What about the case when sidebar is on the left side or on the top?

Looking further, in "WinNew" it tries to restore the separator unless new split is below or to the right of the code window? I guess it is trying to again have the separator hidden between code and the sidebar (but the user will probably load something else in that split?), and then unconditionally sets the "special/hidden" separator on the original code window, so there is no separation between splits. Why? Don't we want to separate the 2 windows (not sidebar) now?

Would it make more sense to set this "hidden" separator if any side of a new window touches one of sidebar containers and restore the separator for all other windows?

yetone:
> @yetone I must admit I do not quite understand what that handler (and setup_colors() in general) is trying to do. I see that on the first pass it tries to suppress the window separator for the code window if the sidebar is on right side or below the code window. What about the case when sidebar is on the left side or on the top?
> 
> Looking further, in "WinNew" it tries to restore the separator unless new split is below or to the right of the code window? I guess it is trying to again have the separator hidden between code and the sidebar (but the user will probably load something else in that split?), and then unconditionally sets the "special/hidden" separator on the original code window, so there is no separation between splits. Why? Don't we want to separate the 2 windows (not sidebar) now?
> 
> Would it make more sense to set this "hidden" separator if any side of a new window touches one of sidebar containers and restore the separator for all other windows?

This is essentially just my aesthetic preference: I don't like borders, and I only like to use different background colors to distinguish the sidebar from the main window. Simply put, I want the style of a split window to become that of a borderless float window. That's why I'm trying hard to use different background colors and hide borders.

PR #2799, opened 2025-10-25
fix(acp): improve process cleanup and add type annotations

carlos-algms:
ACP processes were not being closed, leaving behind orphan zombie processes.

In my machine, I had more than `30` zombie processes.

The previous `transport_self.process:close()` only closes the communication between Neovim and the spawned process, but it doesn't kill the process.

Now, it sends `15` to start a graceful exit flow, and if it gets stuck, it sends a force termination with `9`. 

## Changes

- Add proper process termination with SIGTERM (15), then SIGKILL (9)
- Check if the process is closing before cleanup to prevent errors
- Add type annotations for ACPTransportInstance
- Use vim.uv with fallback to vim.loop for compatibility

### Before, original behaviour:

1. open neovim
2. open avante
3. write any prompt and submit 
4. exit with `:qa`
5. run `ps aux | grep <your acp provider name, mine is claude-code-acp>` 
6. you should see at least 1 orphan process

### Using these changes

After step 4, exit with `:qa` and then run step 5. 
You should see `0` orphan processes.

### DIFF
--- a/lua/avante/libs/acp_client.lua
+++ b/lua/avante/libs/acp_client.lua
@@ -294,13 +294,20 @@ end
 
 ---Create stdio transport layer
 function ACPClient:_create_stdio_transport()
-  local uv = vim.loop
+  local uv = vim.uv or vim.loop
+
+  --- @class avante.acp.ACPTransportInstance
   local transport = {
+    --- @type uv.uv_pipe_t|nil
     stdin = nil,
+    --- @type uv.uv_pipe_t|nil
     stdout = nil,
+    --- @type uv.uv_process_t|nil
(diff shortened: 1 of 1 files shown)


lua/avante/libs/acp_client.lua
```diff
@@ -294,13 +294,20 @@ end
 
 ---Create stdio transport layer
 function ACPClient:_create_stdio_transport()
-  local uv = vim.loop
+  local uv = vim.uv or vim.loop
```
yetone: https://github.com/yetone/avante.nvim/pull/2762/files

PR #2827, opened 2025-11-09
Exclude GPT-5-Chat

you-n-g:
Exclude GPT-5-Chat. It is not a reasoning model.


you-n-g:
@yetone  Could you please merge this PR so users can use the latest version of avante on my gpt-5-chat model

thank you.

PR #2828, opened 2025-11-10
Fix: Snacks picker formatting

jaljoue:
Fixes snacks picker format config, which was trying to display lua tables rather than the filename/model name.

# Before:
2025-11-09-220343_hyprshot

# After:
2025-11-09-220621_hyprshot


BLaurent:
Yes it is indeed thanks a lot !

PR #2838, opened 2025-11-22
Feature/lazy loading new

edmundhighcock:
This PR adds support for lazy loading of MCP tools. Apart from a set of critical tools, MCP tools are not added to the prompt unless requested. A short description of MCP tools is added to the system prompt, and a tool is provided (load_mcp_tool) to load the full tool spec into the tools section of the prompt.


edmundhighcock:
Hi All,

I have fixed the lint and type issues. Thanks for looking at this PR!

yetone:
I took a look at the code, and this PR actually makes not only the MCP tools but also the regular tools lazy-loaded, right?

edmundhighcock:
> I took a look at the code, and this PR actually makes not only the MCP tools but also the regular tools lazy-loaded, right?

Yes, apart from (1) those that are defined in the list of critical tools

```
  -- Define critical tools that should always be eagerly loaded regardless of user configuration
  local critical_tools = {
    "think",
    "attempt_completion",
    "load_mcp_tool",
    "use_mcp_tool",
    "add_todos",
    "update_todo_status",
    "list_tools",
    "dispatch_agent",
  }
```
and (2) those that are in `Config.lazy_loading.always_eager`. We could change the value of this to default to having all the internal tools as always eager, allowing the users to drop the ones they want (apart from the critical tools).

Are there other tools that should be in the list of critical tools?

yetone:
> > I took a look at the code, and this PR actually makes not only the MCP tools but also the regular tools lazy-loaded, right?
> 
> Yes, apart from (1) those that are defined in the list of critical tools
> 
> ```
>   -- Define critical tools that should always be eagerly loaded regardless of user configuration
>   local critical_tools = {
>     "think",
>     "attempt_completion",
>     "load_mcp_tool",
>     "use_mcp_tool",
>     "add_todos",
>     "update_todo_status",
>     "list_tools",
>     "dispatch_agent",
>   }
> ```
> 
> and (2) those that are in `Config.lazy_loading.always_eager`. We could change the value of this to default to having all the internal tools as always eager, allowing the users to drop the ones they want (apart from the critical tools).
> 
> Are there other tools that should be in the list of critical tools?

Thank you for the explanation, I think this is enough.

yetone:
Would it be more appropriate to call it `load_tool` or `search_tool`, because it's not just the tools of mcp that are being lazy loaded?

Regarding why it's called a search tool, I feel this feature is very similar to the `tool search tool` mentioned in this document by Anthropic: https://www.anthropic.com/engineering/advanced-tool-use

edmundhighcock:
> Would it be more appropriate to call it `load_tool` or `search_tool`, because it's not just the tools of mcp that are being lazy loaded?
> 
> Regarding why it's called a search tool, I feel this feature is very similar to the `tool search tool` mentioned in this document by Anthropic: https://www.anthropic.com/engineering/advanced-tool-use

Thanks for the link that's very interesting to hear what Anthropic are doing. I vote for `load_tool` because it is not doing any searching. I'll make the change ASAP.

edmundhighcock:
I'll get to this tomorrow hopefully.. it has been a busy week!

edmundhighcock:
Changed load_mcp_tool to load_tool... ready for review again! Thansk in advance

edmundhighcock:
Bump

edmundhighcock:
Hi everyone, any further thoughts or comments on this PR?

edmundhighcock:
I'm guessing this is not the direction you want to go. I'll close this PR in a couple of days if I don't hear back.

PR #2847, opened 2025-12-01
fix: support cancelling acp providers

PeterCardenas:
AvanteStop doesn't work with acp providers, so this PR adds support for it.


PeterCardenas:
not stale

PR #2849, opened 2025-12-03
fix(acp): return JSON-RPC error for fs/read_text_file on missing files

codefromthecrypt:
When a file doesn't exist, the on_read_file handler was ignoring the error returned by read_file_from_buf_or_disk and returning empty content. This caused agents like gemini-cli to incorrectly think the file exists but is empty, leading to infinite waits in their edit correction flow.

Now propagates errors via error_callback to send proper JSON-RPC errors.

### DIFF
--- a/tests/libs/acp_client_spec.lua
+++ b/tests/libs/acp_client_spec.lua
@@ -7,9 +7,7 @@ describe("ACPClient", function()
 
   before_each(function()
     schedule_stub = stub(vim, "schedule")
-    schedule_stub.invokes(function(fn)
-      fn()
-    end)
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
sorry hopfully got the lint passing now @yetone!

codefromthecrypt:
ok I am done editing this. would love to have this merged so i can progress to non-filesystem related glitch fixes!

codefromthecrypt:
@yetone I have some other small change needed to get gemini working. is it ok to get this merged then I can stack the next on top

PR #2856, opened 2025-12-06
chore: used native action-centric methods for testing

fearphage:
Just cleaned up the workflow to make better use of the platform.

* Updated `actions/checkout` action to latest
* switched to using `gh` CLI over `curl`
* Removed hash of date file in favor of using the date directly as a part of the cache `key`
* Made download conditionally based on cache hit vs directory existence
* Used environment variable for GitHub variables to remove injection vector


Waitzz:
oh check failed

fearphage:
@Waitzz Fixed now.

PR #2863, opened 2025-12-11
refine: replace vim.fn.system with vim.system

Waitzz:
vim.system is a modern Lua API introduced by Neovim, and its use is strongly recommended by the official documentation. Using vim.system is the inevitable trend.


yetone:
LGTM!

PR #2864, opened 2025-12-12
fix: ACP improvements - single-line retry, variable name, and cancel optimization

dascard:
## Summary
This PR includes several ACP-related fixes and improvements:
1. **Fix single-line retry/edit returning empty content**
2. **Fix ACP client variable name inconsistency**
3. **Use `cancel_session` instead of stopping ACP client**
4. **Reset sidebar state after cancellation**
### 1. Single-line retry/edit broken
When a user message is a single line, `get_current_user_request_block()` returns empty content because line 755 unconditionally removes the last element:
```lua
content_lines = vim.list_slice(content_lines, 1, #content_lines - 1)
```
For single-line messages, this results in an empty table.

2. ACP client stop not working
cancel_inflight_request() calls get_registered_acp_clients() which returns M._acp_clients, but register_acp_client() stores clients in M.acp_clients. This mismatch causes the stop function to never find any clients.

3. Full client stop causes reconnection delay
Previously, cancellation would stop the entire ACP client process, requiring a new connection for subsequent requests.

4. UI stuck in "generating" state after cancel
After cancelling, the sidebar state was not properly reset.

Solution
Only slice content_lines when there are multiple lines
Use acp_clients consistently (not _acp_clients)
Use cancel_session() to cancel the current session while keeping the connection alive
Reset current_state, scroll, and clear state extmarks after cancellation
Testing
Send a single-line message → retry/edit now works
Cancel during generation → UI properly resets
Send new message after cancel → responds without reconnection delay

### DIFF
--- a/lua/avante/history/render.lua
+++ b/lua/avante/history/render.lua
@@ -5,6 +5,75 @@ local Highlights = require("avante.highlights")
 
 local M = {}
 
+---Get the width of avante sidebar
+---@return number
+local function get_sidebar_width()
+  local Config = require("avante.config")
+  local width = Config.windows.width
+  if type(width) == "number" then
+    if width < 1 then
+      return math.floor(vim.o.columns * width)
+    end
+    return width
+  end
+  return 40  -- default fallback
+end
+
+---Helper function to wrap long lines and insert into lines table
+---@param lines avante.ui.Line[]
+---@param decoration string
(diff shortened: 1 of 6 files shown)


PeterCardenas:
part of this was superseded by https://github.com/yetone/avante.nvim/pull/2847.

PR #2872, opened 2025-12-15
fix(xdg): properly take into account xdg

teto:
get_config_dir_path would just use $HOME/.config instead of $XDG_CONFIG_HOME like neovim does. Also avante wont adapt to a neovim flavor set with NVIM_APPNAME without this change.


Waitzz:
Among all my plugins, only lazy.nvim uses this directory. Do you want to consider changing the path? I think vim.fn.stdpath("state")/avante would be better. This is more suitable for putting the running status file of the plugin.

PR #2880, opened 2025-12-21
fix tool_calls serialization for Scaleway Mistral (OpenAI Chat Completions)

silenceforest:
- Provider: Scaleway AI (OpenAI Chat Completions compatible)
- Model: mistral-small-3.2-24b-instruct-2506
- Fix: normalize tool_calls[].function.arguments to always be a JSON string ("{}" when empty) and ensure tool outputs are sent as strings.
- Reason: Scaleway/OpenAI schema rejects non‑string arguments (Lua empty table was encoded as []), causing 400 errors during MCP/tool usage.
- Also treats Scaleway endpoint as Mistral‑like for message ordering fixes (avoids tool -> user invalid sequence).
- Suppress assistant content when tool_calls are present to avoid duplicate final messages.

### DIFF
--- a/lua/avante/providers/openai.lua
+++ b/lua/avante/providers/openai.lua
@@ -69,7 +69,7 @@ function M.is_openrouter(url) return url:match("^https://openrouter%.ai/") end
 
 function M.is_mistral(url) return url:match("^https://api%.mistral%.ai/") end
 
-function M.is_mistral_like(url) return M.is_mistral(url) or url:match("^https://api%.scaleway%.ai/") end
+function M.is_scaleway(url) return url:match("^https://api%.scaleway%.ai/") end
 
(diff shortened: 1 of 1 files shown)


lua/avante/providers/openai.lua
```diff
@@ -45,7 +67,7 @@ end
 
 function M.is_openrouter(url) return url:match("^https://openrouter%.ai/") end
 
-function M.is_mistral(url) return url:match("^https://api%.mistral%.ai/") end
+function M.is_mistral(url) return url:match("^https://api%.mistral%.ai/") or url:match("^https://api%.scaleway%.ai/") end
```
Waitzz: @silenceforest The behavior does not match the function name. Can it be optimized?

PR #2882, opened 2025-12-22
feat(config): add option `allow_access_to_git_ignored_files`

pysan3:
Hi @yetone , first of all big thanks to the awesome plugin!

I found out that git-ignored files in a repository cannot be accessed by the agent.

I understand from a privacy perspective this is a good check to prevent the agents accessing `.env` files and so on (which should be git ignored).

However, I also keep logs in the repo and git-ignore them which I want the agent to be able to access.

I thought it'd be nice if there were an option that allows the access to these git-ignored files **opt-in**.

This PR adds an option `behaviour.allow_access_to_git_ignored_files` which when `true`, skips the check whether the file requested from the agent is git-ignored or not.
I kept the default to be `false` making the default bahaviour the same as before meaning that there is no accidental leakage of `.env` or other secret files unless user specifically enables this option. (nil will also fallback to _hiding_ the file as well).

This PR should provide a solution to these issues as well:

- https://github.com/yetone/avante.nvim/issues/1939
- https://github.com/yetone/avante.nvim/issues/2094

Let me hear your opinion on this!

### DIFF
--- a/lua/avante/llm_tools/helpers.lua
+++ b/lua/avante/llm_tools/helpers.lua
@@ -152,8 +152,9 @@ function M.has_permission_to_access(abs_path)
   local config_dir = vim.fn.stdpath("config")
   local in_project = abs_path:sub(1, #project_root) == project_root
   local in_config = abs_path:sub(1, #config_dir) == config_dir
+  local bypass_ignore = Config.behaviour and Config.behaviour.allow_access_to_git_ignored_files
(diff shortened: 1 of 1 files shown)


Waitzz:
There have some errors. Please check the "Actions".

pysan3:
@Waitzz Oops my bad

Fixed with 5e45fa02

Issue #2836, opened 2025-11-21
bug: Avante zen mode with codex doesn't do anything


HekpoMaH:
### Describe the bug

I have enterprise codex. The codex CLI works as normal. I don't have an openai api key.

I tried to test avante with codex, but the conversation "succeeds", without actually doing anything

<img width="1206" height="564" alt="Image" src="https://github.com/user-attachments/assets/cb857e94-ddba-48d6-b353-b4eec527cc5f" />

### To reproduce

Try creating a txt in zen mode via codex.

### Installation method

Use lazy.nvim:
```lua
return {
  "yetone/avante.nvim",
  enabled = true,
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  -- ⚠️ must add this setting! ! !
  build = function()
    -- conditionally use the correct build system for the current OS
    if vim.fn.has("win32") == 1 then
      return "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false"
    else
      return "make"
    end
  end,
  event = "VeryLazy",
  -- version = false, -- Never set this value to "*"! Never!
  -- version = "v0.0.27", -- Never set this value to "*"! Never!
  -- mode="legacy",
  ---@module 'avante'
  ---@type avante.Config
  opts = {
    -- add any opts here
    behaviour = {
      enable_fastapply=false,
      enable_cursor_planning_mode=true,
      auto_suggestions=false,
      auto_approve_tool_permissions=false,
    },
    cursor_applying_provider = "llamacpp_local_gptoss20b",
    provider = "codex",
    acp_providers = {
        ["gemini-cli"] = {
          command = "gemini",
          args = { "--experimental-acp" },
          env = {
            NODE_NO_WARNINGS = "1",
            GEMINI_API_KEY = os.getenv("GEMINI_API_KEY"),
          },
        },
        ["claude-code"] = {
          command = "npx",
          args = { "@zed-industries/claude-code-acp" },
          env = {
            NODE_NO_WARNINGS = "1",
            ANTHROPIC_API_KEY = os.getenv("ANTHROPIC_API_KEY"),
          },
        },
        ["goose"] = {
          command = "goose",
          args = { "acp" },
        },
        ["codex"] = {
          command = "codex-acp",
          env = {
            NODE_NO_WARNINGS = "1",
            -- OPENAI_API_KEY = os.getenv("OPENAI_API_KEY"),
          },
        },
      },
    providers = {
[...]

PR #2886, opened 2025-12-24
Use `kind` instead of `optionId` when checking ACP tool permissions

nielsdekker:
# Summary

Fixes an issue with the codex-acp tool. Codex uses a different value for `optionId` and `kind`. Existing code checked if the `optionId` was one of the `kind` values which it would never be and therefore all tool calls would be rejected. This only occurs when `auto_approve_tool_permissions` is set to `false`.

# Cause

Codex returned the following messages when ACP was used. 

```js
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "session/request_permission",
  "params": {
    "options": [
      { "optionId": "approved", "name": "Yes", "kind": "allow_once" },
      { "optionId": "abort", "name": "No, provide feedback", "kind": "reject_once" }
    ],
    // Other params
  }
}
```

The important thing here is that `optionId != kind`. In `/llm_tools/helpers.lua` a check was done whether the option id was `allow`, `allow_once`, etc. in order to determine if the user approved the action or not.

This check would always fail when codex was used because `approved` isn't in this list. Resulting in the tool call being rejected no matter the selected option.

# The fix

Using `kind` solves the issue of arbitrary values for the `optionId`. The `optionId` is still used for the ACP communication, in `/llm.lua:1223` it looks at a map of the options containing the correct ids to determine what to send.

> Looking at the docs the `optionId` is just a text value that could be anything. https://agentclientprotocol.com/protocol/tool-calls#param-option-id

# Sidenotes

- First PR on this repo so if anything is wrong or missing let me know
- Tested only with codex (don't have a claude code subscription so I can only assume that would still work)
- Probably fixes https://github.com/yetone/avante.nvim/issues/2836
    - Looking at this issue the used config also has `auto_approve_tool_permissions=false`

### DIFF
--- a/lua/avante/llm_tools/helpers.lua
+++ b/lua/avante/llm_tools/helpers.lua
@@ -75,13 +75,13 @@ function M.confirm(message, callback, confirm_opts, session_ctx, tool_name)
   end
 
   if Config.behaviour.confirmation_ui_style == "inline_buttons" then
-    M.confirm_inline(function(option_id)
-      if option_id == "allow" or option_id == "allow_once" or option_id == "allow_always" then
-        if option_id == "allow_always" and session_ctx then session_ctx.always_yes = true end
+    M.confirm_inline(function(kind)
+      if kind == "allow" or kind == "allow_once" or kind == "allow_always" then
+        if kind == "allow_always" and session_ctx then session_ctx.always_yes = true end
 
         callback(true)
       else
-        callback(false, option_id)
+        callback(false, kind)
       end
     end, confirm_opts or {})
     return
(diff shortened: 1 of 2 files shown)


yetone:
Will this modification affect other ACP servers?

nielsdekker:
Also tested this PR with opencode's acp option and that works. Also noticed `opencode` return the options like:

```json
"options":[
    {"optionId":"once","kind":"allow_once","name":"Allow once"},
    {"optionId":"always","kind":"allow_always","name":"Always allow"},
    {"optionId":"reject","kind":"reject_once","name":"Reject"}]
```

> _A different value for optionId and kind_

Testing against the main branch shows opencode has the same issue where all tool calls are rejected when `auto_approve_tool_permissions=false`.

> Will this modification affect other ACP servers?

I don't think so but I could be missing something. The protocol says only the `kind` value is a hardcoded list so checking that instead of optionId seems logical to me.

yetone:
I suspect it will damage other ACP servers.

nielsdekker:
> I suspect it will damage other ACP servers.

Fair, will close the PR and continue using my fork for now.

Issue #2879, opened 2025-12-21
bug: Switching to ACP provider from other providers not working


ljie-PI:
### Describe the bug

When I use "copilot" provider and then switch to acp provider "opencode" (using `:AvanteSwitchProvider opencode` or change the provider in nvim config file), I would get error when sending query again.

I have to remove the history directory in ~/.local/state/nvim/avante/projects and then start neovim again to make the switch work.

### To reproduce

<img width="1798" height="667" alt="Image" src="https://github.com/user-attachments/assets/87fa86bc-f59a-4951-840d-2f3452e34d83" />

### Expected behavior

I would expect acp providers can work after the switch.

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false, -- set this if you want to always pull the latest change
  opts = {
    ---@alias Mode "agentic" | "legacy"
    mode = "agentic",
  
    ---@alias Provider "claude" | "openai" | "azure" | "gemini" | "cohere" | "copilot" | <acp_provider>
    provider = "opencode",
    providers = {
      copilot = {
        model = "claude-sonnet-4.5",
      },
    },
    acp_providers = {
      ["opencode"] = {
        command = "opencode",
        args = { "acp" },
      },
      ["gemini-cli"] = {
        command = "gemini",
        args = { "--experimental-acp" },
      },
    },
  
    input = {
      provider = "native"
    },
  },
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  build = "make",
  -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
  dependencies = {
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}
```

### Environment

❯ nvim -v
NVIM v0.11.4
Build type: Release
LuaJIT 2.1.1753364724
Run "nvim -V1 -v" for more info

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
  spec = {
    -- add any other plugins here
  },
})
```

ljie-PI:
Actually it's not only happening when I switch ACP provider. I may got this "session not found" error even if I quit nvim and open again. I have to clean history every time.

[...]

PR #2889, opened 2025-12-26
Support loading ACP sessions if agent can

ljie-PI:
Issues described in issue https://github.com/yetone/avante.nvim/issues/2879

Now previous ACP sessions is not loaded. When session_id is not `nil`, only `initialize` is executed before "session/prompt" request. As a result, follow up queries will always fail when using Opencode ACP.


ljie-PI:
Hi @yetone , this bug blocks me from using opencode in ACP mode, could you please review this PR?

PR #2915, opened 2026-01-10
feat: show selected item for native selector

T0nd0Tara:


Waitzz:
Would it look better if the symbol was placed at the back? @T0nd0Tara 
PixPin_2026-01-12_17-15-49

T0nd0Tara:
I tried it, but with a lot of models, each with a different line length. It's difficult to quickly see which one is selected. I do preffer it this way but tbh I dont care that much, whatever you say

PR #2920, opened 2026-01-12
Make avante work again after claude blocked max subscription to be used in the 3rd party tool like avante or opencode

ghosert:
1.  Prefixes tool names with av_ for OAuth requests to bypass
Anthropic's tool name validation
  2. Updated transform_tool function - Added use_prefix parameter to conditionally prefix tool names
  3. Added user-agent header - Set to claude-cli/2.1.2 (external, cli) for OAuth requests
  4. Fixed beta headers - Removed fine-grained-tool-streaming-2025-05-14 which causes 400 errors
  5. Added ?beta=true to URL - Required for OAuth requests
  6. Always show OAuth URL - For terminal environments without browsers, URL is always displayed and copied to clipboard
  7. Prefix tool names in parse_messages - Added prefix to tool_use items when auth_type == "max"
  8. Strip tool prefix in parse_response - Removes the av_ prefix from tool names in responses
  9. Pass use_prefix to transform_tool call - Enables tool name prefixing for OAuth

Also instead of using browser to open claude auth link, i asked avante to display the auth link in the message log, so that people can copy and paste to browser if they are in a terminal env instead of GUI, like you ssh to a remote machine and want to neovim there.

in lua/avante/config.lua, just add auth_type = 'max' to make it work like below:

    claude = {
      endpoint = 'https://api.anthropic.com',
      model = 'claude-opus-4-5-20251101', -- Claude Opus 4.5 (latest)
      auth_type = 'max', -- Use Claude Max subscription via OAuth, otherwise use 'api'

Up to you guys if you want to use, but I won't be responsible for any punishment from Anthropic.

### DIFF
--- a/lua/avante/model_selector.lua
+++ b/lua/avante/model_selector.lua
@@ -110,7 +110,10 @@ function M.open()
     end)
     :totable()
 
-  local default_item = vim.iter(models):find(function(item) return item.provider == Config.provider end)
+  local current_model = Config.providers[Config.provider].model
+  local default_item = vim.iter(models):find(
+    function(item) return item.model == current_model and item.provider_name == Config.provider end
+  )
 
   local function on_select(item_ids)
     if not item_ids then return end
--- a/lua/avante/providers/claude.lua
+++ b/lua/avante/providers/claude.lua
@@ -31,7 +31,7 @@ local claude_code_spoof_prompt = "You are Claude Code, Anthropic's official CLI 
 ---@field claude_token ClaudeAuthToken?
 M.state = nil
 
-M.api_key_name = ""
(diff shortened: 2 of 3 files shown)


ghosert:
merge or.not

sebastianst:
@ghosert I can confirm this also works for Enterprise subscriptions. Maybe we shouldn't use `max` but `sub` as the auth-type name, as this works for all subscription-based authentications.

laciferin2024:
Why is no one merging this PR?

laciferin2024:
I don't think this is getting merged.. For what's worth , the opencode.nvim is better nowdays

Issue #2837, opened 2025-11-22
bug: While using copilot I am getting "invalid_tool_call_format" error where "argument" in response token has "[]" instead of "{}"


bidhovbizar:
### Describe the bug

Open any file in neovim 0.11.4 then as "Can you check again?" like the following:

● Datetime: 2025-11-22 17:33:05
● Model:    copilot/claude-sonnet-4
● Selected files:
  ○ /home/bbizar/.tmux.conf

▋ Can you check again?

Error: {
  code = "invalid_tool_call_format",
  message = "Invalid JSON format in tool call
  arguments"
}

<img width="1649" height="918" alt="Image" src="https://github.com/user-attachments/assets/774d40ee-7c50-46bb-81ba-ef18ce419e2c" />

### To reproduce

Not applicable, there is no code needed here. Just ask "Can you check again" after opening the neovim then <leader>aa then type in the chat window

### Expected behavior

The tool calling without any arguements should work fine without throwing the error.

### Installation method

Use lazy.nvim:
```lua
return {
    {
        "yetone/avante.nvim",
        lazy = true,
        version = false,
        opts = {
            provider = "copilot",
            mode = "agentic",
            providers = {
                copilot = {
                    model = "claude-haiku-4.5",
                    auto_select_model = false,
                },
            },
            hints = { enabled = true },
            selector = {
                provider = "telescope",
            },
            behaviour = {
                auto_suggestions = false,
                auto_set_highlight_group = true,
                auto_set_keymaps = true,
                auto_apply_diff_after_generation = false,
                support_paste_from_clipboard = false,
                enable_token_counting = false,
                auto_approve_tool_permissions = false,
                enable_fastapply = false,
            },
        },
        build = "make",
        dependencies = {
            "nvim-lua/plenary.nvim",
            "MunifTanjim/nui.nvim",
            "nvim-telescope/telescope.nvim",
            "stevearc/dressing.nvim",
            "nvim-tree/nvim-web-devicons",
            "nvim-treesitter/nvim-treesitter",
            "zbirenbaum/copilot.lua",
            {
                'MeanderingProgrammer/render-markdown.nvim',
                ft = { "markdown", "Avante" },
            },
        },
    },
}
```

### Environment

NVIM v0.11.5
Build type: RelWithDebInfo
LuaJIT 2.1.1741730670
Run "nvim -V1 -v" for more info

uname: Linux
Ubuntu 24.04.01

### Repro

```lua
[...]

PR #2929, opened 2026-01-20
fix: use dictionaries for empty objects

lyager:


lyager:
Fix for issue https://github.com/yetone/avante.nvim/issues/2837

andy941:
This fix solves it for me. Without it Avante is unusable, at least with copilot.

PR #2935, opened 2026-01-25
feat!: use vim.ui.select for AvanteSwitchProvider

teto:
vim.ui.input is less practical then vim.ui.select which shows all models without having to trigger completion.
Also vim.ui.select can be overriden by pickers like snack or fzf-lua (https://github.com/ibhagwan/fzf-lua#neovim-api) eg:
image

Now this can be seen as a breaking change because:
- `AvanteSwitchProvider MODEL` is not accepted anymore

In my first attempt I had `AvanteSwitchProvider PREFIX` filter the models before showing them to the user.

If you agree with my change, let me know how you want to deal with the "breaking" behavior ?
1. we change nothing. A user updates its config to `:lua require("avante.api").switch_provider("mistral")` instead ?
2. `AvanteSwitchProvider MODEL` does switch provider bypassing vim.ui.select (and/or MODEL acts as a filter)
3. we keep `AvanteSwitchProvider` and introduce `AvanteSelectProvider` instead ? since it's annoying to keep both we deprecate `AvanteSwitchProvider` ?

Worst case I can keep this change in my own init.lua but as I found this version more helpful than avante's default I prefer to upstream.

### DIFF
--- a/plugin/avante.lua
+++ b/plugin/avante.lua
@@ -123,16 +123,13 @@ cmd(
 )
 cmd("Refresh", function() require("avante.api").refresh() end, { desc = "avante: refresh windows" })
 cmd("Focus", function() require("avante.api").focus() end, { desc = "avante: switch focus windows" })
-cmd("SwitchProvider",
-  function(_opts)
-    local providers = vim.tbl_keys(Config.providers)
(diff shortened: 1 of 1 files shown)


teto:
not stale

PR #2951, opened 2026-02-03
fix(mistral): 'Tool call id was call_0fypS1hVX but must be a-z, A-Z, …

teto:
…0-9, with a length of 9'

I started having the `Tool call id was call_0fypS1hVX but must be a-z, A-Z, 0-9, with a length of 9` error when using mistral's devstral.

I had issues with the 2 tools I have updated the tool id generation for but maybe others need to be changed as well.
Also I am not able to test if this breaks other providers than mistral. From what I read such as
https://github.com/pydantic/pydantic-ai/issues/3733, mistral is more restrictive than openai so my hope is that the new tool ids work for every provider.

### DIFF
--- a/lua/avante/providers/openai.lua
+++ b/lua/avante/providers/openai.lua
@@ -445,9 +445,7 @@ end
 function M:add_tool_use_message(ctx, tool_use, state, opts)
   local jsn = JsonParser.parse(tool_use.input_json)
   -- Fix: Ensure empty arguments are encoded as {} (object) not [] (array)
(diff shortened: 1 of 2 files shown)


teto:
I was able to sucessfully test a previous version of the patch, I've changed the implementation to one that I find mor elegant but since then,  mistral tells me I've no tokens left :s

teto:
I was able to test it and this fixed avante for these tools in mistral at least. The CI doesn't seem to test against openai so it would be interesting to make sure it doesn't break other providers.

opengk:
I can confirm the solution works for Mistral AI.

Issue #2953, opened 2026-02-10
feature: Support for kiro-cli through ACP


tejbindu:
### Feature request

Kiro cli now supports ACP: https://kiro.dev/docs/cli/acp/. It would be great if Avante could support using it through acp.

davidbgonz:
As a note. When `kiro-cli acp` is ran, it will spin up a child process `kiro-cli-chat acp`. When ending a session, the process tree for kiro-cli acp needs to be closed fully or it will start orphaning the kiro-cli-chat for each new session. I still need to open a similar issue/discussion with the codecompanion.nvim folks.

Also the notification method for kiro is `_kiro.dev/commands/avaiable`. I'm not sure if an update has to be made here or the blink-cmp-avante plugin for slash commands to render in Avante.

PR #2957, opened 2026-02-16
feat: Add kiro as known ACP provider (#2953)

hador:
Adding kiro-cli as a known ACP provider.  
Most features will work out of the box, though kiro does implement some custom notifications that Avante doesn't handle yet.


hador:
Commenting because stale, any reason not to include this one?

PR #2960, opened 2026-02-17
fix: remove unconditional last line removal in user request block parsing

hador:
The pattern ^>\s+(.+)$ already filters out empty lines by requiring at least one character after the prefix. Unconditionally removing the last collected line assumed a trailing empty line always exists, which breaks retry functionality when the content structure differs.

Removing this line makes the parser robust across all scenarios without making assumptions about trailing formatting artifacts.


hador:
Replying because stale. This breaks retries in ACP, seems worth applying a one-liner fix for.

PR #2961, opened 2026-02-17
ACP improvements

hador:
This PR addresses several annoyances I ran into when using Avante with ACP and Kiro:

- **Add diagnostics support**: ACP stream prompts now include diagnostic information when `@diagnostics` is specified
- **More robust session initialization**: Eagerly create ACP sessions on new chat; this allows "slower" agents to fully load before receiving the first prompt. Perhaps questionable given it's a workaround to what is likely an agent bug, but it doesn't cost us much.
- **Prevent process leaks**: Add __gc cleanup to ACP client to properly terminate processes; this occurs consistently when failing to resume sessions.
- **Fix state management**: Clear generating state when resuming ACP sessions to avoid UI inconsistencies (Avante stuck in "generating" state on session resume)

### DIFF
--- a/lua/avante/llm.lua
+++ b/lua/avante/llm.lua
@@ -1414,6 +1414,27 @@ function M._continue_stream_acp(opts, acp_client, session_id)
       }
       table.insert(prompt, prompt_item)
     end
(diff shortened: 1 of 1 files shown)


hador:
on second thought, too many changes into one PR. 
I will wait for my existing PRs to be addressed and then split this up into manageable chunks.

PR #2968, opened 2026-02-24
ci: add job to release to luarocks.org

teto:
The current upload version https://luarocks.org/modules/neorocks/avante.nvim
is from a third party https://github.com/lumen-oss/nurr that I help with.

Avante is a big enough project that it should be in control of its luarocks release IMO.

I've taken the liberty to mention how to install avante with rocks.nvim.

For the release workflow to work, someone with a Luarocks account will have to add their API key to this repo's secrets.

NB: this is a new submission of https://github.com/yetone/avante.nvim/pull/2946 . While I can understand autoclosing issues, I find autoclosing PRs which usually are higher effort contributions than tickets really hostile. Especially after 15 days.

### DIFF
--- a/.github/workflows/luarocks.yml
+++ b/.github/workflows/luarocks.yml
@@ -1,0 +1,38 @@
+name: Push to luarocks.org
+
+on:
+  push:
+    tags:
+      - '*'
+  release:
+    types:
+      - created
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  luarocks-upload:
+    runs-on: ubuntu-24.04
+    steps:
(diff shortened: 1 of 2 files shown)


teto:
Installing via luarocks has the advantage of autoinstalling lua dependencies so the project could remove/add dependencies without the user doing anything. It's also leveraged to autogenerate the nix package.

One example where this could be useful is for instance if we were to rewrite the command using 
https://github.com/ColinKennedy/mega.cmdparse (which generates native autocompletion), this could be transparent for rocks.nvim user.

PR #2992, opened 2026-03-16
fix(acp): default tool call status to pending per ACP spec

lavignes:
The [ACP spec](https://agentclientprotocol.com/protocol/tool-calls#param-status) says that when undefined, the status should be treated as `pending`.

I'm currently using Kiro as my ACP provider and it doesn't send `pending` status.


lavignes:
Bumping this. This is still relevant.

PR #3015, opened 2026-04-08
:AvanteSwitchProvider now lists acp providers too

r001:
:AvanteSwitchProvider fails to list ACP providers, that is fixed

### DIFF
--- a/plugin/avante.lua
+++ b/plugin/avante.lua
@@ -125,7 +125,7 @@ cmd("Refresh", function() require("avante.api").refresh() end, { desc = "avante:
 cmd("Focus", function() require("avante.api").focus() end, { desc = "avante: switch focus windows" })
 cmd("SwitchProvider", function(_opts)
   local providers = vim.tbl_keys(Config.providers)
-  vim.tbl_extend("force", providers, Config.acp_providers)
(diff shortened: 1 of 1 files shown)


teto:
I noticed it as well (sry for dup in https://github.com/yetone/avante.nvim/pull/3032/changes). Would you mind adding a test/possibly refactoring the function so it can be tested ?

sirmacik:
Confirming this fix works as expected. Hit the same bug after upgrading past #2935: ACP providers (`claude-code`, `gemini-cli`, `codex`) silently disappeared from the `:AvanteSwitchProvider` picker even with proper `acp_providers` configuration.

Root cause is straightforward: `vim.tbl_extend("force", providers, Config.acp_providers)` returns a new table but the result is discarded — it doesn't mutate `providers` in place. `vim.list_extend` is the correct API for appending list-style values, and pairing it with `vim.tbl_keys()` matches the original pre-#2935 behavior.

Tested locally on neovim 0.12.2 with all three ACP providers; all now appear correctly in the picker. +1 to merge.

teto:
if anyone wants to follow up with a test, would be cool

PR #3024, opened 2026-04-24
feat: hide statusline in avante's windows

teto:
it takes too much space to display 0 information of interest.

This can be overriden by restoring the value of statusline in after/ftplugin/AvanteInput.lua


teto:
`laststatus` is already correctly adised in READMe

PR #3032, opened 2026-04-26
fix acp listing

teto:
SwitchInputProvider doesn't list ACP providers.

Marking this as draft because I would like to add a test. Probably after 
https://github.com/yetone/avante.nvim/pull/3031  gets merged


teto:
original issue fixed in another PR. This is just to increase compat with emmylua_ls

PR #3043, opened 2026-05-02
fix: revert vim.text.diff to vim.diff for Neovim 0.10 compatibility

propilideno:
## Summary
Revert `vim.text.diff` back to `vim.diff` to restore Neovim 0.10/0.11 compatibility.

## Problem
Commit 0764277 introduced `vim.text.diff` across 5 files. This API is only available in Neovim 0.12+ (nightly/unstable) and does not exist in stable 0.10 or 0.11.

Users on stable Neovim 0.10/0.11 get:
```
attempt to call field 'diff' (a nil value)
```

## Root Cause
The rename was introduced in Neovim [f3a54e7](https://github.com/neovim/neovim/commit/f3a54e7ccfcfa2752eccd7585e2c4b94a4c8c433) for the upcoming 0.12 release. In that commit, `vim.diff` is kept as a deprecated alias (`---@deprecated`, removal scheduled for Nvim 1.0), so using `vim.diff` is safe across all versions including 0.12+.

## Solution
Use `vim.diff` which works on stable 0.10, 0.11, and 0.12+ nightlies.

Fixes #3042

PR #3045, opened 2026-05-03
fix: revert vim.text.diff to vim.diff for Neovim 0.10 compatibility

teto:
see https://github.com/yetone/avante.nvim/pull/3043

### DIFF
--- a/.github/workflows/lua.yaml
+++ b/.github/workflows/lua.yaml
@@ -69,7 +69,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


.github/workflows/lua.yaml
```diff
@@ -69,7 +69,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        nvim_version: [ stable ]
+        nvim_version: [ v0.11.0 ]
```
teto, line 72: arf this should have been 0.10 here

PR #3046, opened 2026-05-03
refactor: replace Utils.join_paths with vim.fs.joinpath

teto:
available since neovim 0.10


teto:
Ideally I would have first replaced Utils.join_paths with vim.fs.joinpath, check that it passes the spec but I started off the implementation with codex and it managed to make it work. In some ways it is better than the previous implementation of join_paths that did a lof of different things. The new version is more focused and explicity about when it needs to get the absolute paths and so on.

I was a bit worried it would break avante but after a few local tests. It seems to work fine so I am thinking of merging,
Maybe I've missed some usecase so feedback/testing is welcome.

PR #3047, opened 2026-05-03
chore: deprecate require'avante_lib'.load()

teto:
mark it as deprecated to avoid breaking user configs but it is now a noop (does nothing). Updated README to reflect changes in #2324

reenactment of https://github.com/yetone/avante.nvim/pull/2354 that I intend to finish


teto:
will wait for someone on darwin to check this works. should be good though, there can be followup as well

PR #3053, opened 2026-05-05
fix: map <esc> to close in normal mode instead of insert mode

### LINKED ISSUE #2948: feature: Telescope normal mode
### Feature request

Using telescope as the file_selector provider seems to hard-code the behavior that pressing Escape once closes the picker.  It would be great if we could have it enter normal mode first, and then close the picker on second press,.

### Motivation

Entering normal mode is the default Telescope behavior, and enables navigating the list with `j` and `k`, the default vim navigation keys. By staying in insert mode, we are constrained to using arrow keys.

### Other

_No response_

ioannis-chrysostomakis:
Fixes https://github.com/yetone/avante.nvim/issues/2948

Using telescope as the file_selector provider makes pressing Escape once close the picker. This is counter-intuitive based on the default telescope behavior.

Normally you would enter normal mode, enabling navigating the list with j and k - the default vim navigation keys. Then you would press Escape again to close it.

### DIFF
--- a/lua/avante/ui/selector/providers/telescope.lua
+++ b/lua/avante/ui/selector/providers/telescope.lua
@@ -60,7 +60,7 @@ function M.show(selector)
           end,
         }),
         attach_mappings = function(prompt_bufnr, map)
-          map("i", "<esc>", require("telescope.actions").close)
+          map("n", "<esc>", require("telescope.actions").close)
           map("i", "<c-del>", function()
             local picker = action_state.get_current_picker(prompt_bufnr)
 


ioannis-chrysostomakis:
comment

teto:
I can see why the current default was chosen. Changing it will likely upset as users. You  can override the defaults via `provider_opts` though. If you wanna keep "<c-del>" you might need to copy a bit of code but if we can expose the function to make it easier, feel free to propose

PR #3068, opened 2026-05-15
fix: Handle Unicode surrogate pairs

### LINKED ISSUE #3067: bug: Invalid characters returned from unescapeJsonString() when input contains Unicode surrogate pairs
### Describe the bug

When a message contains codepoints outside the basic multilingual plane, it's not decoded correctly in `unescapeJsonString()`.

### To reproduce

As an example, run a session which asks for the Unicode codepoint for 🖱:

https://www.compart.com/en/unicode/U+1F5B1

If the JSON response is `\u`-encoded (rather than plain UTF-8), you should receive `\uD83D\uDDB1`, which should be decoded as a single codepoint. The current implementation handles each of those `\u` escapes separately.

(seen on https://cerebras.ai responses, but would guess that it happens for other providers too)

### Expected behavior

Surrogate pairs are decoded correctly.

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  opts = {
    -- add any opts here
  },
  build = "make",
  dependencies = {
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}
```

### Environment

neovim version: NVIM v0.12.2
Build type: RelWithDebInfo
LuaJIT 2.1.1774638290
Linux

### Repro

```lua
-- this would need API keys?
```

tm604:
Surrogate pairs consist of two `\u`-escaped values which should be combined before decoding.

Attempted to keep the comments in the same language as the rest of the file, but they're likely to need review/editing.

This is to fix #3067.

### DIFF
--- a/lua/avante/libs/jsonparser.lua
+++ b/lua/avante/libs/jsonparser.lua
@@ -110,6 +110,34 @@ local function unescapeJsonString(str)
     end
   end)
 
+  result = result:gsub("\\u(%x%x%x%x)\\u(%x%x%x%x)", function(high_hex, low_hex)
+    local high_surrogate = tonumber(high_hex, 16)
+    local low_surrogate = tonumber(low_hex, 16)
+
+    -- 检查否为代对
+    if
+      high_surrogate
+      and low_surrogate
+      and high_surrogate >= 0xD800
+      and high_surrogate <= 0xDBFF
+      and low_surrogate >= 0xDC00
+      and low_surrogate <= 0xDFFF
+    then
+      -- 计算实际码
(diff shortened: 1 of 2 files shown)


teto:
I beleive this will be better supported in neovim 0.13 but this can help untl then. Can you add a test to tests/libs/jsonparser_spec.lua ?

tm604:
Thanks - I've pushed a basic test and confirmed it fails as expected on `main` before this patch is applied.

Also, for consistency I thought similar logic might be needed in `utils/streaming_json_parser.lua`, but since `\u` escapes don't seem to be handled at all there I left it alone.

Issue #3065, opened 2026-05-12
get rid of setup function


teto:
having to call a setup function instead of setting a vim.g.avante variable is quite the pain, see  https://mrcjkb.dev/posts/2023-08-22-setup.html for more details.
I would love to get rid of the setup call or make it optional. The problematic bits might be in the transition though that doesn't seem hard.
Help welcome.

PR #3073, opened 2026-05-16
feat: support configuration via global vim.g.avante

teto:
Endgame is to remove setup but right now it's not doable because there is some processing needed to customize/provider setup mappings.

Goal is to refactor config such that we can configure exclusively via vim.g.avante see https://github.com/yetone/avante.nvim/issues/3065


teto:
I dont advise to use it yet but merging to reduce diff and help experiment with my branch

Issue #3049, opened 2026-05-04
feature: deterministic vimdoc (help needed)


teto:
### Feature request

there is no `:h avante` available which would be helpful to make the API more discoverable.
These days the doc lives fully in README.md. 

We could document the api via panvimdoc , mini.docs. Ideally we would sitll have some manually written doc at the  top-level.

Would be cool if someone could contribute to this

teto:
added a generated doc as placeholder in https://github.com/yetone/avante.nvim/pull/3051 but it's going to get outdated and it's not well though out either. More work is needed.

teto:
we've got the basics covered, though it can be still improved a lot lets close

PR #3083, opened 2026-05-20
deterministic doc

teto:
moved the generated vimdoc to within the lua modules themselves. It's still quite messy but we can iterate over this. I prefer to merge early since it's prone to merge conflicts.

use `make docgen` to generate doc.

Ideally we would have a github action to detect changes to the help, and auto push updates like for commits. For now I can update manually.

I hope we can trim down the README.md now that we have a help.


teto:
there is a lot of improvements to do, I hope we can document each provider in their own file. Once this is merged it gets easier for the community to contribue I hope.
referencing  https://github.com/yetone/avante.nvim/issues/3049 for tracking

PR #3086, opened 2026-05-22
fix: clean up duplicate entry in avante.txt

srodriguez91:
Removed duplicate `avante.Sidebar` entry which caused docs to fail on Lazy Update. This entry I'm proposing to remove duplicates line 792.


teto:
thanks for the contribution. I have submitted a better fix in the long run here https://github.com/yetone/avante.nvim/pull/3087.

PR #3094, opened 2026-05-24
feat: Cerebras Code provider

tm604:
Cerebras provides high-speed inference, with their Code plan offering zlm-4.7 currently.

https://www.cerebras.ai/blog/introducing-cerebras-code

This is basically OpenAI compatible, with the exception of the `reasoning_content` field, which they label as `reasoning` instead.

This implementation remaps those so that the existing logic in `openai.lua` can be used, and provides provider configuration for the endpoint and API key name.

### DIFF
--- a/lua/avante/providers/cerebras.lua
+++ b/lua/avante/providers/cerebras.lua
@@ -124,4 +124,3 @@ function M:parse_response_without_stream(data, chunk_content, opts)
 end
 
 return M
-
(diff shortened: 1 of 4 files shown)


tm604:
Posting this in case it's useful to others - main advantage of Cerebras over the existing providers is the speed (generation is typically >1k tokens/sec). Will fix up the failing type checks.

lua/avante/providers/cerebras.lua
```diff
@@ -0,0 +1,126 @@
+local OpenAI = require("avante.providers").openai
```
teto, line 1: could you also add emmylua doc ? we use https://github.com/mrcjkb/vimcats to generate the avante.txt doc. 
The section could be named `avante-providers-cerberas`.
None of the providers are currently documented, I can document one first if that helps.

PR #3100, opened 2026-05-26
chore: deprecate require'avante_lib'.load()" (#3096)"

teto:
This reverts commit 98547000c70fd6d8cd198883c526996c47030500.

I reverted in a hurry since I worried it might impact windows users. After checking out lua/neovim runtime path, I think it should work fine.
@RobsonMobarack would you mind checking this PR on top of your PR https://github.com/yetone/avante.nvim/pull/3099 .

You should be able to remove `require("avante_lib").load()` from your setup and still have avante working

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -114,11 +114,7 @@ jobs:
         shell: bash
         run: |
           mkdir -p results
-          if [ "${{ matrix.config.os_name }}" == "linux" ]; then
-            EXT="so"
-          else
-            EXT="dylib"
-          fi
+          EXT="so"
           cp target/release/libavante_templates.$EXT results/avante_templates.$EXT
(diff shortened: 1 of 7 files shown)


teto:
let's go. This should help making avante work out of the box.

xsy420:
Are you fixing release action? It cannot compress on macos for different EXT

teto:
damn the code for CI is a duplicate of the one used in `AvanteBuild`. For instance .github/workflows/release.yaml has this code 

```
          if [ "${{ matrix.config.os_name }}" == "linux" ]; then
            EXT="so"
          else
            EXT="dylib"
          fi
```

which already exists in the Makefile...

I think ideally the CI should call `nvim +"AvanteBuild source=true"` so the code executed on CI is the same as what the user would do. Are you interested in contributing the fix ?

PR #3104, opened 2026-05-28
removed useless tags

teto:
wrongly updated


teto:
I need to be able to run the same checks as CI. Marking as draft till then see https://github.com/yetone/avante.nvim/pull/3107

PR #3107, opened 2026-05-28
chore: address lua-language-server errors

teto:


teto:
de784a7b broke something the path towards dep appear as deps/deps. We shall use vim.pack instead.

PR #3111, opened 2026-05-29
chore: abstract vim.diff and vim.text.diff via Utils.diff

teto:


teto:
we now require 0.12 so not needed anymore

PR #3112, opened 2026-05-29
fix: enhance CI builds, dynamic fork artifact downloads, and Windows paths

RobsonMobarack:
This PR introduces several crucial fixes to improve the CI/CD pipeline, support fork builds, and fix native Windows bugs.

### CI & Releases (`release.yaml`)
- Fixed macOS cross-compilation builds by migrating to `macos-latest` (M1/ARM).
- Fixed `.dylib` extension mismatch for macOS binaries.
- Disabled Luarocks upload on forks to prevent Action failures when `LUAROCKS_API_KEY` is missing.
- Simplified the release action trigger.

### Build Scripts (`Build.ps1` & `build.sh`)
- Removed hardcoded `yetone/avante.nvim` repository owner. The build scripts now dynamically resolve the repository owner and name via `git config --get remote.origin.url`. This allows users running from forks to download their own release artifacts properly without falling back to source compilation.

### Core Path Fixes (`lua/avante/path.lua`)
- Fixed a bug where `mkdir` would fail on Windows due to mixed slashes (`/` and `\`) when creating the cache directory. Forced path normalization to forward slashes before calling `vim.fn.mkdir`.
- Dynamically added the `build/` directory to `package.cpath` before requiring `avante_templates`. This ensures Neovim can instantly load the DLLs/SOs even if the package manager (`lazy.nvim`) hasn't updated the search path yet in the current session.

Tested successfully on both Windows and Linux!

To make testing easier, you can also try replacing "yetone/avante.nvim" with "RobsonMobarack/avante.nvim" to get it directly from my repository.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,7 +2,7 @@ name: Release
 
 on:
   push:
-    tags: [v\d+\.\d+\.\d+]
+    tags: ['v*']
 
 permissions:
   contents: write
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -64,7 +64,7 @@ jobs:
             rust_target: x86_64-unknown-linux-gnu
             docker_platform: linux/amd64
             container: quay.io/pypa/manylinux2014_x86_64 # for glibc 2.17
-          - os: macos-13
+          - os: macos-latest
             os_name: darwin
             arch: x86_64
             rust_target: x86_64-apple-darwin
@@ -95,7 +95,7 @@ jobs:
       - name: Build all crates
         if: ${{ matrix.config.container == null }}
         run: |
-          cargo build --release --features ${{ matrix.feature }}
+          cargo build --release --features ${{ matrix.feature }} --target ${{ matrix.config.rust_target }}
 
       - name: Build all crates with glibc 2.17 # for glibc 2.17
         if: ${{ matrix.config.container != null }}
@@ -116,13 +116,15 @@ jobs:
           mkdir -p results
           if [ "${{ matrix.config.os_name }}" == "darwin" ]; then
             EXT="dylib"
+            TARGET_DIR="target/${{ matrix.config.rust_target }}/release"
           else
             EXT="so"
+            TARGET_DIR="target/release"
           fi
-          cp target/release/libavante_templates.$EXT results/avante_templates.$EXT
-          cp target/release/libavante_tokenizers.$EXT results/avante_tokenizers.$EXT
-          cp target/release/libavante_repo_map.$EXT results/avante_repo_map.$EXT
-          cp target/release/libavante_html2md.$EXT results/avante_html2md.$EXT
+          cp $TARGET_DIR/libavante_templates.$EXT results/avante_templates.$EXT
+          cp $TARGET_DIR/libavante_tokenizers.$EXT results/avante_tokenizers.$EXT
+          cp $TARGET_DIR/libavante_repo_map.$EXT results/avante_repo_map.$EXT
+          cp $TARGET_DIR/libavante_html2md.$EXT results/avante_html2md.$EXT
 
           cd results
           tar zcvf avante_lib-${{ matrix.config.os_name }}-${{ matrix.config.arch }}-${{ matrix.feature }}.tar.gz *.${EXT}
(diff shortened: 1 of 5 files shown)


teto:
sry our work clashed a bit. I was trying to fix the artifacts on the branch `release-v0.1`, I tagged v0.1.1 which possibly might not work on windows/mac without this PR yet. I cherry-picked  the v0.1 changes to main in https://github.com/yetone/avante.nvim/pull/3113 so you may need to resolve conflicts sry.
Once this is fixed we could tag a 0.1.2

.github/workflows/release.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
             rust_target: x86_64-unknown-linux-gnu
             docker_platform: linux/amd64
             container: quay.io/pypa/manylinux2014_x86_64 # for glibc 2.17
-          - os: macos-13
+          - os: macos-latest
```
teto: I updated it to macos-14 not knowing what is best. I guess it's similar, just one is pinned and not the other ?

.github/workflows/release.yaml
```diff
@@ -95,7 +95,7 @@ jobs:
       - name: Build all crates
         if: ${{ matrix.config.container == null }}
         run: |
-          cargo build --release --features ${{ matrix.feature }}
+          cargo build --release --features ${{ matrix.feature }} --target ${{ matrix.config.rust_target }}
```
teto, line 98: rust_target is needed or is it vibed ?

.github/workflows/release.yaml
```diff
@@ -114,11 +114,17 @@ jobs:
...
         run: |
           mkdir -p results
-          EXT="so"
-          cp target/release/libavante_templates.$EXT results/avante_templates.$EXT
-          cp target/release/libavante_tokenizers.$EXT results/avante_tokenizers.$EXT
-          cp target/release/libavante_repo_map.$EXT results/avante_repo_map.$EXT
-          cp target/release/libavante_html2md.$EXT results/avante_html2md.$EXT
+          if [ "${{ matrix.config.os_name }}" == "darwin" ]; then
+            EXT="dylib"
+            TARGET_DIR="target/${{ matrix.config.rust_target }}/release"
```
teto, line 119: same question

lua/avante/path.lua
```diff
@@ -451,6 +452,11 @@ P.repo_map = RepoMap
 ---@return AvanteTemplates|nil
 function P._init_templates_lib()
   if _templates_lib ~= nil then return _templates_lib end
+
+  local build_dir = vim.fn.fnamemodify(debug.getinfo(1, "S").source:sub(2), ":h:h:h") .. "/build"
+  local ext = vim.fn.has("win32") == 1 and "dll" or "so"
+  package.cpath = package.cpath .. ";" .. build_dir .. "/?." .. ext
```
teto, line 458: I dont think you need to modify cpath for dlls. neovim should be looking automatically for it if they exist in the lua folder

Build.ps1
```diff
@@ -9,8 +9,14 @@ $ErrorActionPreference = "Stop"
...
-$REPO_OWNER = "yetone"
-$REPO_NAME = "avante.nvim"
+$remoteUrl = git config --get remote.origin.url 2>$null
+if ($remoteUrl -match "github\.com[:/]([^/]+)/([^/]+?)(?:\.git)?$") {
+    $REPO_OWNER = $matches[1]
+    $REPO_NAME = $matches[2]
+} else {
+    $REPO_OWNER = "yetone"
+    $REPO_NAME = "avante.nvim"
+}
```
teto, line 19: is this how you were able to test releasing in your repo ?

build.sh
```diff
@@ -2,9 +2,17 @@
...
+remote_url=$(git config --get remote.${REPO_REMOTE}.url 2>/dev/null || true)
+if [[ "$remote_url" == *"github.com"* ]]; then
+  tmp="${remote_url#*github.com[:/]}"
+  tmp="${tmp%.git}"
+  REPO_OWNER="${tmp%/*}"
+  REPO_NAME="${tmp#*/}"
+else
+  REPO_OWNER="yetone"
+  REPO_NAME="avante.nvim"
+fi
```
teto, line 15: this looks a bit convoluted and doesn't seem used in CI . If the change is needed, there are simpler ways such as passing those as arguments or accepting a default from envrionment

Build.ps1
RobsonMobarack: Yes, I wanted to put it in my repository so I could test it on a Windows PC and a Linux system before making the pull request, but I came across a hard coded repository in the file. So, I added this condition in case any contributor wants to test it the same way I did.

build.sh
RobsonMobarack: Maybe so, but to be honest, I don't know another way to do it at the moment. Since it worked for me, I figured that if a contributor found a better way, they could refactor it later.

.github/workflows/release.yaml
RobsonMobarack: I think if darwin x86_64 users are able to use the version compiled by macOS 14, then everything is fine.

.github/workflows/release.yaml
RobsonMobarack: By using the --target option, we instruct Cargo not to compile for the architecture of the machine on which it is running, but rather for the specific architecture we wish to export.

PR #3115, opened 2026-05-30
Claude Code Pro/Max API Compliance

dq1Mango:
### Hi Everybody,
Not sure if anyone cares about this, but the Claude Code Pro/Max API kept giving me 429 rate limit errors, but I was able to get it to work by changing a couple things.

### All Models
- changing the oath token end point
- setting the user agent to axios (this actually matters for some reason)

### Opus Specific
- changed model name
- removed temperature from request body

I get that Anthropic doesn't really want you to use their API this way, so I could see why you might not want to continue to support this auth type. However, (at least for the next few months) I am very inclined to keep this working, and would be happy to help maintain it to the best of my abilities.


lua/avante/tokenizers.lua
```diff
@@ -55,7 +55,7 @@ function M.encode(prompt)
   local success, result = pcall(tokenizers.encode, prompt)
   -- Some output like terminal command output might not be utf-8 encoded, which will cause an error here
   if not success then
-    Utils.warn("Failed to encode prompt: " .. result)
+    Utils.warn("Failed to encode prompt: " .. prompt)
```
teto: we want to display error, not the prompt

lua/avante/config.lua
```diff
@@ -746,10 +746,11 @@ M._defaults = {
     ---@type AvanteSupportedProvider
     ["claude-opus"] = {
       __inherited_from = "claude",
-      model = "claude-3-opus-20240229",
+      model = "claude-opus-4-8",
       timeout = 30000, -- Timeout in milliseconds
       extra_request_body = {
-        temperature = 0.75,
+        -- anthropic doesn't let you set the temp anymore
+        --temperature = 0.75,
```
teto: remove it altogether then

lua/avante/config.lua
```diff
@@ -642,7 +642,7 @@ M._defaults = {
       timeout = 30000, -- Timeout in milliseconds
       context_window = 200000,
       extra_request_body = {
-        temperature = 0.75,
+        -- temperature = 0.75,
```
teto: remove

teto:
would be nice if someone knowledgeable about anthropic could explain if/why this is actually needed .

lua/avante/config.lua
```diff
@@ -752,11 +751,9 @@ M._defaults = {
     ---@type AvanteSupportedProvider
     ["claude-opus"] = {
       __inherited_from = "claude",
-      endpoint = "https://api.anthropic.com",
```
teto, line 773: the removal of "endpoint" triggers the CI error, you can add a comment to ignore the error or possibly use a placeholder one if it gets overriden later by the claude initialization.

lua/avante/providers/claude.lua
```diff
@@ -566,6 +566,7 @@ function M:parse_curl_args(prompt_opts)
   local headers = {
     ["Content-Type"] = "application/json",
     ["anthropic-version"] = "2023-06-01",
+    ["User-Agent"] = "axios/1.15.2",
```
teto, line 574: is lying about agent mandatory ?

dq1Mango: from my experience it is. It might be different when using the API key, but I don't have one so I can't test it.

RheaAyase:
Applying this to latest I'm getting invalid json error in response to me entering the authentication token:
```
HTTP 400: {"type":"error","error":{"type":"invalid_request_error","message":"Invalid JSON body"},"request_id":"req_011CcXaLTKEaLJsc7C6S96vr"}
```

I wonder if it's all the trailing `,` on the last element...

RheaAyase:
I tried cleaning up more of those `temperature`'s and the trailing `,` in the "extra request body" fields, didn't help.

Any way to see what it's sending (short of pcap?)

dq1Mango:
It has been a while since I have synced my fork, but I just tested the initial sign in process and on [my fork](https://github.com/dq1Mango/avante.nvim) it "works on my machine". 

My preferred tool for HTTP analysis is [mitmproxy](https://www.mitmproxy.org/), which as the name suggests, is a man-in-the-middle proxy. With this you are able to inspect, edit, block, replay and edit any HTTP requests.

For analyzing traffic from terminals you will also need to set a bunch of environment variables to persuade the terminal to use your proxy. This is what I ended up using; [set-certs.fish](https://github.com/dq1Mango/dotfiles/blob/main/scripts/set-certs.fish) 

I'm happy to explain more, but would encourage you to try your luck with my fork. Ill get around to syncing my fork eventually...

RheaAyase:
I applied all of your changes to mine manually. Something else might be off with other overrides. I'll keep looking

RheaAyase:
Your fork does indeed work for me, perhaps there was something in a newer commit that conflicted with those changes.

dq1Mango:
I just synced my fork and everything appears to be working. The main changes were just to the CI, so I'm not super surprised. That being said I'm not sure why you didn't have such luck, as it seems we did basically the same thing. Perhaps a diff will be revealing. That or there might be some mangled persistent state.

perrin4869:
This fixed the issue for me, the only weird thing that I noticed is that after authenticating, it prompted me for an `ANTHROPIC_API_KEY` which I obviously don't have, otherwise I wouldn't be using this method haha
But I'm not sure if this is just a bug with the native input provider which seems rather buggy

Edit: I tried `snacks.nvim` and it didn't prompt me for the anthropic api key

teto:
If I understood correctly, this fixes avante for pro users but breaks it for API users ?!

`AvanteAnthropicProvider` already has the setting `auth_type` to toggle different settings depending on the authentication type. Could you adjust the code to make use of that ?

Also it would be nice to expose this in the `:help`, you would need to add `lua/avante/providers/claude.lua` to the list here
https://github.com/yetone/avante.nvim/blob/90a0e77c63251c367935caeaa0070670522c960b/Makefile#L92 and regenerate the doc (in separate commits).

dq1Mango:
This might be a dumb question but how does this break API usage? I changed `token_endpoint` which (I think) is only referenced by OAuth authentication functions. Unfortunately I do not have an API key to test it.

teto:
> token_endpoint which (I think) is only referenced by OAuth authentication functions

ha it might not break then, would be nice to have confirmation by someone with claude API access though.

teto:
in absence of any feedback let's merge this since it impacts several users. I've updated the doc in https://github.com/yetone/avante.nvim/pull/3164 to mention the 2 authentication schemes. Could you rebase and squash into one commit please ?

Would be nice to have per-provider maintainers.  Would anyone like to be referenced for claude ? we could create a claude github team in avante's repo that gets notified on claude changes for instance.

teto:
I would like to make a release before changing some requirements for avante. Would be nice to have this in !

tcuthbert:
@teto are you able to merge this PR? this is blocking folks on claude-code from receiving updates since the workaround requires pinning to the head of this PR.

teto:
I have merge rights but this PR doesn't pass CI so either op fixes the conflict or someone can push another PR keeping current attribution. I dont use claude so I prefer someone who does to check wether that works.

teto:
"console.anthropic" is also used in the oauth url `    vim.uri_encode("https://console.anthropic.com/oauth/code/callback"),`, does it need to be changed here ?
We should aslo avoid repeating the use of   `["User-Agent"] = "axios/1.15.2"` . I noticed some headers as well mention the "2023" date. I wonder if those could go away (seems old)

PR #3118, opened 2026-05-31
fix(acp): avoid stale client state

### LINKED ISSUE #3102: bug: switching ACP providers reuses previous ACP client/session
### Describe the bug

After switching from `claude-code` ACP to `codex` ACP, the sidebar displays
`ACP: codex`, but the response still comes from Claude.

The process list also shows `claude-code-acp` running, while `codex-acp` is not started.

Failure screenshot:

Image

Process list:

Image

After applying a local workaround that resets the cached ACP client/session:

Image

### To reproduce

With both `claude-code` and `codex` configured in `acp_providers`:

1. Run `:AvanteSwitchProvider claude-code`.
2. Ask `who are you?`.
3. Run `:AvanteSwitchProvider codex`.
4. Ask `who are you?` again.

Notes from reading the code:

- `providers.refresh()` updates `Config.provider`
- `sidebar.lua` keeps passing `self.acp_client` and `self.chat_history.acp_session_id`
- `llm.lua` reuses `opts.acp_client` whenever it exists, without checking whether it was created for the current ACP provider

I tested a local workaround that resets the cached ACP client/session when the ACP provider changes. With that workaround, switching works as expected.

Would a PR that invalidates the cached ACP client/session when the ACP provider changes be acceptable?

### Expected behavior

After switching ACP providers, the next request should be handled by the newly selected ACP provider, not by the previously selected one.

### Installation method

Use lazy.nvim:
  ```lua
  {
    "yetone/avante.nvim",
    event = "VeryLazy",
    lazy = false,
    version = false,
    build = "make",
    opts = {
      provider = "claude-code",
      acp_providers = {
        ["claude-code"] = {
          command = "npx",
          args = { "-y", "claude-code-acp" },
        },
        ["codex"] = {
          command = "npx",
          args = { "-y", "@agentclientprotocol/codex-acp" },
        },
      },
    },
    dependencies = {
      "nvim-lua/plenary.nvim",
      "MunifTanjim/nui.nvim",
    },
  }
  ```

### Environment

 neovim version: v0.12.2
distribution (if any): Nix
platform: macOS
avante.nvim: v0.1.0

### Repro

```lua
```

p1n9d3v:
Fixes stale ACP state after switching providers or clearing chat history.

Previously, the sidebar could keep reusing `self.acp_client` and `chat_history.acp_session_id` after the selected ACP provider changed. This could make the next request go through the previously selected ACP provider, and could also show stale model/mode values from that provider in the chat history header.

## This fix

1. Uses the existing ACP command invocation (`command` + `args`) to detect stale ACP clients.
2. Resets the cached ACP client/session when the current ACP provider changes.
3. Ignores model/mode config options from a stale ACP client.
4. Clears the stored ACP session id when chat history is cleared.

Fixes #3102

## Test

- `nix develop --command make luastylecheck`
- `nix develop --command make lua-typecheck`
- `nix develop --command make luatest`

## Manual test

- Switched ACP providers between `claude-code` and `codex`.
- Verified each request was handled by the selected ACP provider.
- Switched back from `codex` to `claude-code` and verified stale Codex model/mode values were not shown for Claude Code.
- Ran `/clear` and confirmed `chat_history.acp_session_id` became `nil`.
- Confirmed the next request did not reuse the cleared ACP context.

## Results

image

### DIFF
--- a/lua/avante/libs/acp_client.lua
+++ b/lua/avante/libs/acp_client.lua
@@ -253,7 +253,7 @@ local LOG_SEPARATOR = string.rep("=", 80) .. "\n"
 ---@class ACPConfig
 ---@field transport_type "stdio" | "websocket" | "tcp"
 ---@field command? string Command to spawn agent (for stdio)
----@field args? string[] Arguments for agent command
+---@field args string[] Arguments for agent command
 ---@field env? table Environment variables
 ---@field host? string Host for tcp/websocket
 ---@field port? number Port for tcp/websocket
@@ -394,7 +394,7 @@ function ACPClient:_create_stdio_transport()
       error("Failed to create pipes for ACP agent")
     end
 
-    local args = vim.deepcopy(self.config.args or {})
+    local args = vim.deepcopy(self.config.args)
     local env = self.config.env
 
     -- Start with system environment and override with config env
--- a/lua/avante/llm.lua
+++ b/lua/avante/llm.lua
@@ -981,6 +981,18 @@ function M._stream_acp(opts)
   end
   local acp_client = opts.acp_client
   local session_id = opts.acp_session_id
+  local acp_client_config = acp_client and acp_client.config
+  local is_same_acp_command_invocation = acp_client_config
+    and acp_client_config.command == acp_provider.command
+    and vim.deep_equal(acp_client_config.args, acp_provider.args)
+  if acp_client and not is_same_acp_command_invocation then
+    Utils.debug("reset stale ACP client", acp_client_config and acp_client_config.command, acp_provider.command)
+    pcall(function() acp_client:stop() end)
+    acp_client = nil
+    session_id = nil
+    opts.acp_client = nil
+    opts.acp_session_id = nil
+  end
   if not acp_client then
     local acp_config = vim.tbl_deep_extend("force", acp_provider, {
       ---@type ACPHandlers
--- a/lua/avante/sidebar.lua
+++ b/lua/avante/sidebar.lua
@@ -147,6 +147,17 @@ local SIDEBAR_CONTAINERS = {
 local Sidebar = {}
 Sidebar.__index = Sidebar
 
+---@param acp_client avante.acp.ACPClient | nil
+---@param provider avante.ProviderName
+---@return avante.acp.ConfigOption[]
+local function get_acp_config_options(acp_client, provider)
+  local acp_provider = Config.acp_providers[provider]
+  if not acp_provider or not acp_client or not acp_client.config_options then return {} end
+  if not acp_client.config or acp_client.config.command ~= acp_provider.command then return {} end
+  if not vim.deep_equal(acp_client.config.args, acp_provider.args) then return {} end
+  return acp_client.config_options
+end
+
 ---@class avante.CodeState
 ---@field winid integer
 ---@field bufnr integer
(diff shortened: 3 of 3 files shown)


lua/avante/libs/acp_client.lua
```diff
@@ -264,6 +264,7 @@ local LOG_SEPARATOR = string.rep("=", 80) .. "\n"
 ---@field auth_method? string Authentication method
 ---@field handlers? ACPHandlers
 ---@field on_state_change? fun(new_state: ACPConnectionState, old_state: ACPConnectionState)
+---@field _avante_provider_name? avante.ProviderName
```
teto: one can already use "command" as a differentiator, no need to add a field

lua/avante/llm.lua
```diff
@@ -981,8 +982,20 @@ function M._stream_acp(opts)
...
   local session_id = opts.acp_session_id
+  local acp_client_provider = acp_client and acp_client.config and acp_client.config._avante_provider_name
+  if acp_client and acp_client_provider ~= acp_provider_name then
+    Utils.debug("reset stale ACP client", acp_client_provider, acp_provider_name)
+    pcall(function() acp_client:stop() end)
+    acp_client = nil
+    session_id = nil
+    opts.acp_client = nil
+    opts.acp_session_id = nil
+    if opts.on_save_acp_client then opts.on_save_acp_client(nil) end
```
teto: having `on_save_acp_client` accept nil is quite annoying since now we have to deal with nil values. Why is this necessary ? the default does `opts.acp_client = nil` which is what is already done 2 lines up.

lua/avante/sidebar.lua
```diff
@@ -147,6 +147,15 @@ local SIDEBAR_CONTAINERS = {
 local Sidebar = {}
 Sidebar.__index = Sidebar
 
+---@param acp_client avante.acp.ACPClient | nil
+---@param provider avante.ProviderName
+---@return avante.acp.ConfigOption[]|nil
+local function get_acp_config_options(acp_client, provider)
+  if not acp_client or not acp_client.config_options then return nil end
+  if not acp_client.config or acp_client.config._avante_provider_name ~= provider then return nil end
+  return acp_client.config_options
```
teto, line 158: question: wouldn't it make the code more readable to return an empty list instead of nil ?

lua/avante/libs/acp_client.lua
p1n9d3v: Updated. I removed the `_avante_provider_name` field and now use the existing ACP command invocation to detect stale clients. I compare both `command` and `args` because wrapper commands such as `npx` can have the same command while launching different ACP servers.

lua/avante/llm.lua
p1n9d3v: Agreed. I removed the `nil` callback path and restored the callback annotations.

lua/avante/sidebar.lua
p1n9d3v: Yes, updated to return an empty list.

lua/avante/llm.lua
```diff
@@ -981,6 +981,18 @@ function M._stream_acp(opts)
   end
   local acp_client = opts.acp_client
   local session_id = opts.acp_session_id
+  local acp_client_config = acp_client and acp_client.config
+  local is_same_acp_command_invocation = acp_client_config
+    and acp_client_config.command == acp_provider.command
+    and vim.deep_equal(acp_client_config.args or {}, acp_provider.args or {})
```
teto: why use deep_eval ? `args` should be a plain list : 

`---@field args? string[] Arguments for agent command`

instead of `acp_client_config.args or {}`change the type of  args to 

`---@field args string[] Arguments for agent command`

I am trying to get rid of optional values where it makes sense to simplify the code (by avoiding the need to handle nil)

teto: NB: when you do this, you will need to regenerate the doc with `make docgen` (which requires installing vimcats if you dont use nix) to update the documentation and pass CI

p1n9d3v: Updated. I made `ACPConfig.args` non-optional and removed the `args or {}` fallbacks. I also ran `make docgen` but it did not produce doc changes.

teto: ha sry for that lua/avante/libs/acp_client.lua is not part of the doc yet https://github.com/yetone/avante.nvim/blob/dd0041fc80229952272abfe70c12d7e17e16d2e0/Makefile#L75 . I should add it

Issue #3114, opened 2026-05-30
bug: StackTrace on AvanteAsk


davidmeredith:
### Describe the bug

Mac, M1 Pro, Tahoe 26.5 (25F71)

I installed avante via LazyVim using the :LazyExtras to install the avante plugin, all went ok.  When i try to toggle AvanteAsk the sidebar opens but the following trace repeats when i toggle to/from avante ask. Seems to happen when the insert cursor enters the diaglog to ask a question. 

```
vim.schedule callback: ...th/.local/share/nvim/lazy/avante.nvim/lua/avante/llm.lua:283: Make sure to build avante (missing avante_templates)
stack traceback:
	[C]: in function 'error'
	...h/.local/share/nvim/lazy/avante.nvim/lua/avante/path.lua:293: in function 'get_templates_dir'
	...th/.local/share/nvim/lazy/avante.nvim/lua/avante/llm.lua:283: in function 'generate_prompts'
	...th/.local/share/nvim/lazy/avante.nvim/lua/avante/llm.lua:479: in function 'calculate_tokens'
	...local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:2760: in function 'get_generate_prompts_options'
	...local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:3028: in function 'initialize_token_count'
	...local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:2541: in function 'show_input_hint'
	...local/share/nvim/lazy/avante.nvim/lua/avante/sidebar.lua:3146: in function 'func'
	...al/share/nvim/lazy/avante.nvim/lua/avante/utils/init.lua:783: in function ''
	vim/_core/editor.lua: in function <vim/_core/editor.lua:0>
```

gcc is up to date: 
```
% gcc --version
Apple clang version 21.0.0 (clang-2100.1.1.101)
Target: arm64-apple-darwin25.5.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
% which brew
/opt/homebrew/bin/brew
```

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false, -- set this if you want to always pull the latest change
  opts = {
    -- add any opts here
  },
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  build = "make",
[...]

Issue #3110, opened 2026-05-29
bug: Build fails on macOS due to incorrect file extension (`.so` instead of `.dylib`)


sueno61:
### Problem

When building avante.nvim from source on macOS, the build fails because the `Makefile` uses `.so` as the file extension for shared libraries on all platforms, including macOS. On macOS, shared libraries should use the `.dylib` extension instead.

This causes the following error:

```
cp target/release/libavante_tokenizers.so lua/avante_tokenizers.so
cp: target/release/libavante_tokenizers.so: No such file or directory
make: *** [luajit-tokenizers] Error 1
```

Additionally, the pre-built binary download via `build.sh` also fails because the referenced version (e.g. `v0.1.0`) does not exist in GitHub Releases.

### Root Cause

In `Makefile`, the extension for shared libraries is set to `so` for macOS, but macOS expects `dylib`:

```makefile
# Current (incorrect for macOS)
EXT := so

# Should be for macOS (Darwin)
EXT := dylib
```

### Workaround

Manually fix the `Makefile` before building:

```bash
sed -i '' 's/EXT := so/EXT := dylib/' Makefile
LUA_VERSION=luajit make BUILD_FROM_SOURCE=true
for f in lua/libavante_*.dylib; do
  mv "$f" "${f%.dylib}.so"
done
```

### Expected Behavior

The `Makefile` should correctly detect macOS and set `EXT := dylib` so that the build succeeds without manual intervention.

```makefile
ifeq ($(UNAME), Darwin)
  EXT := dylib
else
  EXT := so
endif
```

### Additional Notes

- The `TARGET_LIBRARY=templates` target also needs to be built separately, as it is not always completed when the default `all` target fails midway.
- It would be helpful if the `Makefile` or `build.sh` handled the macOS case more robustly.

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false, -- set this if you want to always pull the latest change
  opts = {
    -- add any opts here
  },
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  build = "make",
  -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
  dependencies = {
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}
```

### Environment
- OS: macOS (Apple Silicon / aarch64)
[...]

Issue #3105, opened 2026-05-28
Bug: cannot build on latest commit


litoj:
Failed (1)
    ○ avante.nvim   a 
        LUA_VERSION=luajit bash ./build.sh
        Local build is out of date . Downloading latest v0.1.0.
        make: *** [Makefile:47: luajit] Error 1

teto:
https://github.com/yetone/avante.nvim/pull/3100 changed the paths which might have broken some stuff (see last comment).
Help welcome.

cybaol:
So, fix this ? @teto

teto:
planned ;)

teto:
so apparently I had the wrong process for release, I created it from github's UI whereas I should have:
1. tagged a commit
2. wait for CI to publish release as draft 
3. "published" the release draft for everyone to see.

I had to remove the github release otherwise I could not publish the generated draft.
The manually created release did not contain the artifacts while the generated one had, hence the issues.

There is still an issue for macos with dylib vs so https://github.com/yetone/avante.nvim/issues/3110

ownself:
Just pull the latest commit, then try to build, but still failed T_T (Arch Linux)

avante.nvim ⌘ AvanteChat ⌘ AvanteToggle ⌘ AvanteClear ⌘ AvanteRefresh ⌘ AvanteAsk 
        LUA_VERSION=luajit bash ./build.sh
        Local build is out of date . Downloading latest release-v0.1.
        + test_command gh
        + command -v gh
        ++ curl -s https://api.github.com/repos/yetone/avante.nvim/releases/tags/release-v0.1
        ++ grep browser_download_url
        ++ cut -d '"' -f 4
        ++ grep avante_lib-linux-x86_64-luajit
        + ARTIFACT_URL=
        make: *** [Makefile:47: luajit] Error 1

teto:
there is no tag `release-v0.1` anymore, it was a temporary fluke, try bumping version again

cybaol:
`release-v0.1` and `release-v0.1.0` still exists.

teto:
fixed

cybaol:
``` 
 Installed (1)
    ● avante.nvim 42.56ms  VeryLazy
        dir    /home/kino/.local/share/nvim/lazy/avante.nvim
        url    https://github.com/yetone/avante.nvim
        branch main
        commit 20470a7
        readme README.md
        help   |avante-nvim|
        event   VeryLazy 

[...]

PR #3119, opened 2026-06-01
fix: enhance CI builds, dynamic fork testing, and fix Windows paths

RobsonMobarack:
## 🚀 Summary
This Pull Request enhances the CI/CD pipeline, improves local build scripts to support dynamic fork artifacts, and fixes native Windows path bugs. It replaces PR #3112, incorporating the upstream changes introduced in #3113 and fully resolving the merge conflicts.

### 1. CI & Release Pipeline (`release.yaml`)
- **Dynamic Target Directories:** Fixed the artifact extraction path. When `cargo build` is run with a `--target` flag (necessary for macOS cross-compilation), it outputs binaries to `target/${{ matrix.config.rust_target }}/release` rather than the default `target/release`. The copy commands now dynamically target the correct directory to prevent "file not found" errors during the artifact step.
- **Enforced `.so` extensions on macOS:** Even though Cargo natively outputs `.dylib` on macOS, Lua's `require()` inside Neovim expects `.so` universally. The action now automatically renames the extension to `.so` on Unix platforms to ensure seamless loading across the ecosystem.
- **Retained `macos-14`:** Maintained the runner update to `macos-14` as introduced in #3113.

### 2. Fork Testing & Artifact Downloads (`Build.ps1` & `build.sh`)
- **Dynamic Repository Detection:** Removed hardcoded `yetone/avante.nvim` references. The scripts now dynamically fetch the repository owner and name via `git config --get remote.origin.url`. This makes it possible for contributors to test releases and pre-built binaries directly in their forks without modifying the scripts manually.
- **Defensive Validations:** Added better error handling and fallback messages in `Build.ps1`. If a pre-built binary matching the target isn't found in the release, it now gracefully handles the error and instructs the user to build from source.
- **Local Build Fixes (`build.sh`):** Fixed the `cp` loop when building from source so it correctly maps Cargo native extensions (`.dylib`/`.dll`) to Lua standard extensions (`.so`/`.dll`).

### 3. Windows Compatibility Fixes
- **Robust Path Creation (`lua/avante/path.lua`):** Replaced the `cache_prompt_dir:mkdir({ parents = true })` call with a robust `vim.fn.mkdir(cache_dir_str, "p")` implementation. This fixes a critical bug where directory creation fails on Windows due to mixed slashes (`\` and `/`).
- **Safeguard `package.cpath` for Windows DLLs:** Explicitly appended the `build` directory to `package.cpath`. On some Windows environments, depending on the plugin manager and `&rtp` resolution, Neovim fails to reliably find the `.dll` modules. This ensures it loads flawlessly.

## 🤝 Conflict Resolution (from #3113)
This branch successfully resolves the merge conflicts introduced when the `v0.1` branch was cherry-picked into `main` (#3113). The CI definitions now successfully combine both the release workflow improvements introduced by `@teto` and the correct dynamic target paths for Rust cross-compilation from this branch.

## 💬 Addressing feedback from PR #3112
To clarify some questions raised by `@teto` in the previous PR:

- **"Is `rust_target` needed?"**
  Yes, it is crucial for cross-compiling on macOS (building `aarch64` and `x86_64` on the same runner). Using the target changes Cargo's output directory, which is why the `cp` commands in this PR were updated to look into the dynamically generated target subfolder instead of the root `/release` folder.
  
- **"Is this how you were able to test releasing in your repo?"**
  Exactly. The dynamic repo detection logic allows any contributor to push tags to their own fork and test the GitHub Actions (and artifact downloads) cleanly before submitting a PR upstream.
  
- **"I don't think you need to modify cpath for dlls."**
  In standard scenarios, Neovim does automatically look in the `lua/` folder. However, on certain Windows setups, it sometimes struggles to load the compiled Rust `.dll` automatically depending on how the plugin manager structures the paths. The explicit `package.cpath` modification is a safeguard to ensure it loads reliably on Windows. *(If the maintainers strongly prefer to omit it to keep the code cleaner, let me know and I can remove it, but I highly recommend keeping it for Windows stability).*

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,7 +2,7 @@ name: Release
 
 on:
   push:
-    tags: [v\d+\.\d+\.\d+]
+    tags: ['v*']
 
 permissions:
   contents: write
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -126,7 +126,9 @@ jobs:
           EXT="so"
 
           cp -v $TARGET_DIR/libavante_templates.$CARGO_EXT results/avante_templates.$EXT 
-
+          cp -v $TARGET_DIR/libavante_tokenizers.$CARGO_EXT results/avante_tokenizers.$EXT
+          cp -v $TARGET_DIR/libavante_repo_map.$CARGO_EXT results/avante_repo_map.$EXT
+          cp -v $TARGET_DIR/libavante_html2md.$CARGO_EXT results/avante_html2md.$EXT
           cd results
           tar zcvf avante_lib-${{ matrix.config.os_name }}-${{ matrix.config.arch }}-${{ matrix.feature }}.tar.gz *.${EXT}
 
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -125,7 +125,7 @@ jobs:
           # Rename the extension to .so on all Unix platforms
           EXT="so"
 
-          cp -v $TARGET_DIR/libavante_templates.$CARGO_EXT results/avante_templates.$EXT 
+          cp -v $TARGET_DIR/libavante_templates.$CARGO_EXT results/avante_templates.$EXT
           cp -v $TARGET_DIR/libavante_tokenizers.$CARGO_EXT results/avante_tokenizers.$EXT
           cp -v $TARGET_DIR/libavante_repo_map.$CARGO_EXT results/avante_repo_map.$EXT
           cp -v $TARGET_DIR/libavante_html2md.$CARGO_EXT results/avante_html2md.$EXT
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -64,7 +64,7 @@ jobs:
             rust_target: x86_64-unknown-linux-gnu
             docker_platform: linux/amd64
             container: quay.io/pypa/manylinux2014_x86_64 # for glibc 2.17
-          - os: macos-13
+          - os: macos-latest
             os_name: darwin
             arch: x86_64
             rust_target: x86_64-apple-darwin
@@ -95,7 +95,7 @@ jobs:
       - name: Build all crates
         if: ${{ matrix.config.container == null }}
         run: |
-          cargo build --release --features ${{ matrix.feature }}
+          cargo build --release --features ${{ matrix.feature }} --target ${{ matrix.config.rust_target }}
 
       - name: Build all crates with glibc 2.17 # for glibc 2.17
         if: ${{ matrix.config.container != null }}
@@ -116,13 +116,15 @@ jobs:
           mkdir -p results
           if [ "${{ matrix.config.os_name }}" == "darwin" ]; then
             EXT="dylib"
+            TARGET_DIR="target/${{ matrix.config.rust_target }}/release"
           else
             EXT="so"
+            TARGET_DIR="target/release"
           fi
-          cp target/release/libavante_templates.$EXT results/avante_templates.$EXT
-          cp target/release/libavante_tokenizers.$EXT results/avante_tokenizers.$EXT
-          cp target/release/libavante_repo_map.$EXT results/avante_repo_map.$EXT
-          cp target/release/libavante_html2md.$EXT results/avante_html2md.$EXT
+          cp $TARGET_DIR/libavante_templates.$EXT results/avante_templates.$EXT
+          cp $TARGET_DIR/libavante_tokenizers.$EXT results/avante_tokenizers.$EXT
+          cp $TARGET_DIR/libavante_repo_map.$EXT results/avante_repo_map.$EXT
+          cp $TARGET_DIR/libavante_html2md.$EXT results/avante_html2md.$EXT
 
           cd results
           tar zcvf avante_lib-${{ matrix.config.os_name }}-${{ matrix.config.arch }}-${{ matrix.feature }}.tar.gz *.${EXT}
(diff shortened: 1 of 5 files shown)


RobsonMobarack:
Hi everyone, just to provide some extra context on the impact of these changes.

By correcting the macOS extension mapping (forcing `.so` instead of `.dylib`) and improving the fallback validations in the local build scripts, this PR directly fixes:
- Fixes #3110

Consequently, because it fixes the underlying macOS compilation and artifact extraction pipeline, it should also resolve the build/template errors recently reported by macOS users in:
- Resolves #3105
- Resolves #3114

Let me know if there's anything else I should adjust to get this merged!

.github/workflows/release.yaml
```diff
@@ -2,8 +2,7 @@ name: Release
 
 on:
   push:
-    branches: ["release-v*"]
```
teto, line 5: can you keep the `branches: ["release-v*"]` ? it's helpful to check CI is able to create releases without actually tagging a release

.github/workflows/release.yaml
```diff
@@ -69,7 +68,7 @@ jobs:
             os_name: darwin
             arch: x86_64
             rust_target: x86_64-apple-darwin
-          - os: macos-latest
+          - os: macos-14
```
teto: I am fine with both. Maybe `macos-latest`

.github/workflows/release.yaml
```diff
@@ -96,8 +95,7 @@ jobs:
       - name: Build all crates
         if: ${{ matrix.config.container == null }}
         run: |
-          # TODO leverage AvanteBuild
```
teto, line 99: please keep the comment as it can help future contributors see it would be a welcome change. 
Right now we duplicate the build from source, ideally we should run `:AvanteBuild` here (but can be done later as priority is to fix release).

lua/avante/path.lua
```diff
@@ -451,6 +452,11 @@ P.repo_map = RepoMap
 ---@return AvanteTemplates|nil
 function P._init_templates_lib()
   if _templates_lib ~= nil then return _templates_lib end
+
+  local build_dir = vim.fn.fnamemodify(debug.getinfo(1, "S").source:sub(2), ":h:h:h") .. "/build"
+  local ext = vim.fn.has("win32") == 1 and "dll" or "so"
+  package.cpath = package.cpath .. ";" .. build_dir .. "/?." .. ext
```
teto: The goal of https://github.com/yetone/avante.nvim/pull/3100 is precisely to get rid of the "build" folder and put the libraries in `lua/` since neovim natively can load libraries over there. The proper fix is to have `Build.ps1` not install to `build/` but to the `lua/` folder. Then you can get rid of this code.

build.sh
```diff
@@ -2,9 +2,17 @@
 
 set -e
 
-REPO_OWNER="yetone"
-REPO_NAME="avante.nvim"
 REPO_REMOTE="${1:-origin}"
+remote_url=$(git config --get remote.${REPO_REMOTE}.url 2>/dev/null || true)
+if [[ "$remote_url" == *"github.com"* ]]; then
```
teto, line 7: doesn\'t this match yetone/avante.nvim's as well ?

RobsonMobarack:
Thanks for the review and the context, @teto! 

1. I will add `branches: ["release-v*"]` back so we can test the CI without tagging.
2. I'll update the macOS runner back to `macos-latest`.
3. I'll restore the `# TODO leverage AvanteBuild` comment to help guide future contributors.
4. Regarding the repository URL check in `Build.ps1`/`build.sh`: yes, it matches `yetone/avante.nvim` perfectly! It checks the remote origin dynamically, falling back to `yetone/avante.nvim` when needed, so it works flawlessly both in the main repo and in any forks. 
5. I completely agree with the long-term goal of PR #3100. Once the `build/` folder is gone, we can simplify these scripts even further!

I will apply these changes and push them shortly.

teto (approved):

teto:
I squashed them because 13 commits was too many for the final result and didn't want to make you go through another round. Thanks for fixing those issues, I'll try to enable the tests on windows/macos. I wonder if this will burn our github action credits.

build.sh
RobsonMobarack: Yes, exactly! This logic dynamically parses the `REPO_OWNER` and `REPO_NAME` from the Git remote URL. 

If the script is run in the official repository, it will correctly extract `yetone` and `avante.nvim`. If it's run in a fork, it automatically adapts to the fork's owner and repository name. 

This ensures full compatibility with the upstream repository while allowing contributors to seamlessly download and test pre-built artifacts directly from their own forks without having to change any hardcoded values.

RobsonMobarack:
Okay, thanks @teto! One last detail that didn't make it into the merge: the workflow now uses macos-latest for both the x86_64 and aarch64 targets. However, the best approach is to explicitly use macos-14 for the x86_64 build.

If possible, please consider making this change in a future commit. This will ensure that when GitHub Actions updates the runner version tagged as -latest, no build errors will occur for macOS users with x86_64 processors.

teto:
> However, the best approach is to explicitly use macos-14 for the x86_64 build.

why is that ? aren't they compatible ? macos-latest is macos15 i believe. Seems close enough ?

RobsonMobarack:
While macos-latest currently works for both architectures, pinning the x86_64 build explicitly to macos-14 is a safeguard for the future.

Here's why: GitHub's macos-latest now points to Apple Silicon (ARM64) runners. Building x86_64 binaries on these runners relies on cross-compilation and the continued presence of x86_64 support in the macOS SDK (as well as Rosetta 2 for any legacy build tools).

Apple is actively phasing out the Intel architecture. When macos-latest eventually updates to a future macOS version where Apple completely drops the x86_64 SDK or removes Rosetta 2, the x86_64 CI job will silently break.

By explicitly pinning the x86_64 target to macos-14, we ensure a stable, known-working environment for legacy Intel builds, isolating our CI from upstream deprecations in future macos-latest runner updates.

PR #3124, opened 2026-06-02
fix: provider API compatibility and tool schema robustness

gub-7:
## Summary

- **Claude 4+ temperature**: Strip the `temperature` parameter for Claude Opus 4, Sonnet 4, and Haiku 4 models, which use always-on extended thinking and reject `temperature` with a hard API error. Works for both the `claude` and `vertex_claude` providers. Also improves the temperature validation error message to hint users to remove it from `extra_request_body`.
- **Empty tool schemas as `{}`**: When a tool has no parameters, `input_schema.properties` must be a JSON object (`{}`) not an array (`[]`). Fixed for Claude, Vertex Claude, and OpenAI providers, and also for nested object properties in `utils/init.lua`.
- **Shell command fix**: Minor variable shadowing fix in `get_cmd_for_shell` (Linux/macOS path).

## Test plan
- [ ] Configure a Claude 4 model (e.g. `claude-opus-4.5`) and confirm no temperature error
- [ ] Add a tool with no parameters and confirm the API request succeeds

### DIFF
--- a/lua/avante/providers/claude.lua
+++ b/lua/avante/providers/claude.lua
@@ -236,8 +236,12 @@ function M:get_rate_limit_sleep_time(headers)
 end
 
 -- Models that do not support the temperature parameter.
--- Claude 4+ generation models (new naming: claude-{tier}-{version}) deprecate temperature entirely.
-local function is_temperature_unsupported(model)
+-- Claude 4+ generation models (new naming: claude-{tier}-{version}) use always-on extended thinking,
+-- which is incompatible with the temperature parameter (API hard error if sent).
+-- See: https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#important-considerations
+---@param model string|nil
+---@return boolean
+function M.is_temperature_unsupported(model)
   if not model then return false end
   -- Claude 4+ models use new naming convention: claude-{tier}-{major}[-{minor}]-{date}
   -- e.g. claude-opus-4-20250514, claude-sonnet-4-5-20250929, claude-sonnet-4-20250514
@@ -265,10 +269,6 @@ function M:transform_tool(tool, use_prefix)
   local input_schema_properties, required = Utils.llm_tool_param_fields_to_json_schema(tool.param.fields)
   local tool_name = tool.name
   if use_prefix then tool_name = OAUTH_TOOL_PREFIX .. tool.name end
-  -- Ensure properties is always a JSON object (not array) even when empty
-  if not input_schema_properties or vim.tbl_isempty(input_schema_properties) then
-    input_schema_properties = vim.empty_dict()
-  end
   local tool_def = {
     name = tool_name,
     description = tool.get_description and tool.get_description() or tool.description,
@@ -277,9 +277,9 @@ function M:transform_tool(tool, use_prefix)
       properties = input_schema_properties,
       required = required,
     },
+    -- OAuth/claude-code beta requires discriminated tool type
+    type = use_prefix and "custom" or nil,
   }
-  -- OAuth/claude-code beta requires discriminated tool type
-  if use_prefix then tool_def.type = "custom" end
   return tool_def
 end
 
@@ -689,7 +689,7 @@ function M:parse_curl_args(prompt_opts)
   }, request_body)
 
   -- Strip temperature for models that don't support it (e.g. Opus 4 with always-on thinking)
-  if is_temperature_unsupported(provider_conf.model) then body.temperature = nil end
+  if M.is_temperature_unsupported(provider_conf.model) then body.temperature = nil end
 
   return {
     url = Utils.url_join(provider_conf.endpoint, api_path),
@@ -718,7 +718,9 @@ function M.on_error(result)
     error_msg = "You don't have any credits or have exceeded your quota. Please check your plan and billing details."
   elseif error_type == "invalid_request_error" and error_msg:match("temperature") then
     error_msg = error_msg
-      .. " Try removing 'temperature' from your provider's extra_request_body config."
+      .. "\nPossible fixes:"
+      .. "\n  - Remove 'temperature' from your provider's extra_request_body config"
+      .. "\n  - Ensure the temperature value is between 0 and 1"
   end
 
   Utils.error(error_msg, { once = true, title = "Avante" })
--- a/lua/avante/providers/claude.lua
+++ b/lua/avante/providers/claude.lua
@@ -235,6 +235,19 @@ function M:get_rate_limit_sleep_time(headers)
   return Utils.datetime_diff(tostring(now), tostring(reset_dt))
 end
 
+-- Models that do not support the temperature parameter.
+-- Claude 4+ generation models (new naming: claude-{tier}-{version}) deprecate temperature entirely.
+local function is_temperature_unsupported(model)
+  if not model then return false end
+  -- Claude 4+ models use new naming convention: claude-{tier}-{major}[-{minor}]-{date}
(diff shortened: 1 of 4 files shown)


lua/avante/providers/claude.lua
```diff
@@ -261,6 +278,9 @@ function M:transform_tool(tool, use_prefix)
       required = required,
     },
   }
+  -- OAuth/claude-code beta requires discriminated tool type
+  if use_prefix then tool_def.type = "custom" end
```
teto: can we merge this into the previous `if use_prefix` for clarity ?

lua/avante/providers/claude.lua
```diff
@@ -252,7 +265,11 @@ function M:transform_tool(tool, use_prefix)
   local input_schema_properties, required = Utils.llm_tool_param_fields_to_json_schema(tool.param.fields)
   local tool_name = tool.name
   if use_prefix then tool_name = OAUTH_TOOL_PREFIX .. tool.name end
-  return {
+  -- Ensure properties is always a JSON object (not array) even when empty
+  if not input_schema_properties or vim.tbl_isempty(input_schema_properties) then
+    input_schema_properties = vim.empty_dict()
```
teto: do you think this is something that is universal/could be lifted into Utils.llm_tool_param_fields_to_json_schema ?

lua/avante/providers/claude.lua
```diff
@@ -235,6 +235,19 @@ function M:get_rate_limit_sleep_time(headers)
   return Utils.datetime_diff(tostring(now), tostring(reset_dt))
 end
 
+-- Models that do not support the temperature parameter.
+-- Claude 4+ generation models (new naming: claude-{tier}-{version}) deprecate temperature entirely.
+local function is_temperature_unsupported(model)
```
teto: could you link the documentation for this ? we might as well remove the temperature from the default claude provider

lua/avante/providers/claude.lua
```diff
@@ -692,7 +717,8 @@ function M.on_error(result)
   if error_type == "insufficient_quota" then
     error_msg = "You don't have any credits or have exceeded your quota. Please check your plan and billing details."
   elseif error_type == "invalid_request_error" and error_msg:match("temperature") then
-    error_msg = "Invalid temperature value. Please ensure it's between 0 and 1."
+    error_msg = error_msg
+      .. " Try removing 'temperature' from your provider's extra_request_body config."
```
teto: seems like previous warning can still be valid. Maybe keep both as hints eg.

```lua
 error_msg = error_msg..[[
Possible fixes:
- either Try removing 'temperature' from your provider's extra_request_body config
-  ensure temperature value is between 0 and 1
```

lua/avante/providers/claude.lua
```diff
@@ -930,6 +956,7 @@ function M.cleanup_claude()
   if M._file_watcher then
     ---@diagnostic disable-next-line: param-type-mismatch
     M._file_watcher:stop()
+    pcall(function() M._file_watcher:close() end)
```
teto, line 961: why the pcall ?

lua/avante/providers/openai.lua
```diff
@@ -24,6 +24,10 @@ function M:is_disable_stream() return false end
 ---@return AvanteOpenAITool
 function M:transform_tool(tool)
   local input_schema_properties, required = Utils.llm_tool_param_fields_to_json_schema(tool.param.fields)
+  -- Ensure properties is always a JSON object (not array) even when empty
+  if not input_schema_properties or vim.tbl_isempty(input_schema_properties) then
+    input_schema_properties = vim.empty_dict()
+  end
```
teto: (same comment as previously, if we lift into llm_tool_param_fields_to_json_schema, this becomes useless)

lua/avante/providers/vertex_claude.lua
```diff
@@ -64,6 +64,12 @@ function M:parse_curl_args(prompt_opts)
     tools = tools,
   })
 
+  -- Strip temperature for models that don't support it (e.g. Claude 4+ with always-on thinking)
+  local model = provider_conf.model or ""
+  if model:match("claude%-opus%-[4-9]") or model:match("claude%-sonnet%-[4-9]") or model:match("claude%-haiku%-[4-9]") then
```
teto: export the `is_temperature_unsupported` function rather than duplicate the code that will obviously go out of sync

lua/avante/utils/init.lua
```diff
@@ -101,8 +101,8 @@ local function get_cmd_for_shell(input_cmd, shell_cmd)
     cmd = { "powershell.exe", "-NoProfile", "-Command", input_cmd:gsub('"', "'") }
   else
     -- linux and macos we will just do sh -c
-    local effective_shell_cmd = shell_cmd or "sh -c"
-    for _, cmd_part in ipairs(vim.split(effective_shell_cmd, " ")) do
+    shell_cmd = shell_cmd or "sh -c"
+    for _, cmd_part in ipairs(vim.split(shell_cmd, " ")) do
```
teto: why the regression here ? overriding variables trigger modern LSP and can also make coding harder (long live functional languages)

lua/avante/providers/claude.lua
gub-7: Agree. Will fix.

lua/avante/providers/claude.lua
gub-7: Agree will fix.

lua/avante/providers/claude.lua
gub-7: https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#important-considerations

On removing `temperature` from the default provider config: I'd leave the runtime stripping in place regardless, because users can always set `extra_request_body = { temperature = ... }` themselves, and we need to guard against that for Claude 4 models. Whether we also remove it from our own default config is cosmetic at this point. If the default config currently includes a temperature value for Claude, removing it is harmless — happy to do that too.

lua/avante/providers/claude.lua
gub-7: Agree.  The "remove from config" hint was written for the Claude 4 case but the `error_msg:match("temperature")` branch also catches range errors on Claude 3. We should give both actionable hints.

lua/avante/providers/claude.lua
gub-7: This is defensive against libuv handle state.** In Neovim's libuv bindings, calling `close()` on a handle that is already closed or in an error state will throw. `stop()` is idempotent and safe to call unconditionally, but `close()` finalizes the handle and can legitimately fail (e.g. if something else already closed it, or if the underlying OS resource was already gone). Wrapping
just `close()` in `pcall` is a common cleanup pattern for this reason — we want cleanup to always proceed even if `close()` fails.

That said, for consistency we could also `pcall(M._file_watcher.stop, M._file_watcher)` — if you'd prefer that symmetry I'm happy to, though I think the asymmetry here is intentional and reasonable. Also worth noting: this code predates this PR, so not sure it's in scope to change here.

lua/avante/utils/init.lua
gub-7: You're right.  I removed the local keyword to eliminate the shadowing warning I was getting, but this introduced a parameter mutation instead. I revert this change.

lua/avante/providers/vertex_claude.lua
gub-7: Agree, will fix.

lua/avante/providers/openai.lua
gub-7: Agree, will remove.

PR #3127, opened 2026-06-02
feat: enhanced grep tool with ripgrep, context lines, and line numbers

gub-7:
## Summary

The grep tool now returns richer, more actionable results so the agent can read relevant code snippets directly from search output without a separate file-read step:

- **Prefer ripgrep** (`rg`) over `ag`/`grep`; `ack` removed as a fallback (rarely installed)
- **`context_lines` parameter** (default 3): number of lines above and below each match, equivalent to `grep -C` / `rg -C`
- **Line numbers** included in output (`-n` flag) for all search tools
- **Output capped**: `--max-count 50` for ripgrep; total output truncated at 50 000 bytes with an informative suffix
- **Empty result**: returns an empty JSON object instead of an empty string
- Updated tool description, parameter docs, and return-value description

## Test plan
- [ ] Run a grep search and confirm results include line numbers and context
- [ ] Search in a large file and confirm output is truncated at the limit
- [ ] Confirm fallback to `ag` / `grep` when `rg` is not installed

### DIFF
--- a/lua/avante/llm_tools/grep.lua
+++ b/lua/avante/llm_tools/grep.lua
@@ -3,12 +3,26 @@ local Utils = require("avante.utils")
 local Helpers = require("avante.llm_tools.helpers")
 local Base = require("avante.llm_tools.base")
 
+--- Maximum output size in bytes before results are truncated.
+--- Keeps context window usage bounded when searching large codebases.
+local MAX_GREP_OUTPUT_BYTES = 50000
+
+--- Grep tool: search for a pattern across the project using ripgrep (rg), falling
+--- back to ag or grep if rg is unavailable.
+---
+--- Results include file path, line number, and a configurable number of context
+--- lines around each match, so the LLM can understand matches without a separate
+--- file-read step.
+---
+--- Output is capped at MAX_GREP_OUTPUT_BYTES bytes and annotated with a truncation
+--- notice when the limit is hit.
 ---@class AvanteLLMTool
 local M = setmetatable({}, Base)
 
 M.name = "grep"
 
-M.description = "Search for a keyword in a directory using grep in current project scope"
+M.description =
+  "Search for a pattern in files using ripgrep, returning matching lines with surrounding context. Use this to find code, definitions, and usages across the project."
 
 ---@type AvanteLLMToolParam
 M.param = {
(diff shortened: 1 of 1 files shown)


lua/avante/llm_tools/grep.lua
```diff
@@ -100,10 +115,17 @@ function M.func(input, opts)
       table.insert(cmd, "--glob")
       table.insert(cmd, "!" .. input.exclude_pattern)
     end
+    -- Limit output to avoid overwhelming context
+    table.insert(cmd, "--max-count")
+    table.insert(cmd, "50")
     table.insert(cmd, input.query)
     table.insert(cmd, abs_path)
```
teto: I dunno how they compare implementation wise but `vim.list_extend` has probably more opportunity for optimization

lua/avante/llm_tools/grep.lua
```diff
@@ -138,21 +153,23 @@ function M.func(input, opts)
...
 
   Utils.debug("cmd", table.concat(cmd, " "))
   if on_log then on_log("Running command: " .. table.concat(cmd, " ")) end
-  local result = vim.system(cmd, { text = true }):wait().stdout or ""
-  local filepaths = vim.split(result, "\n")
+  local result = vim.system(cmd, { text = true }):wait()
+  local output = result.stdout or ""
+
+  -- Truncate if too large (avoid blowing up context)
+  local max_output_size = 50000
```
teto: make this into a constant at the top of the file so it's easier to tweak

lua/avante/llm_tools/grep.lua
teto, line 1: could you add some emmylua doc presenting the tool a bit, and that the output might get truncated ?

teto:
can you share about the motivations for this ? How does it help the LLM ? Seems like it could consume more tokens/context window.
This also removes support for "ack" so the commit message should start with `feat!` to mark the breaking change

gub-7:
> can you share about the motivations for this ? How does it help the LLM ? Seems like it could consume more tokens/context window. This also removes support for "ack" so the commit message should start with `feat!` to mark the breaking change

ripgrep executes faster, and the added instruction to strongly grep around (ripgrep around) for context rather than adding whole files saves tokens.  I do think it would be beneficial to take a pass at all of the tools and try to write them in a way that will take up fewer tokens on each message.

I will fix the commit message

lua/avante/llm_tools/grep.lua
gub-7: Agree for multi-element cases, will fix.

lua/avante/llm_tools/grep.lua
gub-7: Agree, will fix.

lua/avante/llm_tools/grep.lua
gub-7: Yes

teto:
Fix the test and I'll merge (after some local testing)

PR #3136, opened 2026-06-08
doc: fix syntax in README for Home Manager setup

rokn:


teto:
thanks ! feel free to improve the flake if you are a nix lover. Would be cool to expose avante as a flake package o/

PR #3137, opened 2026-06-09
fix(build): make make build work with build directory

echoface:
- 将 build 声明为 phony 目标，避免被 build 目录遮蔽
- 增强 release 产物下载失败时的回退逻辑
- 按平台处理 tar.gz 和 zip 构建产物

en
```
## Summary

    This fixes `make build` when the repository contains an existing `build/` directory.

    ## Changes

    - Mark `build` as a phony Make target and map it to the default build flow.
    - Make release artifact download more robust when GitHub API lookup fails.
    - Handle `tar.gz` and `zip` release archives according to platform.
    - Fall back to source build when artifact download fails.

    ## Test

    - `bash -n build.sh`
    - `make build`
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,10 @@ TARGET_LIBRARY ?= all
 RAG_SERVICE_VERSION ?= 0.0.11
 RAG_SERVICE_IMAGE := quay.io/yetoneful/avante-rag-service:$(RAG_SERVICE_VERSION)
 
+.PHONY: all build
 all: luajit
+
+build: all
 
 define make_definitions
 ifeq ($(BUILD_FROM_SOURCE),true)
--- a/build.sh
+++ b/build.sh
@@ -25,16 +25,19 @@ Linux*)
   PLATFORM="linux"
   CARGO_EXT="so"
   LIB_EXT="so"
+  ARCHIVE_EXT="tar.gz"
   ;;
 Darwin*)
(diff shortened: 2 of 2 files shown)


echoface:
@yetone  Could you please review this change when you have time? Thanks!

teto:
thanks for trying to fix the windows platform. 
I would like the CI to run the same command a user would so there is no breakage during the release wether you can download the release or not. I did it for macos and linux and wondered if you would mind trying on windows (I dont have access to windows) ?
Basically the idea would be to remove the cargo command:
https://github.com/yetone/avante.nvim/blob/90a0e77c63251c367935caeaa0070670522c960b/.github/actions/build/action.yaml#L59
and just enable windows in https://github.com/yetone/avante.nvim/blob/90a0e77c63251c367935caeaa0070670522c960b/.github/actions/build/action.yaml#L47

Regarding this PR,  

Can `Compress-Archive -Path $dllFiles -DestinationPath "avante_lib-${{ inputs.os_name }}-${{ inputs.arch }}-${{ matrix.feature }}.zip"` be changed to create a tar.gz archive like on other platforms ? this way we would have only one codepath.

teto:
the situation for linux should have been fixed on main and latest release. Windows support shall get fixed in https://github.com/avante-corp/avante.nvim/pull/3213 ty for the PR.

PR #3158, opened 2026-06-25
ci: build libraries as user would on windows as well

teto:
I previously reworked the CI to move the "build" part of the release into its own action. The goal is to test the build in PRs (just for linux by default, easy enough to enable tests for windows on a specifc PR).
The goal is to build the release with the same command a user would do it, ie., through AvanteBuild. Seems to work for darwin/linux.

I am trying to do this on windows but this seems a bit different.

### DIFF
--- a/.github/actions/build/action.yaml
+++ b/.github/actions/build/action.yaml
@@ -51,12 +51,22 @@ runs:
           ls -lRa lua
 
 
+      - uses: rhysd/action-setup-vim@v1
+        if: ${{ inputs.os_name == 'windows' }}
+        with:
+          neovim: true
+          version: v0.11.0
+
       - name: Build all crates (Windows)
         if: ${{ inputs.os_name == 'windows' }}
         shell: bash
         run: |
           # TODO run AvanteBuild instead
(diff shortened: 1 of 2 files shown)


teto:
@RobsonMobarack I would like to do a release soon to check that the release process works smoothly for all platforms. As mentioned in the first post, I made it easier to test build before release. The test CI builds only for linux but it's easy enough in doubt to uncomment some lines to build for darwin/windows as well.
I fixed the macos/linux usecase via nix, but windows requires a different codepath, for instance to find plenary. Would you be interested in becoming a maintainer? at least for windows since it seems to be what you are using ?
To be honest I lack the motivation/time to deal with windows-related matters, I just want to keep it working so any help pushing this PR forward is welcome.

RobsonMobarack:
Hi everyone, thanks for the invitation! I’m glad to accept it and look forward to contributing to the project and collaborating with you all.

teto:
@RobsonMobarack I've just invited you to a "windows" team. I am curious about your opinion on https://github.com/avante-corp/avante.nvim/issues/3202#issuecomment-5409730273

Issue #3021, opened 2026-04-21
Bug: gpt-5.4 API request failed with status 400, previous response not found


wiserfz:
### Describe the bug

Hello, first of all thank you for such amazing plugin.

There is question when I work with openai gpt-5.4 model and use response api, when I ask a question to it, it will list a todo list, and then it will failed as follow:

<img width="597" height="535" alt="Image" src="https://github.com/user-attachments/assets/ace1e69b-ad0f-4729-aef1-197fd87452c5" />

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
  event = "VeryLazy",
  lazy = false,
  version = false, -- set this if you want to always pull the latest change
  opts = {
    instructions_file = "avante.md",
    auto_suggestions_provider = "openai",
    provider = "openai",
    providers = {
      openai = {
        endpoint = base_url .. "/openai/v1",
        model = "gpt-5.4",
        use_response_api = true,
        extra_request_body = {
          reasoning_effort = "high",
          max_completion_tokens = 64000,
          background = true, -- Response API only: set to true to start a background task
        },
      },
    },
    behaviour = {
      auto_suggestions = false,
      allow_access_to_git_ignored_files = true,
    },
    hints = { enabled = true },
    session_recovery = {
      enabled = false,
    },
    -- system_prompt as function ensures LLM always has latest MCP server state
    -- This is evaluated for every message, even in existing chats
    system_prompt = function()
      local hub = mcp.get_hub_instance()
      return hub and hub:get_active_servers_prompt() or ""
    end,
    -- Using function prevents requiring mcphub before it's loaded
    custom_tools = function()
      return {
        require("mcphub.extensions.avante").mcp_tool(),
      }
    end,
  },
  -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
  build = "make",
  -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
  dependencies = {
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
  },
}
```

### Environment

neovim version: NVIM v0.12.1
avante commit: 8c84af0f09e5d9058362ac58401a8e7b047d826e

### Repro

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

require("lazy.minit").repro({
[...]

Issue #2805, opened 2025-10-30
bug: copilot provider ignores use_response_api


Pytness:
### Describe the bug

Commit b95e27b5a60d81814929764f9f6acf3768b98be3 introduced the bug:

lua/avante/provider/copilot.lua:357
```lua
  return {
    url = H.response_url(M.state.github_token.endpoints.api or provider_conf.endpoint),
    timeout = provider_conf.timeout,
    proxy = provider_conf.proxy,
    insecure = provider_conf.allow_insecure,
    headers = Utils.tbl_override(headers, self.extra_headers),
    body = vim.tbl_deep_extend("force", base_body, request_body),
  }
```

Note it returns `H.response_url` without checking `use_response_api`.

I will create a PR.

### Environment

`nvim --version`: NVIM v0.12.0-dev-69bddc0
`uname -a`: Linux xxx 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 GNU/Linux

Pytness:
Close as this was fixed on d45b62219323d1f9f7cf36a3808cd4e97166a6ce

PR #3160, opened 2026-06-29
fix(openai): only send pending tool outputs with previous_response_id

JoshJAL:
## Problem

When using the OpenAI **Response API** (`use_response_api = true`), multi-turn tool conversations break in a few related ways:

1. **`parse_curl_args` re-sends the entire tool-output history.** It scanned all `parsed_messages` for `function_call_output` items and sent *every* one alongside `previous_response_id`. But `previous_response_id` already represents the server-side state of the prior turn, so re-sending all historical outputs causes `400` errors (e.g. *"previous response not found"*, #3021) and duplicated outputs.

2. **`support_previous_response_id` was never checked in `parse_curl_args`.** The branch was gated only on `has_function_outputs and self.last_response_id`. Since `last_response_id` is populated on every response regardless of provider, a provider configured with `support_previous_response_id = false` (e.g. Copilot, #2805) would still incorrectly take the `previous_response_id` path.

3. **`parse_messages` dropped `function_call` items** whenever `support_previous_response_id` was truthy. That left the full-history fallback emitting `function_call_output` items with no paired `function_call`, which the API rejects.

## Fix

- Always keep `function_call` items in the parsed Response API history, so the full-history fallback stays valid.
- Add a small `get_trailing_function_outputs` helper that extracts only the **trailing (pending)** outputs for the current turn.
- Take the `previous_response_id` path only when there are pending outputs, a stored `last_response_id`, **and** `support_previous_response_id ~= false`; otherwise send the full history.

Net effect: when continuing a turn statefully, only the pending tool outputs are sent with `previous_response_id`; when a provider can't use it, the full (and now valid) history is sent instead.

## Testing

Manually verified against the OpenAI Response API (`gpt-5.x`, `use_response_api = true`, `support_previous_response_id = true`) — multi-step agentic tool conversations that previously failed with `400` now complete. The `parse_messages` change is a structural simplification (one fewer branch) and the new helper is self-contained.

Refs #3021, #2805

### DIFF
--- a/lua/avante/providers/openai.lua
+++ b/lua/avante/providers/openai.lua
@@ -290,48 +290,41 @@ function M:parse_messages(opts)
       if #content > 0 then table.insert(messages, { role = self.role_map[msg.role], content = content }) end
       if not provider_conf.disable_tools and not use_ReAct_prompt then
         if #tool_calls > 0 then
-          -- Only skip tool_calls if using Response API with previous_response_id support
-          -- Copilot uses Response API format but doesn't support previous_response_id
-          local should_include_tool_calls = not use_response_api or not provider_conf.support_previous_response_id
+          if use_response_api then
+            -- Keep tool calls in the parsed history. parse_curl_args may send only
+            -- pending tool outputs with previous_response_id, but full-history
+            -- fallback still needs each output paired with its function_call.
+            for _, tool_call in ipairs(tool_calls) do
+              table.insert(messages, {
+                type = "function_call",
+                call_id = tool_call.id,
+                name = tool_call["function"].name,
+                arguments = tool_call["function"].arguments,
+              })
+            end
+          else
+            -- Chat Completions API format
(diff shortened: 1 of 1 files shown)


teto:
this seemed like an interesting PR why close ? also do you have understanding of the problem or was ist just fully AI generated ?

JoshJAL:
> this seemed like an interesting PR why close ? also do you have understanding of the problem or was ist just fully AI generated ?

Not fully AI generated. I understood the problem at a high level. 

When using the Responses API with previous_response_id, only the currently pending function_call_output items should be sent back, rather than replaying every historical tool output.

I closed it because I wasn’t confident enough in the implementation. The local issue I was trying to fix was real, and the direction may be right, but this code path also has to handle providers that use Response API formatting without supporting previous_response_id. I didn’t want to leave a PR open without feeling confident about it.

Issue #2749, opened 2025-10-07
bug: ACP process not stopped after exiting from vim


nisarg-arista:
### Describe the bug

Setup:
MACOS 15.7.1
I'm using `claude-code` ACP provider using the steps mentioned in the README file.

Steps:
1. Launch avante chat and post one query which will launch the ACP server and claude CLI backend.
2. Exit from neovim

Issue:
Both ACP server and claude CLI backend keeps running. Confirmed via `ps -ef | grep claude`

### To reproduce

Avante configuration:
```
    opts = {
      provider = "claude-code",
      auto_suggestions_provider = "claude-code",
      cursor_applying_provider = "claude-code",
      memory_summary_provider = "claude-code",
      behaviour = {
        auto_suggestions = true,
        enable_cursor_planning_mode = true,
      },
      selector = {
        provider = "snacks",
      },
      file_selector = {
        provider = "snacks",
      },
      mappings = {
        suggestion = {
          accept = "<C-a>",
          next = "<C-n>",
          prev = "<C-p>",
          dismiss = "<C-]>",
        },
      },
      acp_providers = {
        ["claude-code"] = {
          command = "npx",
          args = { "@zed-industries/claude-code-acp" },
          env = {
            NODE_NO_WARNINGS = "1",
            ANTHROPIC_API_KEY = os.getenv("ANTHROPIC_API_KEY"),
            CLAUDE_CODE_USE_VERTEX = os.getenv("CLAUDE_CODE_USE_VERTEX"),
            ANTHROPIC_VERTEX_PROJECT_ID = os.getenv("ANTHROPIC_VERTEX_PROJECT_ID"),
          },
        },
      },
    },
    -- if you want to build from source then do `make BUILD_FROM_SOURCE=true`
    build = "make",
    -- build = "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false" -- for windows
    dependencies = {
      "stevearc/dressing.nvim",
      "nvim-lua/plenary.nvim",
      "MunifTanjim/nui.nvim",
      {
        -- Make sure to set this up properly if you have lazy=true
        "MeanderingProgrammer/render-markdown.nvim",
        opts = {
          file_types = { "markdown", "Avante" },
        },
        ft = { "markdown", "Avante" },
      },
    },
  },
```

### Expected behavior

After exiting from neovim ACP process should be cleaned up.

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
[...]

PR #3169, opened 2026-07-13
fix(acp): terminate launcher process trees

### LINKED ISSUE #3168: bug: ACP launcher subprocesses survive Avante shutdown
### Describe the bug

Avante stops only the direct process returned by `uv.spawn`. If an ACP command is a launcher that creates another process, the launcher exits but its child can survive Neovim shutdown and become reparented to PID 1.

This is the remaining process-tree case behind the symptom reported in #2749. The `VimLeavePre` cleanup added by #2723 runs, but `transport.stop()` sends signals only to the direct process handle.

### Real-world reproduction

`@zed-industries/codex-acp@0.16.0` installs a JavaScript launcher. That launcher uses `spawnSync` to run the platform-native `codex-acp` binary. With the normal Avante configuration:

```lua
acp_providers = {
  codex = {
    command = "codex-acp",
    args = {},
  },
}
```

1. Start an Avante ACP request.
2. Stop it with `:AvanteStop`, or exit Neovim normally.
3. Inspect `codex-acp` processes.

The JavaScript launcher is stopped, but the native child can remain alive with PPID 1. Repeating the workflow accumulates native processes.

The same behavior can be reproduced without Codex by using an ACP launcher that starts `sleep 60` as a child and waits for it.

### Expected behavior

Stopping an ACP client or exiting Neovim should terminate the ACP command and the subprocess tree it created.

### Observed environment

- Avante `main`
- Neovim 0.11
- Linux ARM64
- `@zed-industries/codex-acp@0.16.0`

### Root cause

`_create_stdio_transport()` retains the direct `uv.spawn` process handle, and `transport.stop()` signals only that handle. Descendants are not part of the cleanup contract.

### Proposed fix

On POSIX, spawn each ACP command as a process-group leader and signal that process group during transport shutdown. Preserve the current direct-child behavior on Windows. Add a regression test with a launcher that creates a child process and assert that the child does not survive `transport.stop()`.

I have a tested draft implementation ready and will link the PR here.

dianedef:
## What changed

- Spawn POSIX stdio ACP commands as process-group leaders.
- Track the spawned PID alongside the libuv process handle.
- Signal the complete process group during `transport.stop()` on Linux/macOS.
- Preserve the existing direct-child shutdown behavior on Windows.
- Add a regression test whose ACP launcher creates a child process and assert that the child does not survive shutdown.

## Why

Avante currently stops only the direct process returned by `uv.spawn`. Package-manager and JavaScript launchers can create a native child process, so stopping Avante or exiting Neovim may kill the launcher while leaving its child alive and reparented to PID 1.

A concrete example is `@zed-industries/codex-acp@0.16.0`: its JavaScript launcher uses `spawnSync` to execute the platform-native `codex-acp` binary. Repeated Avante sessions can therefore accumulate native processes even though the existing `VimLeavePre` cleanup runs.

This extends the cleanup introduced for #2749 from direct-child cleanup to process-tree cleanup on POSIX.

Closes #3168.

## User impact

Closing an ACP client or exiting Neovim no longer leaves subprocesses created by ACP launchers running in the background on Linux/macOS.

## Validation

- `make luastylecheck`
- `./scripts/run-luatest.sh tests/libs/acp_client_spec.lua` — 13 successes, 0 failures
- Full Lua test run — 230 successes, 0 failures; the local environment emitted unrelated Docker-unavailable callback warnings
- Reproduced originally on Linux ARM64 with Neovim 0.11 and `@zed-industries/codex-acp@0.16.0`

### DIFF
--- a/lua/avante/libs/acp_client.lua
+++ b/lua/avante/libs/acp_client.lua
@@ -368,6 +368,8 @@ function ACPClient:_create_stdio_transport()
     stdout = nil,
     --- @type uv.uv_process_t|nil
     process = nil,
+    --- @type integer|nil
+    pid = nil,
   }
 
   --- @param transport_self avante.acp.ACPTransportInstance
@@ -412,6 +414,9 @@ function ACPClient:_create_stdio_transport()
     ---@diagnostic disable-next-line: missing-fields
     local handle, pid = uv.spawn(self.config.command, {
       args = args,
+      -- A separate process group lets us clean up ACP launchers and every
+      -- subprocess they create. Windows keeps the existing direct-child path.
+      detached = vim.fn.has("win32") == 0,
       env = final_env,
       stdio = { stdin, stdout, stderr },
     }, function(code, signal)
@@ -422,6 +427,7 @@ function ACPClient:_create_stdio_transport()
         transport_self.process:close()
         transport_self.process = nil
       end
+      transport_self.pid = nil
 
       -- Handle auto-reconnect
       if self.config.reconnect and self.reconnect_count < (self.config.max_reconnect_attempts or 3) then
@@ -440,6 +446,7 @@ function ACPClient:_create_stdio_transport()
     end
 
     transport_self.process = handle
+    transport_self.pid = pid
     transport_self.stdin = stdin
     transport_self.stdout = stdout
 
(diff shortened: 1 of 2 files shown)


teto:
> A concrete example is @zed-industries/codex-acp@0.16.0: its JavaScript launcher uses spawnSync to execute the

How do you install it/what do you mean by `JavaScript launcher` ? I install codex-acp via my package-manager and in a quicktest, I could not reproduce this, codex-acp gets killed on nvim exit. My first reaction would be that I dont think it's avante.nvim's responsibility to handle programs that don't kill their children like it should be the case. 
Maybe wrap the codex-acp call with systemd-run, I belive this might kill all children (not sure though) ? or rather just remove the javascript launcher altogether. We could document it's an issue in the help.

PR #3173, opened 2026-07-15
docs: OpenAI-compatible multi-model gateway example (DaoXE)

seven7763:
## Summary

Adds a short **Custom providers** example pointing Avante’s built-in `openai` provider at an OpenAI-compatible multi-model gateway via `endpoint`.

Example: [DaoXE](https://daoxe.com) (`https://daoxe.com/v1`).

- Docs only (`README.md`)
- No Lua/runtime changes

### Notes

- Account-scoped model IDs
- Chat Completions path
- DaoXE not available in mainland China
- Contributor is affiliated with DaoXE

## Test plan

- [ ] README renders
- [ ] Optional: configure `endpoint = \"https://daoxe.com/v1\"` with a real key + model ID

### DIFF
--- a/README.md
+++ b/README.md
@@ -1326,6 +1326,34 @@ Avante provides a set of default providers, but users can also create their own 
 
 For more information, see [Custom Providers](https://github.com/yetone/avante.nvim/wiki/Custom-providers)
 
+### Example: OpenAI-compatible multi-model gateway (DaoXE)
+
+Point the built-in `openai` provider at any OpenAI-compatible Chat Completions host by setting `endpoint`. Example with [DaoXE](https://daoxe.com):
+
+```lua
+require("avante").setup({
+  provider = "openai",
+  providers = {
+    openai = {
+      endpoint = "https://daoxe.com/v1",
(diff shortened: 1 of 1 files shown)


teto:
thanks. I believe the added example doesn't bring much value compared to the existing doc, especially if there is no standard name for the api key.
Adding it to the default providers https://github.com/yetone/avante.nvim/blob/460af39d32f59ad225f1f82acd9158d05dd7c8a7/lua/avante/config.lua#L760 would bring more value (with `__inherit_from = openai`).

> DaoXE not available in mainland China

May I inquire why ? is this a taiwanese company ?

> configure endpoint = \"https://daoxe.com/v1\" with a real key + model ID

What I really would like to do is run avante testsuite against a real provider. I have mistral free tier in mind but if there is a possibility to run daoxe in avante's CI, would be cool.

seven7763:
Thanks @teto for taking a look and for the honest feedback. I agree the example doesn't add enough over the existing docs, so I'll close this one. Thanks for the great project — happy to revisit down the road if there's ever interest.

PR #3174, opened 2026-07-15
feat(bedrock): support Bedrock API key (bearer token) auth

### LINKED ISSUE #2479: feature: using bedrock with API key
### Feature request

# Description

Currently it's only possible to use bedrock with a `access_key_id`/`secret_key combo`, however I received an `API key` from my employer, but due to lack of being AWS certified, I don't know the difference 😅 

After digging into the code a bit, it seems like we're currently calling `curl` with `--user <credentials>`.

To use the API key instead, it's possible to omit the `--user ...` argument and pass an auth header `"Authorization: Bearer $AWS_BEARER_TOKEN_BEDROCK"` instead.

I tested the above and it seems to work.

# Steps to solve

- allow passing an "API key" instead of `BEDROCK_KEYS`
- update `lua/avante/providers/bedrock.lua`: add a new if-branch for when an API key is provided
- update readme with the changes

### Motivation

_No response_

### Other

_No response_

tosih:
## Summary

Adds support for AWS Bedrock's short-form API key / bearer token authentication (via the `AWS_BEARER_TOKEN_BEDROCK` env var), in addition to the existing sigv4 `access_key_id`/`secret_access_key` flow. Also fixes a request-shape incompatibility with Claude 4+ models on Bedrock.

Closes #2479.

## Motivation

AWS now issues per-user Bedrock API keys that authenticate via a plain bearer token instead of sigv4 request signing. Users on machines without full AWS credentials configured (no `~/.aws/credentials`, no IAM role) can just export one env var and go.

The reporter of #2479 verified this works end-to-end: omit `--user` / `--aws-sigv4`, send `Authorization: Bearer $AWS_BEARER_TOKEN_BEDROCK`.

### `lua/avante/providers/bedrock.lua`
- `setup()` skips the sigv4 curl version check when a bearer token is available (useful on older curls like 8.5.x that don't support sigv4 but work fine for bearer tokens).
- `is_env_set()` now also returns true when `AWS_BEARER_TOKEN_BEDROCK` is set (falls back after the existing `BEDROCK_KEYS` / `aws_profile` checks).
- `parse_curl_args()`:
  - If `BEDROCK_KEYS` is unset, falls back to reading `AWS_BEARER_TOKEN_BEDROCK` directly.
  - If the resolved key **contains no commas**, it's treated as a bearer token: sends `Authorization: Bearer <token>` and skips both `--user` and `--aws-sigv4`. Requires `aws_region` in the provider config (errors clearly if missing).
  - The legacy comma-separated `access_key_id,secret_access_key,region[,session_token]` path is unchanged.
  - The `aws_profile` path is unchanged.

### `lua/avante/providers/bedrock/claude.lua`
- `build_bedrock_payload()` now strips `temperature` when the model matches `Claude.is_temperature_unsupported(...)` (Claude 4+ / opus-4 / sonnet-4 / haiku-4). These models have always-on extended thinking and reject `temperature` with a 400. This mirrors the existing behavior in `lua/avante/providers/vertex_claude.lua`.

### Docs
- `README.md` and `README_zh.md`: new paragraph describing the bearer-token mode.
- `lua/avante/init.lua` and `doc/avante.txt`: added the new env var to the module/help header.

## Backward compatibility

Fully backward compatible:
- Existing `BEDROCK_KEYS=access,secret,region[,session]` users: unchanged (comma path).
- Existing `aws_profile` users: unchanged.
- New: users can now `export AWS_BEARER_TOKEN_BEDROCK=...` (or set `BEDROCK_KEYS` to a comma-less token) and set `aws_region` in config.
- Temperature strip only fires for Claude 4+ models (regex `claude%-(opus|sonnet|haiku)%-[4-9]`). Older Sonnet 3.5 / 3.7 / Haiku 3.5 are unaffected.

## Testing

Tested end-to-end against a real AWS Bedrock account with a Bedrock API key:

1. **Direct curl smoke test** against `bedrock-runtime.us-east-1.amazonaws.com` using only `Authorization: Bearer $AWS_BEARER_TOKEN_BEDROCK` (no `--user`, no `--aws-sigv4`) → AWS accepts the auth and streams a 200 response from `amazon.nova-micro-v1:0`. Confirms the wire-level auth mode works with the token AWS issues.

2. **Provider unit trace** — headless Neovim harness calling `bedrock:parse_curl_args(...)` with `AWS_BEARER_TOKEN_BEDROCK` set (and `BEDROCK_KEYS` unset). Result:
   ```
   is_env_set() = true
   url          = https://bedrock-runtime.us-east-1.amazonaws.com/model/<encoded-model>/invoke-with-response-stream
   headers.Authorization = Bearer <token>
   headers.Content-Type  = application/json
   rawArgs               = (empty — no --user, no --aws-sigv4)
   ```

3. **Live end-to-end in Neovim** against `us.anthropic.claude-opus-4-7`: initial request returned `400 "temperature is deprecated for this model."`, which motivated the `bedrock/claude.lua` temperature-strip fix. After the fix, chat/edit flows work correctly with Claude Opus 4.7 on Bedrock.

4. Existing `tests/providers/bedrock_spec.lua` does not cover `parse_curl_args` or `build_bedrock_payload`, so no test updates were needed. Happy to add coverage if desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -947,6 +947,19 @@ Given its early stage, `avante.nvim` currently supports the following basic func
 >
 > Note: The aws_session_token is optional and only needed when using temporary AWS credentials
 >
+> Alternatively, you can authenticate with a [Bedrock API key](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html) (bearer token). Either export it as `AWS_BEARER_TOKEN_BEDROCK`, or set `BEDROCK_KEYS` to the token value (without any commas). In this mode the `aws_region` must be set in the bedrock provider config:
+>
+> ```sh
+> export AWS_BEARER_TOKEN_BEDROCK=your-bedrock-api-key
+> ```
+>
+> ```lua
+> bedrock = {
+>   model = "us.anthropic.claude-3-5-sonnet-20241022-v2:0",
+>   aws_region = "us-east-1",
+> },
+> ```
+>
 > Alternatively Bedrock tries to resolve AWS credentials using the [Default Credentials Provider Chain](https://docs.aws.amazon.com/cli/v1/userguide/cli-chap-authentication.html).
 > This means you can have credentials e.g. configured via the AWS CLI, stored in your ~/.aws/profile, use AWS SSO etc.
 > In this case `aws_region` and optionally `aws_profile` should be specified via the bedrock config, e.g.:
--- a/README.md
+++ b/README.md
@@ -947,19 +947,6 @@ Given its early stage, `avante.nvim` currently supports the following basic func
 >
 > Note: The aws_session_token is optional and only needed when using temporary AWS credentials
 >
-> Alternatively, you can authenticate with a [Bedrock API key](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html) (bearer token). Either export it as `AWS_BEARER_TOKEN_BEDROCK`, or set `BEDROCK_KEYS` to the token value (without any commas). In this mode the `aws_region` must be set in the bedrock provider config:
->
-> ```sh
-> export AWS_BEARER_TOKEN_BEDROCK=your-bedrock-api-key
-> ```
->
-> ```lua
-> bedrock = {
->   model = "us.anthropic.claude-3-5-sonnet-20241022-v2:0",
->   aws_region = "us-east-1",
-> },
-> ```
->
 > Alternatively Bedrock tries to resolve AWS credentials using the [Default Credentials Provider Chain](https://docs.aws.amazon.com/cli/v1/userguide/cli-chap-authentication.html).
 > This means you can have credentials e.g. configured via the AWS CLI, stored in your ~/.aws/profile, use AWS SSO etc.
 > In this case `aws_region` and optionally `aws_profile` should be specified via the bedrock config, e.g.:
--- a/README_zh.md
+++ b/README_zh.md
@@ -651,6 +651,12 @@ Avante.nvim 提供了多个可以与 blink.cmp 集成的补全项：
 > ```
 >
 > 注意：aws_session_token 是可选的，仅在使用临时 AWS 凭证时需要
+>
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -947,6 +947,19 @@ Given its early stage, `avante.nvim` currently supports the following basic func
 >
 > Note: The aws_session_token is optional and only needed when using temporary AWS credentials
 >
+> Alternatively, you can authenticate with a [Bedrock API key](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html) (bearer token). Either export it as `AWS_BEARER_TOKEN_BEDROCK`, or set `BEDROCK_KEYS` to the token value (without any commas). In this mode the `aws_region` must be set in the bedrock provider config:
```
teto: I am trying to reduce the README, I would rather see this in the luacats comments in lua/avante/init.lua in addition to the doc you've already added (and thanks for that).

tosih: Good call — done in 75de5ae. I've:

- Removed the new paragraph from `README.md` (and the mirror in `README_zh.md`).
- Expanded the luacats comments in `lua/avante/init.lua` to cover both auth modes (the classic `BEDROCK_KEYS` / credentials chain path and the new `AWS_BEARER_TOKEN_BEDROCK` bearer token path) with a full config example and a link to the AWS docs.
- Regenerated `doc/avante.txt` to match.

Should be much lighter on the README now.

lukeemhigh:
Hit this independently: Claude Sonnet 5 (`eu.anthropic.claude-sonnet-5`) via Bedrock returns a 400 (`temperature is deprecated for this model`) because `bedrock/claude.lua`'s `build_bedrock_payload()` doesn't strip `temperature` the way `claude.lua`/`vertex_claude.lua` already do. The fix in this PR (checking `Claude.is_temperature_unsupported(...)` before sending) resolves it exactly. Would be great to see this land — happy to help test if useful. In the meantime I'm tracking this branch as a temporary override.

PR #3176, opened 2026-07-25
fix: validate path in is_absolute.

bitsmanent:
Ensure path is valid in `is_absolute` to avoid crashing with `bad argument #1 to 'byte'`.

### DIFF
--- a/lua/avante/utils/path.lua
+++ b/lua/avante/utils/path.lua
@@ -16,6 +16,7 @@ function M.is_win() return IS_WIN end
 ---@param filepath                      string
 ---@return boolean
 function M.is_absolute(filepath)
+  if type(filepath) ~= "string" or filepath == "" then return false end
   if IS_WIN then return #filepath > 1 and string.byte(filepath, 2, 2) == BYTE_COLON end
   return string.byte(filepath, 1, 1) == BYTE_PATHSEP
 end


teto:
can you give an example of when this happened ? it's not supposed to be call with something else than a string so `type(filepath) ~= "string" ` should not be necessary and there is already a check on the length for windows but it could be generalized to take care of ""

bitsmanent:
I got that error while playing around with a code. Maybe session got corrupted at some point, in fact I cannot reproduce it in my normal flow (especially after :AvanteClear).

But we can still achieve the state.

Create an Avante session:

  - cd /tmp
  - nvim .
  - :AvanteAsk hello
  - :qa

Corrupt the session:

  - nvim ~/.local/state/nvim/avante/projects/__tmp/history/1.json
  - set messages[1].message.content[0].input.path to null
  - save an quit

Reproduce:

  - cd /tmp
  - nvim .
  - :AvanteAsk hello

See the `bad argument #1 to 'byte'` error.

Hope this helps.

cybaol:
I also have this problem。

cybaol:
Works well on my machine. Thanks.

cybaol:
> can you give an example of when this happened ? it's not supposed to be call with something else than a string so `type(filepath) ~= "string" ` should not be necessary and there is already a check on the length for windows but it could be generalized to take care of ""

A less intrusive way to fix this
```lua
function M.is_absolute(filepath)  
  if filepath == nil then return false end  -- add this line.
  if IS_WIN then return #filepath > 1 and string.byte(filepath, 2, 2) == BYTE_COLON end  
  return string.byte(filepath, 1, 1) == BYTE_PATHSEP  
end
```

Tested on my neovim and works well.

teto:
corrupted session is not a case we should take into account . I dont like the semantics of `is_absolute(nil)`. It's a pure leaf function, it should remain clear what it is about and it's not obvious that `is_absolute_path(not a path)` should return false. The error is better caught earlier, with enough context that we know what to make of it. 
My suggestion is to revert the change, see what tool triggers it and fix that tool.
For instance the view tool has in lua/avante/llm_tools/view.lua:
`if not input.path then return false, "path is required" end` 
but bash doesn't run any check for instance. I find it weird that we get nils there. What LLM do you use ? if you are on windows it could also be avante's internal failure but earlier in the calltree.

cybaol:
I use `deepseek` on `archlinux`, following is my config:
```lua
        deepseek = {
          __inherited_from = 'openai',
          api_key_name = 'AVANTE_DEEPSEEK_API_KEY',
          endpoint = 'https://api.deepseek.com',
          model = 'deepseek-v4-flash',
          extra_request_body = {
            thinking = { type = 'enabled' },
            reasoning_effort = 'high',
            stream = true,
          },
        },

```

bitsmanent:
I don't think `if filepath == nil` is a better alternative. It's less defensive against non-string types while still keeping the semantic inconsistency pointed out by @teto.

In fact, I completely agree with him: the root cause should be identified and fixed upstream, even though I haven't tracked it down yet.

I'm using Gemini via cli-proxy-api on Debian.

cybaol:
In `is_absolute`, add some logs like this:
```lua
function M.is_absolute(filepath)
  if filepath == nil then
    log.warn(debug.traceback("M.is_absolute received nil filepath", 2))
    return false
  end
  if IS_WIN then return #filepath > 1 and string.byte(filepath, 2, 2) == BYTE_COLON end
  return string.byte(filepath, 1, 1) == BYTE_PATHSEP
end
```
Tell agent to execute tools. The agent monitors  log output. and outputs all function call stack.
I think this is a solution to find root reason.

@bitsmanent

teto:
in https://github.com/avante-corp/avante.nvim/issues/3179, the error shows you the callstack

```
...al/share/nvim/lazy/avante.nvim/lua/avante/utils/path.lua:20: in function 'is_absolute_path'
...e/nvim/lazy/avante.nvim/lua/avante/llm_tools/helpers.lua:19: in function 'get_abs_path'
...hare/nvim/lazy/avante.nvim/lua/avante/llm_tools/init.lua:618: in function 'func'
...hare/nvim/lazy/avante.nvim/lua/avante/llm_tools/init.lua:1411: in function 'process_tool_use'
...in/.local/share/nvim/lazy/avante.nvim/lua/avante/llm.lua:1929: in function 'on_stop'
```

Notably the `...hare/nvim/lazy/avante.nvim/lua/avante/llm_tools/init.lua:618: in function 'func'`  shows what tool triggered the error. I beleive there are several though: all those that dont sanitize their input path. We could do a blank fir for all those tools but I am curious why the llm fails to answer properly, if it's a defect in our tool definition or something else.

cybaol:
Related PR in
https://github.com/vllm-project/vllm/pull/40959
https://github.com/vllm-project/vllm/pull/42879
https://github.com/sgl-project/sglang/pull/12123

cybaol:
In https://github.com/avante-corp/avante.nvim/blob/main/lua/avante/llm_tools/attempt_completion.lua#L93, seems missing `path` field. 
This is my call stack:
```vimscript
vim.schedule callback: ...al/share/nvim/lazy/avante.nvim/lua/avante/utils/path.lua:20: bad argument #1 to 'byte' (string expected, got nil)
stack traceback:
        [C]: in function 'byte'
        ...al/share/nvim/lazy/avante.nvim/lua/avante/utils/path.lua:20: in function 'is_absolute_path'
        ...e/nvim/lazy/avante.nvim/lua/avante/llm_tools/helpers.lua:19: in function 'get_abs_path'
        ...hare/nvim/lazy/avante.nvim/lua/avante/llm_tools/bash.lua:230: in function 'func'
        .../avante.nvim/lua/avante/llm_tools/attempt_completion.lua:93: in function 'func'
        ...hare/nvim/lazy/avante.nvim/lua/avante/llm_tools/init.lua:1250: in function 'process_tool_use'
        ...no/.local/share/nvim/lazy/avante.nvim/lua/avante/llm.lua:1946: in function 'on_stop'
        ...re/nvim/lazy/avante.nvim/lua/avante/providers/openai.lua:804: in function 'parse_response'
        ...no/.local/share/nvim/lazy/avante.nvim/lua/avante/llm.lua:552: in function 'parse_stream_data'
        ...no/.local/share/nvim/lazy/avante.nvim/lua/avante/llm.lua:656: in function <...no/.local/share/nvim/lazy/avante.nvim/lua/avante/llm.lua:652>
Press ENTER or type command to continue
```

Request body log:
```json
{
  "type": "function",
  "function": {
    "name": "bash",
    "parameters": {
      "additionalProperties": false,
      "type": "object",
      "required": ["path", "command"],
      "properties": {
        "path": {
          "description": "Relative path to the project directory, as cwd",
          "type": "string"
        },
        "command": {
          "description": "Command to run",
          "type": "string"
        }
      }
    }
  }
}

```

Response body log:
```json
{
  "name": "attempt_completion",
  "parameters": {
    "required": ["result"],
    "properties": {
      "result": { "type": "string" },
      "command": { "type": "string" } 
    }
  }
}
```

In bash schema,  `path` and `command`  are required.
`path` is relative path to the project directory, as cwd


So one  solution is:
```lua
require("avante.llm_tools.bash").func({ path = '.', command = input.command }, opts)
```

path = '.' is my gauss, not sure. or else value?

PR #3181, opened 2026-08-01
feat(web_search): add DuckDuckGo search provider

EvgeniyPeshkov:


teto:
would be nice to have duckduckgo as it's a major provider. This could be merged as is (with CI fixed, the doc/avante.txt must be generated via `make docgen`), but this shows to me that the `web_search` function gets too complex and also impacts the config too much. 
I believe a better design would be to have all these engines as different tools, we would set one by default and users could add/remove tools as they see fit. 
Let me know if that's something you would be interested in doing next.

teto:
I've just merged the change so your search engine can be its own tool in https://github.com/avante-corp/avante.nvim/pull/3205. Sry you'll have to update the PR but the change should be simple, just imitate what's there.

PR #3215, opened 2026-08-29
refactor!: replace plenary.curl with vim.net for web_search

teto:
There is a regression that we dont support proxy anymore ? can we export http_proxy and still have it working ?

gonna test before setting as "ready"


teto:
if anyone sees issues with it we can rollback/find a compromise before release.

PR #3227, opened 2026-09-09
fix: fix openai responses api calling, support_previous_response_id flag

cybaol:
When `use_response_api = true`,  set `support_previous_response_id = true` or `support_previous_response_id = false`,
the following error will occur:
```
Error: {
~        │  message = "previous_response_id requires an OpenAI API-key
~        │  account for HTTP requests",
~        │  type = "invalid_request_error"
~        │}
```

When  `support_previous_response_id = false`, should execute https://github.com/avante-corp/avante.nvim/blob/main/lua/avante/providers/openai.lua#L914-L917

So following config is working:
```lua
openai = {
          __inherited_from = 'openai',
          api_key_name = 'AVANTE_OPENAI_API_KEY',
          endpoint = 'https://tk.air-class.com:8080/v1',
          model = 'gpt-5.6-sol',
          use_response_api = true,
          support_previous_response_id = false,
          extra_request_body = {
            reasoning_effort = 'high',
            stream = true,
          },
},
```

But `support_previous_response_id = true` is still not working, maybe there are other reasons.

### DIFF
--- a/lua/avante/providers/openai.lua
+++ b/lua/avante/providers/openai.lua
@@ -900,7 +900,7 @@ function M:parse_curl_args(prompt_opts)
       end
     end
 
-    if has_function_outputs and self.last_response_id then
+    if has_function_outputs and self.last_response_id and provider_conf.support_previous_response_id then
       -- When sending function outputs, use previous_response_id
       base_body.previous_response_id = self.last_response_id
       -- Only send the function outputs, not the full history


cybaol:
`support_previous_response_id = true` still failed; this is most likely due to the relay server.

teto:
ty for PR. It looks reasonable and can be merged.

Let me take this opportunity to ask you since you may know more than I do about this:
 I am not familiar with `support_previous_response_id` but had a quick look at `use_response_api` and seems like it was introduced ~2 years ago and we default it to false. I was thinking of toggling it to true. Do you agree ?

Reading quickly about `support_previous_response_id` https://github.com/anomalyco/opencode/issues/34133 and https://developers.openai.com/api/docs/guides/migrate-to-responses seems like this helps use fewer tokens (even though openai doc says they still bill it ?). Seems worth enabling by default as well ?

cybaol:
I agree that making `use_response_api = true` by default is reasonable for the built-in OpenAI provider, especially since OpenAI recommends migrating to the Responses API over time.

  I would avoid making it the default for every OpenAI-compatible provider, though, since support varies between providers. Copilot is one concrete example: it supports the Responses API format for some models, but not `previous_response_id`.
  
  Reference docs: https://deepwiki.com/yetone/avante.nvim/7.5-provider-configuration

PR #3228, opened 2026-09-12
fix: Fix not closed file handle in view.lua

cybaol:
No need to open a file to determine whether it exists.
And the file was not closed at the end, resulting in a file handle resource leak.
Also, move the `exists` check to before the `has_permission_to_access` check.


teto:
ty

PR #3239, opened 2026-09-18
fix(templates): forbid inclusion of a file outside directory

teto:
reported in https://github.com/avante-corp/avante.nvim/issues/3185


teto:
seems like the wrong approach.
Either we use lib wrong considering https://github.com/mitsuhiko/minijinja/issues/413 or we shall ask for the feature upstream.

PR #3244, opened 2026-09-19
ci: run build test on various OSes

teto:
current release breaks on windows https://github.com/avante-corp/avante.nvim/actions/runs/35451577518/job/105919571831

PR #3247, opened 2026-09-20
Build/explicit cargo targets

RobsonMobarack:
### Description
This PR aligns the CI pipeline and the local Windows build script to ensure deterministic builds and prevent regressions caused by Cargo workspace default features. 

Resolves the build failure related to #3244.

### Changes
- **`.github/actions/build/action.yaml`**: Added `--no-default-features` and explicitly defined target crates using `-p` flags (`avante-tokenizers`, `avante-templates`, `avante-repo-map`, `avante-html2md`). This optimizes the pipeline by avoiding unnecessary workspace compilation.
- **`Build.ps1`**: Added `--no-default-features` to maintain strict parity with the CI workflow, ensuring consistent and deterministic local builds on Windows.


teto:
wow you are awesome. I can merge this as is to unblock 0.3 . 

Ideally I would like the CI to run `:AvanteBuild` for a single source of truth, so that the user runs the same code as CI.
This is actually done on linux but not on windows because the CI doesnt setup avante-nvim over there.
I'll try to see if I can improve the setup.

PR #3251, opened 2026-09-22
fix: windows locked dlls

### LINKED ISSUE #3250: bug: Update fails on Windows due to locked DLL file (UnauthorizedAccessException)
### Describe the bug

# Update fails on Windows due to locked DLL file (UnauthorizedAccessException)

### Description
When updating the plugin on Windows (e.g., using the Lazy plugin manager), the build script fails. 

Image

This happens because Neovim is running during the update, which means the `avante_html2md.dll`, `avante_repo_map.dll`, `avante_templates.dll` and `avante_tokenizers.dll` files are currently loaded in memory. Windows places a strict lock on files that are in use. Because of this lock, the PowerShell script cannot delete or overwrite the old DLL to install the new version.

### Actual Behavior
1. The build script runs during the update process.
2. PowerShell throws an `UnauthorizedAccessException` (`Remove-Item : Access to the path is denied`).
3. The plugin manager shows a "build failed" message.
4. **Hidden issue:** If the user restarts Neovim, the plugin appears to work normally. However, it is actually still using the **old** DLL, because the new one was never installed.

### Expected Behavior
The build script should successfully overwrite the DLL and finish the update without throwing permission errors, ensuring the new binaries are ready for the next Neovim restart.

### Steps to Reproduce
1. Use Windows.
2. Have an older version of `avante.nvim` installed and loaded in the current Neovim session.
3. Trigger a plugin update via your plugin manager (e.g., `:Lazy update`).
4. Look at the build logs to see the PowerShell permission error.

### Potential Solution
In Windows, you cannot delete a file that is in use, but **you can rename it**. 
A standard workaround for this is the "rename-and-replace" pattern in the `Build.ps1` script:
1. Rename the currently locked `.dll` to `.old` (this frees up the original file name).
2. Extract the new `.dll` in its place.
3. Add a step at the beginning of the script to silently delete any existing `.old` files from previous updates (since they are no longer in use).

### To reproduce

_No response_

### Expected behavior

_No response_

### Installation method

Use lazy.nvim:
```lua
{
  "yetone/avante.nvim",
        
  -- ⚠️ must add this setting! ! !
  build = vim.fn.has("win32") ~= 0
        and "powershell -ExecutionPolicy Bypass -File Build.ps1 -BuildFromSource false"
    or "make",
        
  event = "VeryLazy",
  version = false, -- Never set this value to "*"! Never!
  ---@module 'avante'
  ---@type avante.Config
  opts = {
    -- add any opts here
    -- this file can contain specific instructions for your project
    instructions_file = "avante.md",
    provider = "claude",
    providers = {
      claude = {
        endpoint = "https://api.anthropic.com",
        model = "claude-sonnet-4-20250514",
        timeout = 30000, -- Timeout in milliseconds
        extra_request_body = {
          temperature = 0.75,
          max_tokens = 20480,
        },
      },
    },
  },
  dependencies = {
    "nvim-lua/plenary.nvim",
    "MunifTanjim/nui.nvim",
    --- The below dependencies are optional,
    "nvim-mini/mini.pick", -- for file_selector provider mini.pick
    "nvim-telescope/telescope.nvim", -- for file_selector provider telescope
    "hrsh7th/nvim-cmp", -- autocompletion for avante commands and mentions
    "ibhagwan/fzf-lua", -- for file_selector provider fzf
    "stevearc/dressing.nvim", -- for input provider dressing
    "folke/snacks.nvim", -- for input provider snacks
    "nvim-tree/nvim-web-devicons", -- or echasnovski/mini.icons
    {
      -- support for image pasting
      "HakonHarnes/img-clip.nvim",
        
      event = "VeryLazy",
      opts = {
        
        -- recommended settings
        default = {
        
          prompt_for_file_name = false,
          drag_and_drop = {
        
            insert_mode = true,
          },
          -- required for Windows users
          use_absolute_path = true,
        },
      },
    },
    {
      -- Make sure to set this up properly if you have lazy=true
      opts = {
        
        latex = { enabled = false },
      },
    },
  },
}
```

### Environment

Neovim version: v0.12.5
Platform: Windows 11

### Repro

```lua

```

RobsonMobarack:
# Fix: Windows DLL file lock error during update

Closes #3250 

## The Problem
On Windows, updating the plugin while Neovim is open fails with an `UnauthorizedAccessException`. This happens because the OS locks `.dll` files while they are in use, preventing `Build.ps1` from overwriting them during the update or build process. This resulted in a "build failed" message or the silent loading of the old binaries.

## The Solution
To bypass the strict Windows file lock, I implemented the "rename-and-replace" pattern across two commits:

1. **Safe Unlock (PowerShell):** 
   Added a `Move-LockedDlls` function to the `Build.ps1` script. While Windows prevents deleting locked files, it allows renaming them. The script now renames the existing `.dll` files to `.old` before extracting or copying the new binaries. This frees up the original file names, allowing the update to complete successfully.
2. **Asynchronous Cleanup (Lua):** 
   To prevent leaving unused trash in the user's plugin folder, I added a Lua function that automatically finds and deletes these `.old` files. It uses `vim.defer_fn` (with a 2-second delay) to run in the background on the *next* Neovim startup.

## Impact and Benefits
* **Keeps the System Clean:** The `.old` files are safely deleted because, on the next session, the old Windows file locks are completely gone.
* **Zero Performance Impact:** The cleanup runs asynchronously in the background, ensuring Neovim's startup time is not slowed down.


lua/avante/init.lua
```diff
@@ -543,6 +543,9 @@ function M.setup(opts)
   H.keymaps()
   H.signs()
 
+  -- Clean up residual DLL files in Windows in the background
+  Utils.file.clean_windows_dll_trash()
+
```
teto, line 548: this slows down startup. Can it be done in the updater powershell script instead ? before moving the DLL, remove any old remaining .dll .

teto:
another solution I had in mind at point was to add the version into the name of the library: like `lua/avante_html2md.v0.3.1.so`. But then it means one has to clear up old libraries (or not), and it's probalbly going to be a pain to update everything (updater scripts/CI etc) so I prefer your small fix for now :)

RobsonMobarack:
The `*.old` file is intentionally temporary. The updater cannot remove it permanently during the update because the old DLL is still loaded by the current Neovim process. The Lua cleanup removes this Windows-specific temporary file on the next startup, so the final installed state is consistent with the other platforms.

```text
Normal state:
    avante_*.dll

Update:
    DLL is loaded
          ↓
    DLL → DLL.old
          ↓
    New DLL → DLL
```

After the update, the installation looks like this:

```text
    avante_*.dll
    avante_*.dll.old   ← still there
```

Without the cleanup, the next update can remove the previous `.old` file before moving the currently installed library:

```text
    Remove previous .old file
          ↓
    Rename current DLL → DLL.old
          ↓
    Install new DLL
```

But the result is again:

```text
    avante_*.dll
    avante_*.dll.old   ← again
```

In other words, saying that "the next update will clean it up" does not solve the directory hygiene problem. The next update removes the previous `.old` file, but then immediately creates a new one for the library that is still in use by the running Neovim process.

With `Utils.file.clean_windows_dll_trash()`, the lifecycle becomes:

```text
    Windows after update
          ↓
    New DLL installed
    Old DLL still loaded by Neovim
          ↓
    The .old file must remain temporarily
          ↓
    Neovim restarts
          ↓
    Old DLL is no longer in use
          ↓
    The .old file can be removed
          ↓
    Final state = normal state, as on other platforms
```

My intention is not to create a special persistent behavior for Windows. It is the opposite: Windows needs a temporary `.old` file because of the file-locking behavior, and the cleanup removes that implementation detail so the final plugin installation stays clean and consistent across platforms.

## About startup performance

The cleanup is deferred with `vim.defer_fn()`, so it does not block `M.setup()` or add a two-second delay to plugin startup. The cleanup runs later as a small filesystem operation, after the plugin has already been initialized.

I also do not see how moving this cleanup into the updater would remove the need for it. The updater can remove an older `.old` file before creating the new one, but it cannot remove the new `.old` file at the end of the update while that library may still be loaded by the running Neovim process.

## In short

**Without the cleanup:**

```text
Linux/macOS:
    normal installed state

Windows:
    normal installed state
    + one residual .old file
```

**With the cleanup:**

```text
Linux/macOS:
    normal installed state

Windows:
    normal installed state
```

The internal update mechanism is different on Windows, but the final state of the installation is kept consistent across platforms.

So the question, from my point of view, is not only whether we can delete an old file in the updater. It is whether we want to leave a Windows-specific temporary artifact in the plugin directory after every update, or clean it up on the next startup.

teto:
I think I can summarize your wall of text by "I dont want any lingering .dll.old in my folder". If you are not going to restore these dlls in case of failure, why not move them into a temporary folder like any OS has. The windows equivalent of `mktemp`, maybe https://learn.microsoft.com/en-us/windows/win32/fileio/creating-and-using-a-temporary-file?redirectedfrom=MSDN . This way we dont have to clean up anything, the OS will take care of it.

PR #3253, opened 2026-09-22
feat: introduce mega.cmdparse and applies it to a RAG

teto:
Introducing `:Avante rag start/stop/status/query`

Long term goal could be to unify all Avante commands with the prefix `:Avante` so it's easier to lazyload ?


teto:
my starting command:

```
uv run --python python3.13 avante-rag-service /tmp/avante-rag-service --log-level debug --embed-provider openai_like --embed-endpoint http://localhost:9932/v1 --embed-model embedding-gemma-300 --llm-provider openai_like --llm-endpoint http://localhost:9932/v1 --llm-model gemma4-12b --embed-api-key=toto --llm-api-key=toto
```

todo: 
- maybe bump python version, not sure what's the default these days but 3.13 looks old
- the executable from `nix build .#ragService` appears as "python3.13" in process manager. I would rather see `avante-rag-service`. 
- we should save more of avante's config into ~/.config/avante/rag_config.json so it's easier to start the executable without providing any argument.
- we need the service to log somewhere discoverable. For instance `:log avante-rag` should work. And we shall make it wrappable via systemd-cat or let user start the RAG via a function so it can wrap the start using a systemd service or to start its own varian (vectorcode for instance)

teto:
@cybaol @RobsonMobarack as you are the two most visible users out there, I wanted to warn you: this introduces a new dependency to the project `mega.cmdparse`. It's a very nice library, akin to argparse but for lua, and makes creation of vim commands a breeze. This is a new lua dependency and package managers are not great (yet) at handling dev rockspecs (the package format, you can see avante's at https://github.com/avante-corp/avante.nvim/blob/main/avante.nvim-scm-1.rockspec), hence I intend to release 0.4 just after merging this.

I've updated the installation instructions  (for rocks.nvim, it doesn't change a thing, it remains `:Rocks install avante.nvim`) so I am curious if this works for you/if you find the upgrade hard, if there is anything I should add to the doc. 

Cheers

PR #3256, opened 2026-09-23
fix: Optimize the UUID generator to prevent duplicate UUIDs.

### LINKED ISSUE #3245: bug: First prompt after reopening a conversation overwrites previous messages and mixes responses
### Describe the bug

If I type “hi” in a new conversation, it responds normally. However, if I close the file, reopen it, and then load the previous conversation, typing “hello” overwrites the previous “hi,” and the conversation mixes the responses to ‘hi’ and “hello” together.

This is picture:
Image

Image

### To reproduce

_No response_

### Expected behavior

_No response_

### Installation method

none

### Environment

none

### Repro

```lua

```

cybaol:
The old UUID generator relied on an unseeded `math.random()` sequence. After restarting Neovim, generated message UUIDs could repeat IDs already present in the loaded history, causing `add_history_messages()` to
  update those entries instead of appending new messages.

Fixes #3245


teto:
interesting catch. One quickfix could be to just set the randomizer seed to time().
I like that we call random once instead of multiple times. One problem though is that the generator specifically generates UUIDv4 see the template `local template = "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx"`, which is not the case after your change. Providers have different expectations about tool UUID so this could easily break for some providers. So please keep it UUIDv4.

NB: If you could write the explanation in the commit message  that would be helpful. This is helpful when one runs `git blame` on a piece of code, it means you dont have to understand why the code was changed. Ideally it would be a shorter explanation than the long LLM generated ones.

cybaol:
• The UUID is generated in four steps:
```lua
  function M.uuid()
    local bytes = { string.byte(assert(vim.uv.random(16)), 1, 16) }

    bytes[7] = (bytes[7] % 16) + 64
    bytes[9] = (bytes[9] % 64) + 128

    return string.format(
      "%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x",
      unpack(bytes)
    )
  end
  ```

  First, `vim.uv.random(16)` obtains 16 cryptographically secure random bytes from the operating system. These are raw bytes and do not contain hyphens.

  Next, `string.byte(..., 1, 16) ` converts those bytes into 16 numeric values stored in the bytes table.

  The function then sets the bits required by UUID v4:

  - `bytes[7] = (bytes[7] % 16) + 64`  sets the high four bits of byte 7 to 0100. This makes the third UUID group begin with 4, indicating version 4.
  - `bytes[9] = (bytes[9] % 64) + 128`  sets the high two bits of byte 9 to 10. Therefore, the fourth group begins with 8, 9, a, or b, which is the required UUID variant.

  Finally, string.format converts each byte into two lowercase hexadecimal characters:

  - %02x means lowercase hexadecimal, padded to two digits with a leading zero when necessary.
  - unpack(bytes) passes the 16 byte values as separate arguments.
  - The hyphens are fixed characters in the format string; they are not generated randomly.
  - The bytes are grouped as 4-2-2-2-6, producing the standard 8-4-4-4-12 UUID layout.

  For example:

  550e8400-e29b-41d4-a716-446655440000

  Therefore, the generated value conforms to the UUID v4 format:

  xxxxxxxx-xxxx-4xxx-[89ab]xxx-xxxxxxxxxxxx

cybaol:
Use 16 OS-generated random bytes from `vim.uv.random()`, set the UUID v4 version and variant bits, and formats them as a standard UUID.

Seeding `math.random()` with time is insufficient because processes started within the same time resolution can still produce identical sequences.

teto:
please now paste that message into the commit message, right now it has just the title `fix: Optimize the UUID generator to prevent duplicate UUIDs`.
It seems to generate UUIDv4 now but the new code kinda obfuscates this fact. Is there a way to make it clearer ?
for instance by hardcoding  the 4 like `string.format("%02x%02x%02x%02x-%02x%02x-4%01x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x", unpack(bytes))` (not tested).

We could also rename uuid() to uuid_v4() or have uuid() call uuid_v4() not to break api users (I can do that in a followup).

cybaol:
@teto Done.

cybaol:
@teto  Add the `UUID_V4_VERSION_BITS` and `UUID_VARIANT_BITS` constants to make it clearer.

